PR #16, opened 2018-10-05
Adds openid-user-attribute protocol mapper

alexashley:


mrparkers:
I fixed the CircleCI config so it builds PRs submitted via forks.  Closing and re-opening to trigger a build.

keycloak/openid_user_attribute_protocol_mapper.go
```diff
@@ -0,0 +1,151 @@
...
+	parentResourceId := clientId
+	parentResourcePath := "clients"
+
+	if clientScopeId != "" {
+		parentResourceId = clientScopeId
+		parentResourcePath = "client-scopes"
+	}
+
+	return fmt.Sprintf("/realms/%s/%s/%s/protocol-mappers/models", realmId, parentResourceId, parentResourcePath)
+}
```
mrparkers: Nice work supporting both client and client-scope :+1:

keycloak/protocol_mapper.go
```diff
@@ -0,0 +1,10 @@
+package keycloak
+
+// https://www.keycloak.org/docs-api/4.2/rest-api/index.html#_protocolmapperrepresentation
+type protocolMapper struct {
+	Id             string                 `json:"id,omitempty"`
+	Name           string                 `json:"name"`
+	Protocol       string                 `json:"protocol"`
+	ProtocolMapper string                 `json:"protocolMapper"`
+	Config         map[string]interface{} `json:"config"`
```
mrparkers: Why is the val of type `interface{}`?  It looks like it's always a string in the API responses I'm seeing.

mrparkers: Haha, yeah you probably should not assume that.  The Keycloak API is all over the place with stuff like that.

provider/keycloak_openid_user_attribute_protocol_mapper.go
```diff
@@ -0,0 +1,227 @@
...
+	data.Set("user_attribute", mapper.UserAttribute)
+	data.Set("claim_name", mapper.ClaimName)
+	data.Set("claim_value_type", mapper.ClaimValueType)
+}
+
+func resourceKeycloakOpenIdUserAttributeProtocolMapperCreate(data *schema.ResourceData, meta interface{}) error {
+	clientId := data.Get("client_id")
+	clientScopeId := data.Get("client_scope_id")
+
+	if clientId == "" && clientScopeId == "" {
```
alexashley: Do you know of a better way to do this? I have `ConflictsWith` set for both, but I don't know of a way to validate that one or the other must be set.

provider/keycloak_openid_user_attribute_protocol_mapper.go
```diff
@@ -0,0 +1,227 @@
...
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				Description:   "The mapper's associated client scope. Cannot be used at the same time as client_id.",
+				ConflictsWith: []string{"client_id"},
+			},
+			"add_to_id_token": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
```
mrparkers: I recently opened #27 to use `Computed` instead of `Default` so Keycloak provides the default value instead of the provider.  It might be worth looking at some of these attributes to see if Keycloak will send back a default when it is not specified.

alexashley: I considered using `Computed` for `claim_value_type`, but Keycloak  defaults to the empty string and treats that as the `String` type rather than setting a sensible default.

provider/keycloak_openid_user_attribute_protocol_mapper.go
```diff
@@ -0,0 +1,229 @@
...
+	return &schema.Resource{
+		Create: resourceKeycloakOpenIdUserAttributeProtocolMapperCreate,
+		Read:   resourceKeycloakOpenIdUserAttributeProtocolMapperRead,
+		Update: resourceKeycloakOpenIdUserAttributeProtocolMapperUpdate,
+		Delete: resourceKeycloakOpenIdUserAttributeProtocolMapperDelete,
+		Importer: &schema.ResourceImporter{
+			// import a mapper tied to a client:
+			// {{realmId}}/client/{{clientId}}/{{protocolMapperId}}
+			// or a client scope:
+			// {{realmId}}/client-scope/{{clientScopeId}}/protocolMapperId
```
mrparkers: `{{protocolMapperId}}`

provider/keycloak_openid_user_attribute_protocol_mapper.go
```diff
@@ -0,0 +1,229 @@
...
+				Optional:     true,
+				Description:  "Claim type used when serializing tokens.",
+				Default:      "String",
+				ValidateFunc: validation.StringInSlice([]string{"String", "long", "int", "boolean"}, true),
+			},
+		},
+	}
+}
+
+func mapFromDataToOpenIdUserAttributeProtocolMapper(data *schema.ResourceData) *keycloak.OpenIdUserAttributeProtocolMapper {
```
mrparkers: I've calling these functions something like `getOpenIdUserAttributeProtocolMapperFromData` but I think I like your name better, since the term `get` can mean a lot of different things here.  I might revisit these functions in the future and rename them.

provider/keycloak_openid_user_attribute_protocol_mapper.go
mrparkers: Thoughts on pulling out the validation code to a `Validate` function for consistency?  Other resources do this as well.

provider/keycloak_openid_user_attribute_protocol_mapper.go
```diff
@@ -0,0 +1,229 @@
...
+	clientId := data.Get("client_id").(string)
+	clientScopeId := data.Get("client_scope_id").(string)
+
+	var openIdUserAttributeMapper *keycloak.OpenIdUserAttributeProtocolMapper
+	var err error
+	if clientId != "" {
+		openIdUserAttributeMapper, err = keycloakClient.GetOpenIdUserAttributeProtocolMapperForClient(realmId, clientId, data.Id())
+	} else {
+		openIdUserAttributeMapper, err = keycloakClient.GetOpenIdUserAttributeProtocolMapperForClientScope(realmId, clientScopeId, data.Id())
+	}
```
mrparkers: I wonder if this would look better if there was just a single function, `keycloakClient.GetOpenIdUserAttributeProtocolMapper`, that took both client id and client scope id, and figured out which of the two this mapper belongs to based on which string is empty.  I'd rather have the code in the `keycloak` package do stuff like this.

provider/keycloak_openid_user_attribute_protocol_mapper_test.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+resource "keycloak_openid_user_attribute_protocol_mapper" "user-attribute-mapper-validation" {
+  	name = "%s"
+	realm_id = "${keycloak_realm.realm.id}"
+  	user_attribute = "foo"
+  	claim_name = "bar"
+	claim_value_type = "%s"
+}
+`, realmName, mapperName, invalidClaimValueType)
```
mrparkers: I don't have a good reason for this, but conventionally this is pulled out into a function that lives near the bottom of this file.

provider/keycloak_openid_user_attribute_protocol_mapper_test.go
```diff
@@ -0,0 +1,329 @@
...
+  realm_id            = "${keycloak_realm.realm.id}"
+}
+
+resource "keycloak_openid_user_attribute_protocol_mapper" "user-attribute-mapper-client-scope" {
+  	name = "%s"
+	realm_id = "${keycloak_realm.realm.id}"
+  	client_scope_id = "${keycloak_openid_client_scope.client-scope.id}"
+  	user_attribute = "foo"
+  	claim_name = "bar"
+}`, realmName, clientScopeId, mapperName)
```
mrparkers: Can you use tabs for indentation and align properties on `=` for all tf config?  I just opened #30 because other acceptance tests I wrote also don't do this, and I'd like it to be consistent.

keycloak/openid_user_attribute_protocol_mapper.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil
+}
+
+//func (keycloakClient *KeycloakClient) GetOpenIdUserAttributeProtocolMapperForClient(realmId, clientId, mapperId string) (*OpenIdUserAttributeProtocolMapper, error) {
+//	return keycloakClient.getOpenIdUserAttributeProtocolMapper(realmId, clientId, "", mapperId)
+//}
+//
+//func (keycloakClient *KeycloakClient) GetOpenIdUserAttributeProtocolMapperForClientScope(realmId, clientScopeId, mapperId string) (*OpenIdUserAttributeProtocolMapper, error) {
+//	return keycloakClient.getOpenIdUserAttributeProtocolMapper(realmId, "", clientScopeId, mapperId)
+//}
```
mrparkers: Forgot these

provider/keycloak_openid_user_attribute_protocol_mapper_test.go
```diff
@@ -0,0 +1,326 @@
...
+				Config: testKeycloakOpenIdUserAttributeProtocolMapper_claim(newRealmName, clientId, mapperName, attributeName),
+				Check:  testKeycloakOpenIdUserAttributeProtocolMapperExists(resourceName),
+			},
+		},
+	})
+}
+
+func testAccKeycloakOpenIdUserAttributeProtocolMapperDestroy() resource.TestCheckFunc {
+	return func(state *terraform.State) error {
+
```
mrparkers: Can you remove the extra whitespace here?

PR #17, opened 2018-10-05
refactor: rename keycloak_client to keycloak_openid_client

mrparkers:
After some thought and looking at the Keycloak API, I think it makes more sense to separate saml and openid clients into their own resources.  Although the APIs for clients aren't that much different between the two protocols, they really diverge for protocol mappers, and I think those will have to be different resources anyways.  This should hopefully trim down the amount of validation code I'd need to write to prepare for those.


README.md
```diff
@@ -12,13 +12,13 @@ Long term, I'd like to support as much as I can while I tinker with Keycloak in
...
 - [`keycloak_client_scope`](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/keycloak_client_scope.go)
 - [`keycloak_ldap_user_federation`](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/keycloak_ldap_user_federation.go)
 - [`keycloak_ldap_user_attribute_mapper`](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/keycloak_ldap_user_attribute_mapper.go)
 - [`keycloak_ldap_group_mapper`](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/keycloak_ldap_group_mapper.go)
 - [`keycloak_ldap_full_name_mapper`](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/keycloak_ldap_full_name_mapper.go)
-- [`keycloak_ldap_msad_user_account_controls_mapper`](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/keycloak_ldap_msad_user_account_controls_mapper.go)
+- [`keycloak_ldap_msad_user_account_control_mapper`](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/keycloak_ldap_msad_user_account_control_mapper.go)
```
mrparkers: This was an unrelated fix since the README link in master is broken right now

PR #18, opened 2018-10-05
new resource: keycloak_custom_user_federation

mrparkers:
I expect tests to fail on CI because I don't think it's possible to mount files within the docker executor for CircleCI.


provider/keycloak_custom_user_federation_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/mrparkers/terraform-provider-keycloak/keycloak"
+	"regexp"
+	"testing"
+)
+
+func TestAccKeycloakCustomUserFederation_basic(t *testing.T) {
+	skipIfEnvSet(t, "CI") // temporary while I figure out how to load this custom provider in CI
```
alexashley: Thoughts on a custom Keycloak image with the provider pre-installed?

mrparkers: It's either that, or setting up an entrypoint script that `wget`s the jar directly from github and places it in the correct folder.  I'm not sure I'm a fan of maintaining both a jar and a custom docker image just for these tests.

PR #27, opened 2018-10-09
refactor: allow keycloak to compute default values

mrparkers:
Instead of specifying the default values in the provider (based on the default values I got from the API), we can specify them as "computed" so I don't have to change the provider if a newer keycloak version uses a different default value.

### DIFF
--- a/provider/keycloak_ldap_user_federation.go
+++ b/provider/keycloak_ldap_user_federation.go
@@ -219,7 +219,7 @@ func getLdapUserFederationFromData(data *schema.ResourceData) *keycloak.LdapUser
 		userObjectClasses = append(userObjectClasses, userObjectClass.(string))
 	}
 
-	return &keycloak.LdapUserFederation{
+	ldap := &keycloak.LdapUserFederation{
 		Id:      data.Id(),
 		Name:    data.Get("name").(string),
 		RealmId: data.Get("realm_id").(string),
@@ -227,7 +227,6 @@ func getLdapUserFederationFromData(data *schema.ResourceData) *keycloak.LdapUser
 		Enabled:  data.Get("enabled").(bool),
 		Priority: data.Get("priority").(int),
 
-		ImportEnabled:     data.Get("import_enabled").(bool),
 		EditMode:          data.Get("edit_mode").(string),
 		SyncRegistrations: data.Get("sync_registrations").(bool),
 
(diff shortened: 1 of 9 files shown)


provider/keycloak_ldap_user_federation.go
```diff
@@ -55,7 +54,7 @@ func resourceKeycloakLdapUserFederation() *schema.Resource {
 			"import_enabled": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     true,
```
alexashley: I think there might be an issue here with fields that default to true (like `import_enabled)`. When the provider does the mapping from resource data to api model, the initial `data.Get("import_enabled")` [will return false (the empty value for booleans)](https://github.com/hashicorp/terraform/blob/15dda6335876145e306a23abf82b9840738f72c8/helper/schema/resource_data.go#L518-L520) since there's no user provided value. This will cause the request to Keycloak to have `config.importEnabled` set to false, and Keycloak won't supply the default value of true. 

Running a stripped down version of the example on master results in `config.importEnabled` being true, running it on this branch sets it to false.

```hcl
provider "keycloak" {
  client_id     = "terraform"
  client_secret = "884e0f95-0f42-4a63-9b1f-94274655669e"
  url           = "http://localhost:8080"
}

  realm                = "test"
  enabled              = true
  display_name         = "foo"

  account_theme        = "base"

  access_code_lifespan = "30m"
}

resource "keycloak_ldap_user_federation" "openldap" {
  name                    = "openldap"
  realm_id                = "master"

  enabled                 = true

  username_ldap_attribute = "cn"
  rdn_ldap_attribute      = "cn"
  uuid_ldap_attribute     = "entryDN"
  user_object_classes     = [
    "simpleSecurityObject",
    "organizationalRole"
  ]
  connection_url          = "ldap://openldap"
  users_dn                = "dc=example,dc=org"
  bind_dn                 = "cn=admin,dc=example,dc=org"
  bind_credential         = "admin"
}
```

mrparkers:
@alexashley nice catch.  I suppose the only bool values I was testing this with happened to be false by default anyways.

Fortunately, we can still compute default bool values from Keycloak by using `data.GetOkExists` instead of `data.Get`, but in the case of the `keycloak_ldap_user_federation` resource, this will be difficult since this is represented by a `Component` in the Keycloak API, which we eventually convert the `LdapUserFederation` model into.  I am not sure how to gracefully handle this without exposing the `Component` internal model to the provider namespace.

I'll try a couple of things before I attempt that refactor to see what you think.

mrparkers:
Closing in favor of #28

PR #29, opened 2018-10-13
feat: adds fields to keycloak_openid_client resource

mrparkers:


provider/keycloak_openid_client.go
```diff
@@ -26,31 +31,112 @@ func resourceKeycloakOpenidClient() *schema.Resource {
...
+				Optional:  true,
+				Computed:  true,
+				Sensitive: true,
+			},
+			"valid_redirect_uris": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Optional: true,
+				DiffSuppressFunc: func(k, remoteRedirectUri, local string, data *schema.ResourceData) bool {
+					if remoteRedirectUri == "" || local == "" {
```
alexashley: I didn't dig deep into the Keycloak source code, but I'm wondering if the sort behavior of the `redirectUri` list is due to the fact that Keycloak represents the list internally as a [`Set<string>`](https://github.com/keycloak/keycloak/blob/b478472b3578b8980d7b5f1642e91e75d1e78d16/server-spi/src/main/java/org/keycloak/models/ClientModel.java#L75) (the concrete implementations usually being a `HashSet`).  There's technically no guaranteed iteration order of a `HashSet`, but it does seem consistently weird. Here's an example: https://ideone.com/ab52Y7 

https://stackoverflow.com/questions/18648521/how-is-this-hashset-producing-sorted-output

alexashley: The best I can guess here is that the keys are divided into buckets based on the hash code value and are inserted into their buckets in order of increasing hash code; the buckets are in the insertion order of the array that backs a `HashMap`.

mrparkers:
This is ready to review.  I don't plan on implementing all of the fields for this resource quite yet.

keycloak/openid_client.go
```diff
@@ -4,15 +4,41 @@ import (
...
 type OpenidClient struct {
-	Id       string `json:"id,omitempty"`
-	ClientId string `json:"clientId"`
-	RealmId  string `json:"-"`
-	Protocol string `json:"protocol"`
+	Id                      string `json:"id,omitempty"`
+	ClientId                string `json:"clientId"`
+	RealmId                 string `json:"-"`
+	Name                    string `json:"name"`
+	Protocol                string `json:"protocol"`                // always openid-connect for this resource
```
alexashley: I don't have an actionable suggestion, but I do wish Go had some support for immutable fields on a struct for things like `protocol` here.

keycloak/realm.go
```diff
@@ -64,7 +64,13 @@ func (keycloakClient *KeycloakClient) UpdateRealm(realm *Realm) error {
 }
 
 func (keycloakClient *KeycloakClient) DeleteRealm(id string) error {
-	return keycloakClient.delete(fmt.Sprintf("/realms/%s", id))
+	err := keycloakClient.delete(fmt.Sprintf("/realms/%s", id))
+	if err != nil {
+		// For whatever reason, this fails sometimes with a 500 during acceptance tests. try again
```
alexashley: Is it worth making an issue to track this?

PR #51, opened 2018-11-02
new resource: keycloak_group

mrparkers:


keycloak/group.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (keycloakClient *KeycloakClient) GetGroup(realmId, id string) (*Group, error) {
+	var group Group
+
+	err := keycloakClient.get(fmt.Sprintf("/realms/%s/groups/%s", realmId, id), &group)
+	if err != nil {
+		return nil, err
+	}
+
+	group.RealmId = realmId // its important to set RealmId here because fetching the ParentId depends on it
```
alexashley: `its` -> `it's`

keycloak/group.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (keycloakClient *KeycloakClient) DeleteGroup(realmId, id string) error {
+	return keycloakClient.delete(fmt.Sprintf("/realms/%s/groups/%s", realmId, id))
+}
+
+func (keycloakClient *KeycloakClient) listGroups(realmId string) ([]*Group, error) {
+	var groups []*Group
+
+	err := keycloakClient.get(fmt.Sprintf("/realms/%s/groups", realmId), &groups)
```
alexashley: There are `first`/`max` query params, so I don't know if this would need to account for exceeding the default `max` (or if there even is one).

keycloak/group.go
```diff
@@ -0,0 +1,118 @@
...
+func (keycloakClient *KeycloakClient) groupParentId(group *Group) (string, error) {
+	// Check the path of the group being passed in.
+	// If there is only one group in the path, then this is a top-level group with no parentId
+	parts := strings.Split(strings.TrimPrefix(group.Path, "/"), "/")
+
+	if len(parts) == 1 {
+		return "", nil
+	}
+
+	groups, err := keycloakClient.listGroups(group.RealmId)
```
alexashley: Would it be possible to use the `search` query param with the path name at `len(parts)-2` and then check if `SubGroups` contains  the child id?

https://www.keycloak.org/docs-api/4.2/rest-api/index.html#_getgroups

PR #58, opened 2018-11-06
new resource: keycloak_group_memberships

mrparkers:


keycloak/user.go
```diff
@@ -44,3 +47,69 @@ func (keycloakClient *KeycloakClient) UpdateUser(user *User) error {
...
+	for _, user := range users {
+		if user.Username == username {
+			user.RealmId = realmId
+
+			return user, nil
+		}
+	}
+
+	// the requested user does not exist
+	// I don't think raising an error is appropriate here - consumers should check if the user is nil
```
alexashley: > //  I don't think raising an error is appropriate here - consumers should check if the user is nil

 This comment is a little strange because the method is private to the package, so all the consumers are in the package.

keycloak/user.go
```diff
@@ -44,3 +47,69 @@ func (keycloakClient *KeycloakClient) UpdateUser(user *User) error {
...
+}
+
+func (keycloakClient *KeycloakClient) RemoveUsersFromGroup(realmId, groupId string, usernames []interface{}) error {
+	for _, username := range usernames {
+		user, err := keycloakClient.getUserByUsername(realmId, username.(string)) // we need the user's id in order to remove them from a group
+		if err != nil {
+			return err
+		}
+
+		err = keycloakClient.RemoveUserFromGroup(user, groupId)
```
alexashley: Isn't there a chance that `user` is nil here, so this call will panic?

mrparkers: good catch.  I have this elsewhere in the file but forgot to do it here.

Issue #56, opened 2018-11-05
assign client scopes to a client


mrparkers:
Tagged as "new resource" because I think a separate resource is the best way to do this.

Authoritative:
```hcl
resource "keycloak_openid_client" "test_client" {
  client_id           = "test-client"
  name                = "test-client"
  realm_id            = "master"
  description         = "a test client"

  access_type         = "CONFIDENTIAL"
  valid_redirect_uris = [
    "http://localhost:8080/callback"
  ]
}

resource "keycloak_openid_client_scope" "test_client_scope" {
  name                = "foo1"
  realm_id            = "master"

  description         = "test"
  consent_screen_text = "hello"
}

resource "keycloak_openid_client_default_client_scopes" "authoritative_scopes" {
    realm_id       = "master"
    client_id      = "${keycloak_openid_client.test_client.id}"
    default_scopes = [
        "profile",
        "email",
        "${keycloak_openid_client_scope.test_client_scope.name}"
    ]
}
```

Non authoritative:
```hcl
resource "keycloak_openid_client" "test_client" {
  client_id           = "test-client"
  name                = "test-client"
  realm_id            = "master"
  description         = "a test client"

  access_type         = "CONFIDENTIAL"
  valid_redirect_uris = [
    "http://localhost:8080/callback"
  ]
}

resource "keycloak_openid_client_scope" "test_client_scope" {
  name                = "foo1"
  realm_id            = "master"

  description         = "test"
  consent_screen_text = "hello"
}

resource "keycloak_openid_client_default_client_scope" "non_authoritative_scopes" {
    realm_id      = "master"
    client_id     = "${keycloak_openid_client.test_client.id}"
    default_scope = "${keycloak_openid_client_scope.test_client_scope.name}"
}
```

thoughts:

1. it should be okay to reference scopes by their name since those are supposed to be unique across the realm.  my only issue with this is all of the other resources reference other resources by id, but doing this would make it hard to reference pre-existing scopes like "profile"
2. I'm borrowing the concept of authoritative vs non-authoritative resources from the Google provider which applies this concept very well.  There needs to be big warnings to inform users that you shouldn't use these resources together on the same client.
3. as mentioned in #14, I'm trying to separate openid and saml versions of the same resource (such as client) because the APIs are so different.  however it might not be needed in this case.
4. optional scopes will have a different resource than the example above but will work the same

mrparkers:
[...]

PR #59, opened 2018-11-08
new resource: keycloak_openid_client_default_scopes

mrparkers:
Addresses half of #56

### DIFF
--- a/keycloak/openid_client.go
+++ b/keycloak/openid_client.go
@@ -118,7 +118,7 @@ func (keycloakClient *KeycloakClient) GetOpenidClientDefaultScopes(realmId, clie
 	return scopes, nil
 }
 
-func (keycloakClient *KeycloakClient) AttachOpenidClientDefaultScopes(realmId, clientId string, scopeNames []interface{}) error {
+func (keycloakClient *KeycloakClient) AttachOpenidClientDefaultScopes(realmId, clientId string, scopeNames []string) error {
 	openidClient, err := keycloakClient.GetOpenidClient(realmId, clientId)
 	if err != nil && ErrorIs404(err) {
 		return fmt.Errorf("validation error: client with id %s does not exist", clientId)
@@ -145,7 +145,7 @@ func (keycloakClient *KeycloakClient) AttachOpenidClientDefaultScopes(realmId, c
 	return nil
 }
 
-func (keycloakClient *KeycloakClient) DetachOpenidClientDefaultScopes(realmId, clientId string, scopeNames []interface{}) error {
+func (keycloakClient *KeycloakClient) DetachOpenidClientDefaultScopes(realmId, clientId string, scopeNames []string) error {
 	allOpenidClientScopes, err := keycloakClient.listOpenidClientScopesWithFilter(realmId, includeOpenidClientScopesMatchingNames(scopeNames))
 	if err != nil {
 		return err
--- a/keycloak/openid_client.go
+++ b/keycloak/openid_client.go
@@ -93,6 +93,13 @@ func (keycloakClient *KeycloakClient) GetOpenidClientDefaultScopes(realmId, clie
 }
 
 func (keycloakClient *KeycloakClient) AttachOpenidClientDefaultScopes(realmId, clientId string, scopeNames []interface{}) error {
+	_, err := keycloakClient.GetOpenidClient(realmId, clientId)
+	if err != nil && ErrorIs404(err) {
+		return fmt.Errorf("validation error: client with id %s does not exist", clientId)
+	} else if err != nil {
+		return err
+	}
+
 	allOpenidClientScopes, err := keycloakClient.listOpenidClientScopes(realmId)
 	if err != nil {
 		return err
--- a/keycloak/openid_client.go
+++ b/keycloak/openid_client.go
@@ -70,6 +70,32 @@ func (keycloakClient *KeycloakClient) GetOpenidClient(realmId, id string) (*Open
 	return &client, nil
 }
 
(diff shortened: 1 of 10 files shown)


keycloak/openid_client.go
```diff
@@ -80,3 +106,69 @@ func (keycloakClient *KeycloakClient) UpdateOpenidClient(client *OpenidClient) e
...
+	if openidClient.BearerOnly {
+		return fmt.Errorf("validation error: client with id %s uses access type BEARER-ONLY which does not use scopes", clientId)
+	}
+
+	allOpenidClientScopes, err := keycloakClient.listOpenidClientScopes(realmId)
+	if err != nil {
+		return err
+	}
+
+	for _, scopeName := range scopeNames {
```
zpratt: this block looks very similar to the one that starts on L160. Is there a way to extract some bits of this to combine them?

keycloak/openid_client.go
```diff
@@ -80,3 +106,57 @@ func (keycloakClient *KeycloakClient) UpdateOpenidClient(client *OpenidClient) e
...
+	err := keycloakClient.get(fmt.Sprintf("/realms/%s/clients/%s/default-client-scopes", realmId, clientId), &scopes)
+	if err != nil {
+		return nil, err
+	}
+
+	return scopes, nil
+}
+
+func (keycloakClient *KeycloakClient) AttachOpenidClientDefaultScopes(realmId, clientId string, scopeNames []interface{}) error {
+	openidClient, err := keycloakClient.GetOpenidClient(realmId, clientId)
```
alexashley: Might be worth adding a length check for the `scopeNames` slice, since there's no reason to make those two API calls below if there are no scopes to attach.

alexashley: Could `scopeNames` be a `[]string`?

keycloak/openid_client_scope.go
```diff
@@ -51,3 +51,35 @@ func (keycloakClient *KeycloakClient) UpdateOpenidClientScope(clientScope *Openi
...
+	var openidClientScopes []*OpenidClientScope
+
+	err := keycloakClient.get(fmt.Sprintf("/realms/%s/client-scopes", realmId), &clientScopes)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, clientScope := range clientScopes {
+		if clientScope.Protocol == "openid-connect" && filter(&clientScope) {
+			func(cs OpenidClientScope) {
```
alexashley: What's with the IIFE here?

keycloak/openid_client.go
```diff
@@ -70,6 +70,32 @@ func (keycloakClient *KeycloakClient) GetOpenidClient(realmId, id string) (*Open
 	return &client, nil
 }
 
+func (keycloakClient *KeycloakClient) GetOpenidClientByClientId(realmId, clientId string) (*OpenidClient, error) {
+	var clients []OpenidClient
+	var clientSecret openidClientSecret
+
+	err := keycloakClient.get(fmt.Sprintf("/realms/%s/clients", realmId), &clients)
```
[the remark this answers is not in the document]
mrparkers: Good catch.  I really should reference the official API docs more.

keycloak/openid_client.go
mrparkers: > Might be worth adding a length check for the `scopeNames` slice, since there's no reason to make those two API calls below if there are no scopes to attach.

I considered this as well but ended up not doing it because there is still value in using this resource to ensure that no default scopes are attached to a client, like so:

```hcl
resource "keycloak_openid_client_default_scopes" "no_default_scopes" {
    realm_id       = "test-realm"
    client_id      = "test-client"
    default_scopes = []
}
```

Since this is a valid use case, then checking the length of the slice before making those API calls would allow a potentially invalid configuration to exist and blow up as soon as one scope is added.

mrparkers: > Could `scopeNames` be a `[]string`?

Technically yes, but the return value of `*schema.Set -> List()` is `[]interface{}`, and you need to loop over each item in order to convert that to `[]string` which I have to do anyways in `includeOpenidClientScopesMatchingNames`.

keycloak/openid_client_scope.go
mrparkers: Good question!  This one had me scratching my head for a while.

The reason for this is because the loop variable `clientScope` is the same memory reference holding a different value for each iteration.  Since I'm appending a pointer to `clientScope` to a slice when my conditional is true, the end result (without the IIFE) is a slice filled with the same pointer to whatever value happens to be last in `clientScopes`.

The IIFE allows me to pass a copy of the value to the function, meaning that a different memory address is stored in the resulting slice.

keycloak/openid_client.go
alexashley: Fair enough, but the `keycloak` package is useful independently of the Terraform provider. So from that perspective, it makes sense for it to be `[]string`.

mrparkers: Yeah, that's a good point.  You've convinced me, I'll make this change.

keycloak/openid_client_scope.go
alexashley: @mrparkers After looking at https://github.com/mrparkers/terraform-provider-keycloak/commit/8bf7dc82bd1de8053c3a7ddbdbe42604318b4187, I think an explicit `new` would be clearer (and could be done conditionally, instead of copying every client scope and not always using it).

IMO, the IIFE and the new `filter` function signature are sort of hiding that a copy has to happen somewhere.

mrparkers: Yeah, I support it is idiomatic in Go to not use syntax to hide stuff like this.

Fixed in https://github.com/mrparkers/terraform-provider-keycloak/pull/59/commits/f0941814c92673e41c8c4a7b751c4e8ff38c3c61

PR #65, opened 2018-11-18
Adds docs

mrparkers:


docs/resources/keycloak_group.md
```diff
@@ -0,0 +1,55 @@
...
+    parent_id = "${keycloak_group.parent_group.id}"
+    name      = "child-group"
+}
+```
+
+### Argument Reference
+
+The following arguments are supported:
+
+- `realm_id` - (Required) The realm this groups exists in.
```
alexashley: Thoughts on auto-generating the arguments section by reading the TF schema and extracting descriptions?

mrparkers: I considered this, but decided it would probably be a lot of work to do, especially for resources or attributes that have a lot of configuration options.

I would be totally open to trying something out if you have any ideas though!

Issue #53, opened 2018-11-05
Add ability to disable direct access grants for a client


alexashley:
This would entail adding `direct_access_grants_enabled` to `keycloak_openid_client`. Most applications should go through the standard flow and should not be able to use the resource owner credentials flow.

PR #68, opened 2018-11-21
feat: adds support for enabling and disabling oauth flows for keycloak_openid_client

mrparkers:


mrparkers:
Closes #53

keycloak/openid_client.go
```diff
@@ -25,12 +25,25 @@ type OpenidClient struct {
...
+	ServiceAccountsEnabled    bool `json:"serviceAccountsEnabled"`
+
 	ValidRedirectUris []string `json:"redirectUris"`
 }
 
 func (keycloakClient *KeycloakClient) ValidateOpenidClient(client *OpenidClient) error {
-	if !client.BearerOnly && len(client.ValidRedirectUris) == 0 {
-		return fmt.Errorf("validation error: must specify at least one valid redirect uri if access type is PUBLIC or CONFIDENTIAL")
+	if client.BearerOnly && (client.StandardFlowEnabled || client.ImplicitFlowEnabled || client.DirectAccessGrantsEnabled || client.ServiceAccountsEnabled) {
+		return fmt.Errorf("validation error: bearer-only clients cannot issue tokens; no oauth2 flows can be enabled for this client")
```
alexashley: >bearer-only clients cannot issue tokens

It's Keycloak that issues the tokens to clients, not so much clients making tokens themselves.

keycloak/openid_client.go
```diff
@@ -25,12 +25,25 @@ type OpenidClient struct {
...
+	}
+
+	if (client.StandardFlowEnabled || client.ImplicitFlowEnabled) && len(client.ValidRedirectUris) == 0 {
+		return fmt.Errorf("validation error: standard (authorization code) and implict flows require at least one valid redirect uri")
```
alexashley: `implict` -> `implicit`

provider/keycloak_openid_client.go
```diff
@@ -55,6 +55,26 @@ func resourceKeycloakOpenidClient() *schema.Resource {
...
+			},
+			"implicit_flow_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+			"direct_access_grants_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
```
alexashley: Much better than Keycloak's default.

keycloak/openid_client.go
mrparkers: yeah the wording is a bit off here, it should probably be something like

> Keycloak cannot issue tokens for bearer-only clients

provider/keycloak_openid_client.go
mrparkers: I also changed the default for the standard flow to false as well, thoughts on that?  I figured this made sense because the standard flow requires at least one valid redirect uri and I didn't want users to get validation errors with a config that doesn't specify either

alexashley: I lean towards having it on by default, since it's the generally recommended flow (although a lot of things use the hybrid flow, which is probably why implicit is on by default). It's also more useful than creating a client with no available flows as the default.

mrparkers: I gave it some thought and I still think I'm going to turn it off by default.  My issue is that I don't want a resource like this failing due to not specifying redirect uris:

```hcl
resource "keycloak_openid_client" "client" {
    realm_id = "test"
    client_id = "test"
}
```

The obvious downside is that the example above is pretty much useless with every flow turned off, but I like the idea of being explicit anyways and it anyone who runs into this should quickly figure out the issue.  I can also make a point to call this out in the docs.

PR #86, opened 2019-01-29
Internationalization parameters on realm

tomrutsaert:
Added internationalizationEnabled, supportedLocales and DefaultLocale.
Probably needs more tests, (First time writing something in go yeah)


mrparkers:
Also, it might be a good idea to hit `/serverinfo` to check if the given locales are present on the server as an extra validation step.

tomrutsaert:
> Also, it might be a good idea to hit `/serverinfo` to check if the given locales are present on the server as an extra validation step.

Checking the available locales via /serverinfo is not that simple. It also depends which themes was selected per section (login, admin, email, account, ....)
Furthermore I do not know if I need to crosscheck if the locale is present in each of the selected themes of each section (login, admin, email, account, ....)

At the moment this seems like a lot of work, and investigation, I might take this up later.

Does this stop merging the PR in master?
I have added more tests and put internationalization in a separate object.

(Next on my agenda is adding SMTP, Security Defense Browser Header, ... Required Actions, and custom autenticators + bindings)

mrparkers:
I won't block the PR for not checking via `/serverinfo`, I was just curious if you had considered that.  I understand your concerns about that so I can live without that extra validation.

It looks like the git history for this PR is a bit messed up - a lot of the contributions shown already exist in master.  Would you be able to cherry pick your changes against a fresh copy of master?  As is, the PR is difficult to review when it showing over 9000 lines changed.

Thanks!

tomrutsaert:
Ok, I will close this PR and create a new one based on a refresh copy of master.

PR #89, opened 2019-02-07
Avoid Keycloak requests in provider setup

Whazor:
These changes make it possible to let other resources setup Keycloak. For example a docker container.

I also had to change the `create-terraform-client.sh` in order to setup Keycloak within docker, see: https://gist.github.com/Whazor/50db4e39252577f4fa6622a339551bf9

Can be these that these changes fail other tests or break configurations...

### DIFF
--- a/keycloak/keycloak_client.go
+++ b/keycloak/keycloak_client.go
@@ -16,6 +16,7 @@ type KeycloakClient struct {
 	baseUrl           string
 	clientCredentials *ClientCredentials
 	httpClient        *http.Client
+	initialLogin      bool
 }
 
 type ClientCredentials struct {
@@ -42,12 +43,8 @@ func NewKeycloakClient(baseUrl, clientId, clientSecret string) (*KeycloakClient,
 			ClientId:     clientId,
 			ClientSecret: clientSecret,
 		},
-		httpClient: httpClient,
-	}
-
-	err := keycloakClient.login()
-	if err != nil {
-		return nil, err
+		httpClient:   httpClient,
+		initialLogin: false,
 	}
 
 	return &keycloakClient, nil
@@ -154,6 +151,13 @@ func (keycloakClient *KeycloakClient) addRequestHeaders(request *http.Request) {
 Sends an HTTP request and refreshes credentials on 403 or 401 errors
 */
 func (keycloakClient *KeycloakClient) sendRequest(request *http.Request) ([]byte, string, error) {
+	if !keycloakClient.initialLogin {
+		keycloakClient.initialLogin = true
+		err := keycloakClient.login()
+		if err != nil {
+			return nil, "", fmt.Errorf("error logging in: %s", err)
+		}
+	}
 	requestMethod := request.Method
 	requestPath := request.URL.Path
 


mrparkers:
Hi @Whazor, thanks for the PR!  Unfortunately, I don't understand the benefit of moving the initial login process from the client instantiation to the first call to `sendRequest`.  Could you clarify this?  I think a repo with steps to re-produce the issue you're having would be helpful.

If this adds new functionality that you'd like to depend on, I'd also like to see if we can get a test case written to ensure that this behavior doesn't accidentally change in the future.

Whazor:
The following Terraform definitions replicates the issue, where you use terraform to setup Keycloak: 
```hcl
provider "docker" { }
resource "docker_image" "keycloak" {
  name = "jboss/keycloak"
}
resource "random_uuid" "terraform_keycloak" { }
resource "docker_container" "keycloak" {
  image = "${docker_image.keycloak.latest}"
  name  = "keycloak"
  env = [
    "KEYCLOAK_USER=admin",
    "KEYCLOAK_PASSWORD=admin"
  ]
  ports {
      internal = 8080
      external = 8080
  }
  provisioner "local-exec" {
    command = "KEYCLOAK_CLIENT_SECRET=${random_uuid.terraform_keycloak.result} ./create-terraform-client.sh"
  }
}
provider "keycloak" {
    client_id     = "terraform"
    client_secret = "${random_uuid.terraform_keycloak.result}"
    # Adding variables of other resources should make the provider depend on the other resource
    url           = "http://localhost:${docker_container.keycloak.ports.0.external}"
}
    realm   = "test"
    enabled = true
}
```
with the modified `create-terraform-client.sh` that waits for docker setup:  https://gist.github.com/Whazor/50db4e39252577f4fa6622a339551bf9

Whazor:
Maybe it is a bug within terraform? If the provider setup is delayed, especially the call to `ConfigureFunc` then this PR would not be needed.

mrparkers:
Thanks for the explanation, that makes it much more clear.  While I'm open to merging this change if it helps for your use case, I did, however, discover this part of the Terraform docs that seems to warn against configuring providers with computed values that are unknown until after running `tf apply`: https://www.terraform.io/docs/configuration/providers.html#interpolation

> Although in principle it is possible to use any interpolation expression within a provider configuration argument, providers must be configurable to perform almost all operations within Terraform, and so it is not possible to use expressions whose value cannot be known until after configuration is applied, such as the id of a resource.

I am not sure if this applies to your exact problem, but it does concern me a little bit that we may be diverging a little bit from how Terraform providers are expected to be configured.  From anecdotal experience, I've always had to do a tiny bit of manual configuration with every Terraform provider I've worked with before I could start using it.

Whazor:
Normally it should expected that provider information can be coming from data sources, which is mentioned in the next paragraph.

However, I currently believe that Terraform itself should call the `ConfigureFunc` method at the correct time. Not at the planning stage, but only when the provider is going to be used... Because I saw that multiple providers are having similar issues.

So I suggest ignoring my pull request and I keep my repo with the fix for people who might need it. Furthermore I post in an issue within Terraform and will keep an eye on it over the next couple weeks.

mrparkers:
Sounds good!  Please link the issue here if you open one so I can follow along and perhaps learn a few things along the way.

Thanks!

ltscif:
Any chance this can be reopened? The referenced quote above no longer exists in the .12 docs and not having this capability clearly prevents keycloak itself from being provisioned by terraform and used as input to this provider.  At the very least, what about implementing this PR with an optional boolean attribute for the provider to omit initial login, disabled by default.  Would that be an acceptable compromise?

mrparkers:
Just an fyi - the warning I quoted still exists, but the verbiage is different:

```
However, since provider configurations must be evaluated in order to perform any resource type action, provider configurations may refer only to values that are known before the configuration is applied. In particular, avoid referring to attributes exported by other resources unless their values are specified directly in the configuration.
```

I would be okay with the compromise you proposed if it helps your workflow.

ltscif:
Compromise provided in #136

PR #92, opened 2019-02-13
added identity provider resource

AndrewChubatiuk:
Restored #61 PR


mrparkers:
@AndrewChubatiuk what are your thoughts into splitting each identity provider into multiple resources?  I'm concerned that trying to handle every identity provider in one resource would make it too complex and difficult to properly test.

AndrewChubatiuk:
Okay, I can split them

mrparkers:
Cool.  Let's just start with a couple to see if that route makes sense.  I understand that, for the most part, the APIs are very similar, so I expect a decent amount of code duplication.

AndrewChubatiuk:
Well, yes, duplicating all code means that we need duplicate all test cases and make changes everywhere simultaneously. I have one suggestion:
I could leave only generic attributes in `keycloak_identity_provider` and add `config` attribute of type map that accepts any key-value pairs and create datasources which configure and make validation of each identity provider:
`
resource keycloak_identity_provider oidc1 {
  alias   = "oidc"
  realm   = "master"
  enabled = true
  config {
      "token_url" = "value"
      "lol" = ""
  }
}

resource keycloak_identity_provider oidc2 {
  realm   = "master"
  enabled = true
  config = "${data.keycloak_google_identity_provider.example.config}" {
}

data keycloak_oidc_identity_provider_config example {
  token_url         = "https://example.com"
  authorization_url = "https://example.com"
  client_id         = "<your_client_id>"
  client_secret     = "<your_client_secret>"
}
`

AndrewChubatiuk:
This approach is similar to `aws_iam_policy_document` datasource in `terraform-provider-aws` but it's less complex

mrparkers:
That's an interesting way to approach this, but I don't think data sources are intended to be used that way.  Typically, a data source is used when you want Terraform to fetch some configuration from a remote server so you can use it to help configure other resources.

I agree that the code duplication this would produce is unfortunate, but a lot of the Keycloak APIs for different resources are very similar so I don't think there is much we can do to avoid that.  There is already a fair amount of this in the `keycloak_*_protocol_mapper` resources, and the `keycloak` package tries to consolidate similar API fields into a single file to help with this.

In general, I think I still prefer code duplication over a single resource that attempts to handle many use cases.  I want to avoid a scenario where a change to this resource to fix an issue with the Google identity provider (for example) may inadvertently break the support for the Twitter identity provider.

AndrewChubatiuk:
[`aws_iam_policy_document` datasource in aws provider](https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html) doesn't fetch data, it just makes a policy validation. On a top of [this page](https://www.terraform.io/docs/configuration/data-sources.html) page there's a definition of data source. It allows data not only to be fetched but also computed. 
Also all [ignition datasources](https://www.terraform.io/docs/providers/ignition/index.html) do not fetch data from internet, they just prepare a valid configuration in the output

mrparkers:
Thanks for sharing.  I see how having a data source that provides a consistent interface for configuration could be useful, especially for validation purposes.

Despite this, I am still of the opinion that these identity providers should be split to multiple resources.  If the code duplication issue becomes a problem, we can take advantage of other solutions, such as code generation or factory functions/methods.

AndrewChubatiuk:
@mrparkers 
Created generic Identity provider, identity provider mapper and created OIDC and SAML providers with all mappers that are used by those providers.

mrparkers:
At first glance, this seems like a good way to have separate identity provider resources while avoiding code duplication.  Nice work!

I need to familiarize myself with this Keycloak feature and the corresponding API before I can leave a proper review.  In the meantime, would you be able to add some acceptance tests for this change?

AndrewChubatiuk:
Ok, no problem, but I'm going to start with tests immediately after your review if you don't mind

mrparkers:
Yup that's fair.  Could you update `example/main.tf` with a couple of resource definitions that closely match your use case?  That will help me get a better understanding of this PR.

AndrewChubatiuk:
Added examples that are close to the cases I have with identity providers right now

keycloak/util.go
```diff
@@ -1,11 +1,70 @@
...
 
 import (
+	"bytes"
 	"strconv"
 	"strings"
 	"time"
 )
 
+type KeycloakBoolQuoted bool
+type KeycloakBool bool
```
mrparkers: I think I understand what these are for, but let me just confirm:

- `KeycloakBool` can be either `""` (false), `true`, or `false`.
- `KeycloakBoolQuoted` can be either `""` (false), `"true"`, or `"false"`. This is used in "config" API objects where all values are in quotes.

Is this correct?

If so, could you help me understand why `KeycloakBool` is needed?  It looks like you used it in a few places in the generic identity provider helpers, but it looks like those APIs always return booleans.  I haven't seen any of them return empty quotes (running v4.8.3.Final).

provider/generic_keycloak_identity_provider.go
```diff
@@ -0,0 +1,196 @@
...
+
+var getIdentityProviderDataFunctions = map[string]func(data *schema.ResourceData) (*keycloak.IdentityProvider, error){
+	"oidc": getOidcIdentityProviderFromData,
+	"saml": getSamlIdentityProviderFromData,
+}
+
+var setIdentityProviderDataFunctions = map[string]func(data *schema.ResourceData, identityProvider *keycloak.IdentityProvider) error{
+	"oidc": setOidcIdentityProviderData,
+	"saml": setSamlIdentityProviderData,
+}
```
mrparkers: I like the code re-use here, nice idea!

One suggestion - it seems a little strange that the file for the generic identity provider values has to be aware of the helper functions for the `oidc` and `saml` identity providers.  What do you think about passing those helper functions in as an argument to `resourceKeycloakIdentityProviderCreate`, etc instead?

I'll push a commit to your fork with my suggested changes.  If you agree, could you also refactor the mappers to follow this pattern as well?

provider/keycloak_oidc_identity_provider.go
```diff
@@ -0,0 +1,114 @@
...
+
+func resourceKeycloakOidcIdentityProvider() *schema.Resource {
+	oidcSchema := map[string]*schema.Schema{
+		"backchannel_supported": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     true,
+			Description: "Does the external IDP support backchannel logout?",
+		},
+		"use_jwks_url": {
```
mrparkers: This switch in the console, when turned on, reveals an input where the user can specify the JWKs url to use.  I think if the ability to configure this switch is exposed, we also need to support the ability to specify the JWKs url.

My suggestion would be to change this attribute to `jwks_url`, which would be an optional string.  If it is provided, we will set `useJwksUrl` to `true` and `jwksUrl` to the string that was provided.  If this is not provided, we will set `useJwksUrl` to `false`.

Alternatively, if you don't want to add this, you can just remove this attribute altogether.  It's okay if you only want to support the functionality you need right now, and it can always be added later if someone else needs it.

provider/generic_keycloak_identity_provider.go
AndrewChubatiuk: Right, I agree with that solution

keycloak/util.go
AndrewChubatiuk: I tested with keycloak 4.5.0 and earlier and all boolean values are just empty quotes if they are default and quoted boolean values if they aren't

mrparkers: I took a closer look and it looks like attributes using `KeycloakBool` are only empty strings when using the console to create them with default values.  All subsequent fetches seem to be actual booleans instead of strings.

So I don't _think_ `KeycloakBool` is needed, but I am fine with leaving it if you think it is necessary.

This wouldn't really be an issue if Keycloak had a more consistent REST API :smile:

mrparkers:
Overall this PR looks good.  I don't have any other suggested changes for the resource attributes, so I think it's okay to start adding test coverage.

AndrewChubatiuk:
@mrparkers added some tests

mrparkers:
Thanks, I'll take some time this weekend to review it.

provider/keycloak_oidc_identity_provider.go
```diff
@@ -0,0 +1,111 @@
...
+	oidcResource.Update = resourceKeycloakIdentityProviderUpdate(getOidcIdentityProviderFromData, setOidcIdentityProviderData)
+	return oidcResource
+}
+
+func getOidcIdentityProviderFromData(data *schema.ResourceData) (*keycloak.IdentityProvider, error) {
+	rec, _ := getIdentityProviderFromData(data)
+	rec.ProviderId = "oidc"
+	rec.Config = &keycloak.IdentityProviderConfig{
+		BackchannelSupported: keycloak.KeycloakBoolQuoted(data.Get("backchannel_supported").(bool)),
+		UseJwksUrl:           keycloak.KeycloakBoolQuoted(true),
```
mrparkers: I left this comment originally and I don't think it was addressed - I don't think this attribute is useful without the ability to customize the jwks url (correct me if I'm wrong here).  If this is the case, I think this attribute should either be removed or updated to allow users to customize this url

provider/keycloak_saml_identity_provider.go
```diff
@@ -0,0 +1,173 @@
...
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Hide On Login Page.",
+		},
+		"name_id_policy_format": {
+			Type:         schema.TypeString,
+			Optional:     true,
+			Default:      "",
+			ValidateFunc: validation.StringInSlice(keys(nameIdPolicyFormats), false),
+			StateFunc: func(value interface{}) string {
```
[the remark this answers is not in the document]
AndrewChubatiuk: What exactly?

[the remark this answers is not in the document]
mrparkers: The `StateFunc`.  I don't necessarily have a problem with it, it's just the first time it has been used in this provider and I am curious what the need was.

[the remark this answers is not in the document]
AndrewChubatiuk: Allowed values for `name_id_policy_format` attribute are same as in Keycloak UI (Windows Domain Qualified Name, Persistent, etc) and StateFunc converts this value to an actual value which is used in API requests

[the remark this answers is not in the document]
mrparkers: Okay, so users would be expected to do something like this?

```
name_id_policy_format = "Windows Domain Qualified Name"
```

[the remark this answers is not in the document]
AndrewChubatiuk: Yes

provider/keycloak_oidc_identity_provider.go
AndrewChubatiuk: Removed

provider/keycloak_attribute_importer_identity_provider_mapper.go
```diff
@@ -0,0 +1,79 @@
...
+	rec.Config = &keycloak.IdentityProviderMapperConfig{
+		UserAttribute: data.Get("user_attribute").(string),
+	}
+	if identityProvider.ProviderId == "saml" {
+		if attr, ok := data.GetOk("attribute_friendly_name"); ok {
+			rec.Config.AttributeFriendlyName = attr.(string)
+		} else if attr, ok := data.GetOk("attribute_name"); ok {
+			rec.Config.Attribute = attr.(string)
+		} else {
+			return nil, fmt.Errorf(`provider.keycloak: keycloak_attribute_importer_identity_provider_mapper: %s: neither "attribute_name" nor "attribute_friendly_name" are set`, data.Get("name").(string))
```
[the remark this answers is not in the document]
AndrewChubatiuk: There are about 15 identity providers and mappers for these providers are the same or almost same as for OIDC or SAML provider. If we're going to have a separate mapper for each IDP type there'll be about 60 different mappers which confuses more than several conditions in the same provider.

mrparkers:
A couple of general comments unrelated to the review -

1) Apologies for taking so long with getting this PR across the finish line, I spent all of last week at a Conference so I wasn't able to dedicate time into reviewing this.
2) After my comments from the last review are addressed and this is merged, I'll spend some time within the next few weeks writing docs for this so I can better learn how to use this feature myself.  I'll cut a 1.7.0 release so you can use this right away, but I may make slight changes after I understand this feature better and release them in 1.7.1.  Is that alright with you?

AndrewChubatiuk:
1. That's okay.
2. Yes, no problem

provider/keycloak_attribute_importer_identity_provider_mapper.go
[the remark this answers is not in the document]
AndrewChubatiuk: What do you think about it?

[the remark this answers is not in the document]
mrparkers: That's a fair point.  I didn't consider that there were so many different mappers for different identity providers.

With this in mind, what would you think about updating the error message to explain that a particular field is only required because this is a `saml` or `oidc` mapper as opposed to re-using the Terraform core error string for required fields?  This will make it more clear to the user that, while an attribute is not marked as `Required`, it is still required because of the provider id.

If you're on board with this, then I am fine with you just leaving it as-is as long as you're okay with me updating this in the future.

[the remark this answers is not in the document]
AndrewChubatiuk: Done

example/main.tf
```diff
@@ -343,3 +343,34 @@ resource "keycloak_saml_user_property_protocol_mapper" "saml_user_property_mappe
...
+	attribute_name = "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress"
+	identity_provider_alias = "${keycloak_saml_identity_provider.adfs.alias}"
+	user_attribute = "email"
+}
+
+resource keycloak_oidc_user_attribute_idp_mapper email {
+	realm = "${keycloak_realm.test.id}"
+	name = "Attribute: email"
+	claim_name = "upn"
+	identity_provider_alias = "${keycloak_oidc_identity_provider.adfs.alias}"
```
mrparkers: This reference is invalid - it doesn't point to another resource within this tf file

example/main.tf
```diff
@@ -343,3 +343,34 @@ resource "keycloak_saml_user_property_protocol_mapper" "saml_user_property_mappe
...
+
+resource keycloak_saml_user_attribute_idp_mapper email {
+	realm = "${keycloak_realm.test.id}"
+	name = "Attribute: email"
+	attribute_name = "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress"
+	identity_provider_alias = "${keycloak_saml_identity_provider.adfs.alias}"
+	user_attribute = "email"
+}
+
+resource keycloak_oidc_user_attribute_idp_mapper email {
```
mrparkers: It doesn't look like this resource is defined within `provider.go` - was this renamed to something else after this was committed?

example/main.tf
```diff
@@ -343,3 +343,34 @@ resource "keycloak_saml_user_property_protocol_mapper" "saml_user_property_mappe
...
+	client_secret = "example_token"
+}
+
+resource keycloak_saml_identity_provider adfs {
+	realm = "${keycloak_realm.test.id}"
+	alias = "adfs"
+	single_sign_on_service_url = "https://example.com/auth"
+}
+
+resource keycloak_saml_user_attribute_idp_mapper email {
```
mrparkers: It doesn't look like this resource is defined within `provider.go` - was this renamed to something else after this was committed?

AndrewChubatiuk: Forgot to update example.tf after I updated naming of idp mapper resources. Just pushed my changes

provider/keycloak_hardcoded_role_identity_provider_mapper.go
```diff
@@ -0,0 +1,37 @@
...
+	genericMapperResource.Schema = mergeSchemas(genericMapperResource.Schema, mapperSchema)
+	genericMapperResource.Create = resourceKeycloakIdentityProviderMapperCreate(getHardcodedRoleIdentityProviderMapperFromData, setHardcodedRoleIdentityProviderMapperData)
+	genericMapperResource.Read = resourceKeycloakIdentityProviderMapperRead(setHardcodedRoleIdentityProviderMapperData)
+	genericMapperResource.Update = resourceKeycloakIdentityProviderMapperUpdate(getHardcodedRoleIdentityProviderMapperFromData, setHardcodedRoleIdentityProviderMapperData)
+	return genericMapperResource
+}
+
+func getHardcodedRoleIdentityProviderMapperFromData(data *schema.ResourceData, _ interface{}) (*keycloak.IdentityProviderMapper, error) {
+	rec, _ := getIdentityProviderMapperFromData(data)
+	rec.IdentityProviderMapper = "oidc-hardcoded-role-idp-mapper"
```
mrparkers: Can this resource not be used for SAML?

AndrewChubatiuk: This resource is also used for SAML, but looks like it's an issue in Keycloak API

AndrewChubatiuk: You can try to create this mapper manually

PR #106, opened 2019-05-06
Keycloak v6.0.1

AndrewChubatiuk:


provider/resource_keycloak_openid_client_optional_scopes_test.go
```diff
@@ -602,6 +605,7 @@ resource "keycloak_openid_client_optional_scopes" "optional_scopes" {
 		"address",
 		"phone",
 		"offline_access",
+                "microprofile-jwt",
```
mrparkers: Can you fix the indentation for each instance of this string?

AndrewChubatiuk: Done

mrparkers: Thanks!

PR #107, opened 2019-05-08
Upgrade to TF 0.12

ctrox:
This upgrades the vendored TF version to `v0.12.0-rc1` to make the provider compatible with Terraform 0.12.

Two acceptance tests still fail and I'm not sure how to handle them:
```
--- FAIL: TestAccKeycloakOpenIdAudienceProtocolMapper_validateClientConflictsWithClientScope (0.72s)
    testing.go:549: Step 0, no error received, but expected a match to:

        .+ conflicts with .+


=== RUN   TestAccKeycloakOpenIdAudienceProtocolMapper_validateClientAudienceConflictsWithCustomAudience
--- FAIL: TestAccKeycloakOpenIdAudienceProtocolMapper_validateClientAudienceConflictsWithCustomAudience (0.72s)
    testing.go:549: Step 0, no error received, but expected a match to:

        .+ conflicts with .+
```
It looks like behaviour of the `ConflictsWith` [has changed if the values are unknown](https://github.com/hashicorp/terraform/commit/06acc3f6c82ca30ec6aa721f810a454a97c91e28). So with Terraform 0.12 the "conflicts with x" error will only appear if the values are known so for example this will not result in an error:

```terraform
resource "keycloak_openid_audience_protocol_mapper" "audience_mapper" {
  name            = "foo"
  realm_id        = "${keycloak_realm.realm.id}"
  client_id       = "${keycloak_openid_client.openid_client.id}"
  client_scope_id = "${keycloak_openid_client_scope.client_scope.id}"

  included_custom_audience = "foo"
}
```

But this will:

```terraform
resource "keycloak_openid_audience_protocol_mapper" "audience_mapper" {
  name            = "foo"
  realm_id        = "bar"
  client_id       = "bar"
  client_scope_id = "${keycloak_openid_client_scope.client_scope.id}"

  included_custom_audience = "foo"
}
```

I'm actually not sure if this is a bug in TF 0.12 since it also does not catch the `ConflictsWith` once the values are known (during apply and subsequent plans).


mrparkers:
Hi @ctrox, thanks for the PR!  I agree that it seems like a bug, or at least a major change in the behavior of `ConflictsWith`.  I don't think this will be hard to get around though - the audience protocol mapper has a [validation function](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/keycloak/openid_audience_protocol_mapper.go#L97) where this logic can be moved to.

Let me know if you need some assistance.  This has been on my todo list for a while, so as soon as this is merged I will cut a release for you :smile:

ctrox:
@mrparkers Thanks for the pointers. I added two additional validations, is this what you meant? Maybe I will still open an issue in the TF repo about this as i think the `ConflictsWith` should still work with computed values during apply.

mrparkers:
Looks perfect!  I can merge this as soon as the merge conflicts are fixed.

I also agree that opening an issue in the Terraform core repo would be a good idea.  It would be nice to know if they intend on restoring the original behavior of `ConflictsWith`.

Thanks again for the PR!

ctrox:
Great, thanks! I rebased on master.

PR #113, opened 2019-05-23
[WIP] Add hostname attribute to realm

alexashley:
This attribute configures the [fixed hostname provider](https://www.keycloak.org/docs/latest/server_admin/index.html#fixed-provider); it controls the issuer claim and any other generated URLs. Out of the box, Keycloak uses the `request` hostname provider, which reads the `Host` header in order to set the issuer. 

From running the example, this PR seems to work but I'd like to update the config for an acceptance test. Also, this needs a feature check, since Keycloak is happy to accept changes to hostname even when the fixed provider isn't being used. All of the providers are part of `serverinfo`, so I'm planning to add to that struct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ https://mrparkers.github.io/terraform-provider-keycloak/
 ## Building
 
 This project uses [Go Modules](https://github.com/golang/go/wiki/Modules) which requires Go 1.11.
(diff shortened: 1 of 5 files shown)


tomrutsaert:
This PR might be obsolete, as since keycloak 8+ the _fixed_ and _request_ hostname provider are replaced with a new default hostname provider and frontend url variable on realm level

PR #114, opened 2019-05-23
Allow for using the password grant with a confidential client

alexashley:
The current implementation only allows direct access grants with a public client.


keycloak/keycloak_client.go
```diff
@@ -43,13 +43,13 @@ func NewKeycloakClient(baseUrl, clientId, clientSecret, realm, username, passwor
 	}
 	clientCredentials := &ClientCredentials{
 		ClientId: clientId,
+		ClientSecret: clientSecret,
```
alexashley: The zero value for a string is the empty string, so this shouldn't affect public clients.

PR #122, opened 2019-06-12
adding possibility to add SMTP configuration

tomrutsaert:
Adding possibility to add SMTP Server configuration.

### DIFF
--- a/keycloak/realm.go
+++ b/keycloak/realm.go
@@ -20,6 +20,9 @@ type Realm struct {
 	LoginWithEmailAllowed       bool `json:"loginWithEmailAllowed"`
 	DuplicateEmailsAllowed      bool `json:"duplicateEmailsAllowed"`
 
+	//SMTP Server
+	SmtpServer SmtpServer `json:"smtpServer"`
+
 	// Themes
 	LoginTheme   string `json:"loginTheme,omitempty"`
 	AccountTheme string `json:"accountTheme,omitempty"`
@@ -40,6 +43,21 @@ type Realm struct {
 	AccessCodeLifespanUserAction        int  `json:"accessCodeLifespanUserAction,omitempty"`
 	ActionTokenGeneratedByUserLifespan  int  `json:"actionTokenGeneratedByUserLifespan,omitempty"`
 	ActionTokenGeneratedByAdminLifespan int  `json:"actionTokenGeneratedByAdminLifespan,omitempty"`
+}
+
+type SmtpServer struct {
+	StartTls           string `json:"starttls,omitempty"`
+	Auth               string `json:"auth,omitempty"`
+	Port               string `json:"port,omitempty"`
+	Host               string `json:"host,omitempty"`
+	ReplyTo            string `json:"replyTo,omitempty"`
+	ReplyToDisplayName string `json:"replyToDisplayName,omitempty"`
+	From               string `json:"from,omitempty"`
+	FromDisplayName    string `json:"fromDisplayName,omitempty"`
+	EnvelopeFrom       string `json:"envelopeFrom,omitempty"`
+	Ssl                string `json:"ssl,omitempty"`
+	User               string `json:"user,omitempty"`
+	Password           string `json:"password,omitempty"`
 }
 
 func (keycloakClient *KeycloakClient) NewRealm(realm *Realm) error {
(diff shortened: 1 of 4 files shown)


keycloak/realm.go
```diff
@@ -92,6 +110,14 @@ func (keycloakClient *KeycloakClient) ValidateRealm(realm *Realm) error {
 		return err
 	}
 
+	if (SmtpServer{}) != realm.SmtpServer && realm.SmtpServer.Host == "" {
```
BernhardBerbuir: is there a reason why this is not checked by Terraform?
This attribute is configured as `Optional: true,`
=> could it be configured as`Required: true,`?

tomrutsaert: Well it is an optional field in the sense that you do need to set smtp server, but the moment you set one field of the smtp server, this is one of two required fields. (host and from).
For that reason, I do not think I can make the field required, can I?
Not really a Golang orTerraform expert, But I will take a look tomorrow, what it gives when I make this field required but not set the smtpserver

tomrutsaert: changed it

provider/resource_keycloak_realm.go
```diff
@@ -73,6 +73,70 @@ func resourceKeycloakRealm() *schema.Resource {
...
+						},
+						"user": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"password": {
+							Type:      schema.TypeString,
+							Optional:  true,
+							Sensitive: true,
+							DiffSuppressFunc: func(_, remoteBindCredential, _ string, _ *schema.ResourceData) bool {
```
mrparkers: Could you rename `remoteBindCredential`?  I think this was copied from the ldap user federation resource where this particular attribute was called `bind_credential`.

provider/resource_keycloak_realm_test.go
```diff
@@ -429,6 +488,76 @@ resource "keycloak_realm" "realm" {
...
+	return fmt.Sprintf(`
+resource "keycloak_realm" "realm" {
+	realm        = "%s"
+	enabled      = true
+	display_name = "%s"
+	smtp_server {
+      host = "%s"
+	  port = 25
+	  from_display_name = "Tom"
+      from = "%s"
```
mrparkers: would you be able to fix the indentation for these SMTP config blocks?

provider/resource_keycloak_realm_test.go
```diff
@@ -86,6 +86,46 @@ func TestAccKeycloakRealm_import(t *testing.T) {
...
+				Config:      testKeycloakRealm_WithSmtpServerWithoutHost(realm, "My Host"),
+				ExpectError: regexp.MustCompile("config is invalid: Missing required argument: The argument \"host\" is required, but no definition was found."),
+			},
+			{
+				Config:      testKeycloakRealm_WithSmtpServerWithoutFrom(realm, "myhost.com"),
+				ExpectError: regexp.MustCompile("config is invalid: Missing required argument: The argument \"from\" is required, but no definition was found."),
+			},
+		},
+	})
+}
```
mrparkers: thanks for adding these tests!  would you be able to add one more that creates a realm with some SMTP settings and updates those settings in a second test step?

keycloak/realm.go
mrparkers: thanks for reviewing this @BernhardBerbuir!  I would have left the same comment :smile:

provider/resource_keycloak_realm.go
tomrutsaert: ok

provider/resource_keycloak_realm_test.go
tomrutsaert: strange, I do not have this weird indentation, I will copy and paste the whole block again

provider/resource_keycloak_realm_test.go
tomrutsaert: Yes, I will do that.

tomrutsaert:
All request changes were added + I have updated the branch to latest version of master

mrparkers:
@tomrutsaert I just noticed this after merging, but what would you think of handling SMTP auth like this instead?

```hcl
  realm        = "test"

  smtp_server {
    host = "mysmtphost.com"
    from = "tom@myhost.com"
    auth {
      user = "tom"
      password = "tom"
    }
  }
}
```

If you're okay with this, I'll push up a quick fix and cut a release for you within the next few days.

tomrutsaert:
Seems like a good idea, go ahead

PR #124, opened 2019-06-13
Realm internationalization

tomrutsaert:


keycloak/realm.go
```diff
@@ -108,5 +113,34 @@ func (keycloakClient *KeycloakClient) ValidateRealm(realm *Realm) error {
 		return fmt.Errorf("validation error: theme \"%s\" does not exist on the server", realm.EmailTheme)
 	}
 
+	if realm.InternationalizationEnabled == false && len(realm.SupportLocales) > 0 {
+		return fmt.Errorf("validation error: SupportLocales cannot be set if InternationalizationEnabled is false")
+	}
+
+	if realm.InternationalizationEnabled == false && len(realm.DefaultLocale) > 0 {
+		return fmt.Errorf("validation error: DefaultLocale cannot be set if InternationalizationEnabled is false")
+	}
```
mrparkers: Are these two error messages still needed?  Based on the provider code, it looks like these simply can't be set without specifying the `internationalization {}` set.

keycloak/realm.go
```diff
@@ -108,5 +113,34 @@ func (keycloakClient *KeycloakClient) ValidateRealm(realm *Realm) error {
...
+
+	if realm.InternationalizationEnabled == true && len(realm.SupportLocales) <= 0 {
+		return fmt.Errorf("validation error: SupportLocales should be set if Internationalization is present")
+	}
+
+	if realm.InternationalizationEnabled == true && len(realm.DefaultLocale) <= 0 {
+		return fmt.Errorf("validation error: DefaultLocale should be set if Internationalization is present")
+	}
```
mrparkers: These two error messages could probably be removed if these attributes were marked as `Required` in the `internationalization` set.

provider/resource_keycloak_realm_test.go
```diff
@@ -150,6 +150,66 @@ func TestAccKeycloakRealm_themesValidation(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		Providers:    testAccProviders,
+		PreCheck:     func() { testAccPreCheck(t) },
+		CheckDestroy: testAccCheckKeycloakRealmDestroy(),
+		Steps: []resource.TestStep{
+			{
+				Config: testKeycloakRealm_internationalizationValidation(realm, "en", "en"),
+				Check:  testAccCheckKeycloakRealmInternationalizationIsEnabled("keycloak_realm.realm", "en"),
+			},
```
mrparkers: Could you add another test step after this one that updates the internationalization settings to something different?

keycloak/realm.go
tomrutsaert: I have removed these two.

keycloak/realm.go
tomrutsaert: I made them required and I have removed these two messages as well.

provider/resource_keycloak_realm_test.go
tomrutsaert: added an extra step that changes the default locale to es

tomrutsaert:
@mrparkers I do not see ant reason why test-password ci failed or how it is related to my changes.

provider/resource_keycloak_realm.go
```diff
@@ -229,11 +229,47 @@ func resourceKeycloakRealm() *schema.Resource {
 				Computed:         true,
 				DiffSuppressFunc: suppressDurationStringDiff,
 			},
+
+			//internationalization
+			"internationalization": {
+				Type:     schema.TypeSet,
```
mrparkers: Could you use `TypeList` like I did in #128?

tomrutsaert: done

PR #126, opened 2019-06-18
implement a data source for realm keys

BernhardBerbuir:
- add new ressource keycloak_realm_keys
- add GO struct for storing realm keys
- add filter for status and algorithm

This is a first version of the implemention
=> Feedback is welcome
I'm currently creating documentaion and test cases

### DIFF
--- a/docs/resources/datasource_keycloak_realm_keys.md
+++ b/docs/resources/datasource_keycloak_realm_keys.md
@@ -1,0 +1,38 @@
+# datasource keycloak_realm_keys
+
+Use this data source to get the keys of a realm. Keys can be filtered by algorithm and status.
+
+Remarks:
+
+- A key must meet all filter criteria
+- This datasource may return more than one value.
+- If no key matches the filter criteria, then an error is returned.
+
+### Example Usage
+
+```hcl
+resource "keycloak_realm" "realm" {
+    realm   = "my-realm"
+    enabled = true
+}
+
+data "keycloak_realm_keys" "keys" {
+  realm_id  = keycloak_realm.realm
+  algorithms = ["AES", "RS256"]
+  status = ["ACTIVE", "PASSIVE"]
+}
+
+# show certificate of first key:
+output "certificate" {
+  value = data.keycloak_realm_keys.realm.keys[0].certificate
+}
+
+```
+
+### Argument Reference
+
+The following arguments are supported:
+
+- `realm_id` - (Required) The realm of which the keys are retrieved.
+- `algorithms` - (Optional) When specified, keys are filtered by algorithm (values for algorithm: `HS256`, `RS256`,`AES`, ...)
+- `status` - (Optional) When specified, keys are filtered by status (values for status: `ACTIVE`, `DISABLED` and `PASSIVE`)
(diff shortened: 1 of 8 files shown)


docs/resources/datasource_keycloak_realm_keys.md
```diff
@@ -0,0 +1,38 @@
+# datasource keycloak_realm_keys
+
+Use this data source to get the keys of a realm. Keys can be filtered by algorithm and status.
+
+Remarks:
+
+- A key must meet all filter criteria
+- This datasource may return more than one value.
```
mrparkers: Is the behavior of the Keycloak API returning these keys defined?  I'd like to document what users should expect as the first, second, third, etc value of this.  Or maybe we could consider a data source that will only return a single key based on the filter logic?  As-is, I think this may be difficult to use, as you'd have to try each index before finding the one you care about.

I'm interested to hear your thoughts on this.

provider/data_source_keycloak_realm_keys.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	return keyMap
+}
+
+func setRealmKeysData(data *schema.ResourceData, keys *keycloak.Keys) error {
+	data.SetId(data.Get("realm_id").(string))
+
+	err := data.Set("keys", flattenRealmKeys(keys.Keys))
+	if err != nil {
+		return fmt.Errorf("could set keys with values '%+v'\n%+v", keys.Keys, err)
```
mrparkers: this error message is a bit unclear to me

BernhardBerbuir: I fixed the error message. Normally `data.Set()`should not fail, but when the data (i.e result of `flattenRealmKeys()`) has the wrong structure than an error might occur.

docs/resources/datasource_keycloak_realm_keys.md
BernhardBerbuir: The Keycloak API for [realm keys](https://www.keycloak.org/docs-api/6.0/rest-api/index.html#_keysmetadatarepresentation) does not guarantee the order. It is possible to have multiple keys with the same algorithm and status (in this case you should assign different priorities).
The datasource _could_ sort the keys by priority, but you can have several keys with the same priority.

Providing more than one key is useful when:
- you want to output all keys for debug purposes (then you can adjust the filter accordingly)
- you want to iterate over all keys in Terraform (could be helpful when using templates)

My use case: I just need the certificate of one key which can be identified by status and algorithm (the realm administrator takes care that there is only one key).

When the datasource has to guarantee that there is always just one key, then
- it has to be renamed to `keycloak_realm_key` (without `s`)
- it has to throw an error when the filter criteria return more than one key

The Terraform provider for AWS has two datasources:
- [aws_availability_zone](https://www.terraform.io/docs/providers/aws/d/availability_zone.html) - returns just *one* AZ
- [aws_availability_zones](https://www.terraform.io/docs/providers/aws/d/availability_zones.html) returns *one or more* AZs

I personally prefer the multi value option. In this case I have all data and can later on decide what to do. When I just want one value then I can apply the correct filter.

Nonetheless I can think of the following enhancement:
- additional attribute `only_highest_priority` => only the key with the highest priority is returned
- additional attribute `error_when_multiple_keys` => an error is retuned when there is more than one key

(we still need a better name for these attributes :smile:)
The result of the datasource wil be still a list => access by index `0`is required

mrparkers: If the data source is useful for you as-is then I have no problem with the current implementation.  Perhaps we could add the complimentary `keycloak_realm_key` data source at a later time.

PR #128, opened 2019-06-18
update SMTP auth config attribute for keycloak_realm resource

mrparkers:
cc @tomrutsaert

### DIFF
--- a/docs/resources/keycloak_realm.md
+++ b/docs/resources/keycloak_realm.md
@@ -26,6 +26,15 @@ resource "keycloak_realm" "realm" {
         	password = "password"
         }
     }
+    
+    internationalization {
+        supported_locales = [
+            "en",
+            "de",
+            "es"
+        ]
+        default_locale    = "en"
+    }
 }
 ```
 
@@ -98,6 +107,13 @@ This block supports the following attributes:
 - `auth` - (Optional) Enables authentication to the SMTP server.  This block supports the following attributes:
     - `username`- (Required) The SMTP server username.
     - `password` - (Required) The SMTP server password.
+    
+##### Internationalization
+
+Internationalization support can be configured by using the `internationalization` block, which supports the following attributes:
+
+- `supported_locales` - (Required) A list of [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) locale codes that the realm should support.
+- `default_locale` - (Required) The locale to use by default. This locale code must be present within the `supported_locales` list.
 
 ### Import
 
--- a/docs/resources/keycloak_realm.md
+++ b/docs/resources/keycloak_realm.md
@@ -16,6 +16,16 @@ resource "keycloak_realm" "realm" {
     login_theme          = "base"
 
     access_code_lifespan = "1h"
+    
+    smtp_server {
+        host = "smtp.example.com"
+        from = "example@example.com"
+        
+        auth {
+        	username = "tom"
+        	password = "password"
+        }
+    }
 }
 ```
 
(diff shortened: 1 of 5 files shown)


keycloak/realm.go
```diff
@@ -46,18 +46,18 @@ type Realm struct {
...
-	Host               string `json:"host,omitempty"`
-	ReplyTo            string `json:"replyTo,omitempty"`
-	ReplyToDisplayName string `json:"replyToDisplayName,omitempty"`
-	From               string `json:"from,omitempty"`
-	FromDisplayName    string `json:"fromDisplayName,omitempty"`
-	EnvelopeFrom       string `json:"envelopeFrom,omitempty"`
-	Ssl                string `json:"ssl,omitempty"`
-	User               string `json:"user,omitempty"`
-	Password           string `json:"password,omitempty"`
+	StartTls           KeycloakBoolQuoted `json:"starttls,omitempty"`
```
mrparkers: This is a special type that serializes to a JSON string of `"true"` or `"false"`, which is helpful for parts of the Keycloak API that expect boolean strings.  Under the hood it's a regular boolean that is easier to work with in the provider code.

provider/resource_keycloak_realm.go
```diff
@@ -76,17 +76,13 @@ func resourceKeycloakRealm() *schema.Resource {
 			//Smtp server
 
 			"smtp_server": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
```
mrparkers: This won't change how this attribute is used, but it's much easier to use `TypeList` for one-off block attributes.

provider/resource_keycloak_realm.go
```diff
@@ -425,9 +452,17 @@ func setRealmData(data *schema.ResourceData, realm *keycloak.Realm) {
...
+		if realm.SmtpServer.Auth {
+			auth := make(map[string]interface{})
+
+			auth["username"] = realm.SmtpServer.User
+			auth["password"] = realm.SmtpServer.Password
+
+			smtpSettings["auth"] = []interface{}{auth}
+		}
+
+		data.Set("smtp_server", []interface{}{smtpSettings})
```
mrparkers: This line is why `TypeList` is easier to work with than `TypeString` in my opinion.

keycloak/realm.go
tomrutsaert: Ok, good to know, I will use that in new development

provider/resource_keycloak_realm.go
tomrutsaert: ok

provider/resource_keycloak_realm.go
tomrutsaert: I will be using this.

docs/resources/keycloak_realm.md
```diff
@@ -71,6 +81,24 @@ The attributes below should be specified as [Go duration strings](https://golang
...
 - `action_token_generated_by_admin_lifespan` - (Optional) The maximum time a user has to use an admin-generated permit before it expires.
 
+##### SMTP
+
+The `smtp_server` block can be used to configure the realm's SMTP settings, which can be found in the "Email" tab in the GUI.
+This block supports the following attributes:
+
+- `host` - (Required) The host of the SMTP server.
+- `port` - (Optional) The port of the SMTP server (defaults to 25).
+- `from` - (Optional) The email address for the sender.
```
tomrutsaert: Isn't "from" a required field?

mrparkers: Good catch!

PR #132, opened 2019-07-05
fix for user attribute value with length bigger than 255 symbols

AndrewChubatiuk:
Hello again, @mrparkers
Prepared a small PR for a case when user attribute value is bigger than 255 symbols.
Keycloak web automatically splits too long strings into a slice with a string length up to 255 symbols.
My PR does the same

### DIFF
--- a/docs/resources/datasource_keycloak_realm_keys.md
+++ b/docs/resources/datasource_keycloak_realm_keys.md
@@ -1,0 +1,38 @@
+# datasource keycloak_realm_keys
+
+Use this data source to get the keys of a realm. Keys can be filtered by algorithm and status.
+
+Remarks:
+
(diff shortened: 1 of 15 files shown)


AndrewChubatiuk:
PR is ready, but build is failing from time to time

mrparkers:
Thanks for the PR, would you be able to add a test for the scenario you're fixing?

AndrewChubatiuk:
Yep. Just added

AndrewChubatiuk:
@mrparkers Yep. Just added

AndrewChubatiuk:
@tomrutsaert 
What platform do you use?

tomrutsaert:
Ubuntu 18.04.2 LTS

PR #134, opened 2019-07-11
allow extra config for a custom user federation implementation

tomrutsaert:
allow extra config for a custom user federation implementation.

I had some troubles writing a good test, as keycloak only returns custom config keys and values that the custom implementation understands. I used the written tests on a keycloak installation with my custom user federation spi, and everything worked as expected.
Any suggestions how to make better tests for this type of configuration?

### DIFF
--- a/custom-user-federation-example/src/main/kotlin/com/github/mrparkers/keycloak/CustomUserStorageProviderFactory.kt
+++ b/custom-user-federation-example/src/main/kotlin/com/github/mrparkers/keycloak/CustomUserStorageProviderFactory.kt
@@ -2,15 +2,32 @@ package com.github.mrparkers.keycloak
 
 import org.keycloak.component.ComponentModel
 import org.keycloak.models.KeycloakSession
+import org.keycloak.provider.ProviderConfigProperty
 import org.keycloak.storage.UserStorageProviderFactory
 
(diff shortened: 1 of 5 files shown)


mrparkers:
Yeah this is tricky.  In its current state, anyone else who runs the tests locally (using the built user federation provider example) will see this test fail, since the example doesn't have any config properties.

If you'd like to put in the effort, it would be great if you could update the example provider to include some dummy config value.  It looks like it doesn't have to be used, it just has to exist in order for the test to do what you want it to.

Otherwise, you can remove the test.  This isn't running in CI anyways so I don't feel too bad about missing that test.

tomrutsaert:
> Yeah this is tricky. In its current state, anyone else who runs the tests locally (using the built user federation provider example) will see this test fail, since the example doesn't have any config properties.
> 
> If you'd like to put in the effort, it would be great if you could update the example provider to include some dummy config value. It looks like it doesn't have to be used, it just has to exist in order for the test to do what you want it to.
> 
> Otherwise, you can remove the test. This isn't running in CI anyways so I don't feel too bad about missing that test.

I have extended the custom-user-federation-example to have a dummyConfig key.
My test is using that dummyConfig key, and everything works as expected.

People who want the run the tests locally, will of course need to rebuild the custom-user-federation-example with ```./gradlew shadowJar```

PR #137, opened 2019-07-18
custom config for custom oidc identity provider with test and example

tomrutsaert:
see also: https://github.com/mrparkers/terraform-provider-keycloak/issues/135

### DIFF
--- a/keycloak/identity_provider.go
+++ b/keycloak/identity_provider.go
@@ -1,40 +1,45 @@
 package keycloak
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
+	"reflect"
+	"strconv"
+	"strings"
 )
 
 type IdentityProviderConfig struct {
-	Key                              string             `json:"key,omitempty"`
-	HostIp                           string             `json:"hostIp,omitempty"`
-	UseJwksUrl                       KeycloakBoolQuoted `json:"useJwksUrl,omitempty"`
-	JwksUrl                          string             `json:"jwksUrl,omitempty"`
-	ClientId                         string             `json:"clientId,omitempty"`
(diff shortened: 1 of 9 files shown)


keycloak/identity_provider.go
```diff
@@ -54,6 +59,60 @@ type IdentityProvider struct {
...
+	v := reflect.ValueOf(f).Elem()
+	for i := 0; i < v.NumField(); i++ {
+		jsonKey := strings.Split(v.Type().Field(i).Tag.Get("json"), ",")[0]
+		if jsonKey != "-" {
+			field := v.Field(i)
+			if field.IsValid() && field.CanSet() {
+				if field.Kind() == reflect.String {
+					out[jsonKey] = field.String()
+				} else if field.Kind() == reflect.Bool {
+					out[jsonKey] = KeycloakBoolQuoted(field.Bool())
```
mrparkers: I'm a little confused here - are you expecting that all custom boolean config values will be json strings?  I know some Keycloak APIs do this but I'm curious why a custom identity provider would also do this.  Could you clear this up for me?

tomrutsaert: Idp config maps (and most other config maps) are always a map of string,string
see: https://github.com/keycloak/keycloak/blob/master/core/src/main/java/org/keycloak/representations/idm/IdentityProviderRepresentation.java#L59
so yes this will always be a string,

mrparkers: Thanks for the explanation!

tomrutsaert:
@mrparkers, Do I still need to do anything for this PR?

mrparkers:
Sorry, I forgot to come back to this after merging master.

PR #138, opened 2019-07-26
Authentication flows and executions

tomrutsaert:
First attempt in adding support for authentication-flows and executions.

It supports nested flows etc.

TODO: find a solution or ordering of subflows and executions with raise and lower priority api calls only. 
I was thinking of making a execution/subflow depend on the execution/subflow they are behind.
```
resource "keycloak_authentication_flow" "browser-copy-flow" {
	alias    = "browserCopyFlow"
	realm_id = "${keycloak_realm.test.id}"
	description = "browser based authentication"
}

resource "keycloak_authentication_execution" "browser-copy-cookie" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
	authenticator = "auth-cookie"
	requirement = "ALTERNATIVE"
}

resource "keycloak_authentication_execution" "browser-copy-kerberos" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
	authenticator = "auth-spnego"
	requirement = "DISABLED"
        isBehind = "${keycloak_authentication_execution.browser-copy-cookie.id}"
}
```
**isBehind = "${keycloak_authentication_execution.browser-copy-cookie.id}"**

This way you might lookup the priority of execution/subflow it depends on, and the raise or lower the priority  as needed
I am open for suggestions.....

### DIFF
--- a/example/main.tf
+++ b/example/main.tf
@@ -52,41 +52,6 @@ resource "keycloak_realm" "test" {
 	}
 }
 
-resource "keycloak_authentication_flow" "myflow" {
-	alias    = "myTestAliasFlow"
-	realm_id = "${keycloak_realm.test.id}"
-
-}
-
-resource "keycloak_authentication_subflow" "subflow" {
-	realm_id = "${keycloak_realm.test.id}"
-	parent_flow_alias = "${keycloak_authentication_flow.myflow.alias}"
-	alias    = "subTest"
-	requirement = "REQUIRED"
-}
-
-resource "keycloak_authentication_subflow" "subsubflow" {
-	realm_id = "${keycloak_realm.test.id}"
-	parent_flow_alias = "${keycloak_authentication_subflow.subflow.alias}"
-	alias    = "subsubTest"
-	requirement = "REQUIRED"
-}
-
-resource "keycloak_authentication_execution" "execution" {
-	realm_id = "${keycloak_realm.test.id}"
-	parent_flow_alias = "${keycloak_authentication_subflow.subsubflow.alias}"
-	authenticator = "auth-cookie"
-	requirement = "ALTERNATIVE"
-}
-
-resource "keycloak_authentication_subflow" "subflow2" {
-	realm_id = "${keycloak_realm.test.id}"
-	parent_flow_alias = "${keycloak_authentication_flow.myflow.alias}"
-	alias    = "subTest3"
-	requirement = "REQUIRED"
-	provider_id = "form-flow"
-}
-
 resource "keycloak_authentication_flow" "browser-copy-flow" {
 	alias    = "browserCopyFlow"
 	realm_id = "${keycloak_realm.test.id}"
@@ -98,6 +63,7 @@ resource "keycloak_authentication_execution" "browser-copy-cookie" {
 	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
 	authenticator = "auth-cookie"
 	requirement = "ALTERNATIVE"
+	depends_on = ["keycloak_authentication_execution.browser-copy-kerberos"]
 }
 
 resource "keycloak_authentication_execution" "browser-copy-kerberos" {
@@ -112,6 +78,7 @@ resource "keycloak_authentication_execution" "browser-copy-idp-redirect" {
 	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
 	authenticator = "identity-provider-redirector"
 	requirement = "ALTERNATIVE"
+	depends_on = ["keycloak_authentication_execution.browser-copy-cookie"]
 }
 
 resource "keycloak_authentication_subflow" "browser-copy-flow-forms" {
@@ -119,6 +86,7 @@ resource "keycloak_authentication_subflow" "browser-copy-flow-forms" {
 	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
 	alias    = "browser-copy-flow-forms"
 	requirement = "ALTERNATIVE"
+	depends_on = ["keycloak_authentication_execution.browser-copy-idp-redirect"]
 }
 
 resource "keycloak_authentication_execution" "browser-copy-auth-username-password-form" {
@@ -133,6 +101,7 @@ resource "keycloak_authentication_execution" "browser-copy-otp" {
 	parent_flow_alias = "${keycloak_authentication_subflow.browser-copy-flow-forms.alias}"
 	authenticator = "auth-otp-form"
 	requirement = "OPTIONAL"
+	depends_on = ["keycloak_authentication_execution.browser-copy-auth-username-password-form"]
 }
 
 
--- a/example/main.tf
+++ b/example/main.tf
@@ -135,530 +135,530 @@ resource "keycloak_authentication_execution" "browser-copy-otp" {
 	requirement = "OPTIONAL"
 }
 
-//
-//resource "keycloak_required_action" "custom-terms-and-conditions" {
-//	realm_id		= "${keycloak_realm.test.realm}"
-//	alias			= "terms_and_conditions"
-//	default_action 	= true
-//	enabled			= true
-//	name			= "Custom Terms and Conditions"
-//}
-//
-//resource "keycloak_required_action" "custom-configured_totp" {
-//	realm_id		= "${keycloak_realm.test.realm}"
-//	alias			= "CONFIGURE_TOTP"
-//	default_action 	= true
-//	enabled			= true
-//	name			= "Custom configure totp"
-//	priority		= "${keycloak_required_action.custom-terms-and-conditions.priority+15}"
-//}
-//
-//resource "keycloak_group" "foo" {
-//	realm_id = "${keycloak_realm.test.id}"
-//	name     = "foo"
-//}
 
-//resource "keycloak_group" "nested_foo" {
-//	realm_id  = "${keycloak_realm.test.id}"
-//	parent_id = "${keycloak_group.foo.id}"
-//	name      = "nested-foo"
-//}
-//
-//resource "keycloak_group" "bar" {
-//	realm_id = "${keycloak_realm.test.id}"
-//	name     = "bar"
-//}
-//
-//resource "keycloak_user" "user" {
-//	realm_id = "${keycloak_realm.test.id}"
-//	username = "test-user"
-//
-//	email      = "test-user@fakedomain.com"
-//	first_name = "Testy"
-//	last_name  = "Tester"
-//}
-//
-//resource "keycloak_user" "another_user" {
-//	realm_id = "${keycloak_realm.test.id}"
-//	username = "another-test-user"
-//
-//	email      = "another-test-user@fakedomain.com"
-//	first_name = "Testy"
-//	last_name  = "Tester"
-//}
-//
-//resource "keycloak_user" "user_with_password" {
-//	realm_id = "${keycloak_realm.test.id}"
-//	username = "user-with-password"
-//
-//	email      = "user-with-password@fakedomain.com"
-//	first_name = "Testy"
-//	last_name  = "Tester"
-//
-//	initial_password {
-//		value     = "my password"
-//		temporary = false
-//	}
-//}
-//
-//resource "keycloak_group_memberships" "foo_members" {
-//	realm_id = "${keycloak_realm.test.id}"
-//	group_id = "${keycloak_group.foo.id}"
-//
-//	members = [
-//		"${keycloak_user.user.username}",
-//		"${keycloak_user.another_user.username}",
-//	]
-//}
-//
-//resource "keycloak_openid_client" "test_client" {
-//	client_id   = "test-openid-client"
-//	name        = "test-openid-client"
-//	realm_id    = "${keycloak_realm.test.id}"
-//	description = "a test openid client"
-//
-//	standard_flow_enabled = true
-//
-//	access_type = "CONFIDENTIAL"
-//
-//	valid_redirect_uris = [
-//		"http://localhost:5555/callback",
-//	]
-//
-//	client_secret = "secret"
-//}
-//
-//resource "keycloak_openid_client_scope" "test_default_client_scope" {
-//	name     = "test-default-client-scope"
-//	realm_id = "${keycloak_realm.test.id}"
-//
-//	description         = "test"
-//	consent_screen_text = "hello"
-//}
-//
-//resource "keycloak_openid_client_scope" "test_optional_client_scope" {
-//	name     = "test-optional-client-scope"
-//	realm_id = "${keycloak_realm.test.id}"
-//
-//	description         = "test"
-//	consent_screen_text = "hello"
-//}
-//
-//resource "keycloak_openid_client_default_scopes" "default_client_scopes" {
-//	realm_id  = "${keycloak_realm.test.id}"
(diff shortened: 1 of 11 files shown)


tomrutsaert:
> First attempt in adding support for authentication-flows and executions.
> 
> It supports nested flows etc.
> 
> TODO: find a solution or ordering of subflows and executions with raise and lower priority api calls only.
> I was thinking of making a execution/subflow depend on the execution/subflow they are behind.
> 
> ```
> resource "keycloak_authentication_flow" "browser-copy-flow" {
> 	alias    = "browserCopyFlow"
> 	realm_id = "${keycloak_realm.test.id}"
> 	description = "browser based authentication"
> }
> 
> resource "keycloak_authentication_execution" "browser-copy-cookie" {
> 	realm_id = "${keycloak_realm.test.id}"
> 	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
> 	authenticator = "auth-cookie"
> 	requirement = "ALTERNATIVE"
> }
> 
> resource "keycloak_authentication_execution" "browser-copy-kerberos" {
> 	realm_id = "${keycloak_realm.test.id}"
> 	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
> 	authenticator = "auth-spnego"
> 	requirement = "DISABLED"
>         isBehind = "${keycloak_authentication_execution.browser-copy-cookie.id}"
> }
> ```
> 
> **isBehind = "${keycloak_authentication_execution.browser-copy-cookie.id}"**
> 
> This way you might lookup the priority of execution/subflow it depends on, and the raise or lower the priority as needed
> I am open for suggestions.....

Perhaps it is better to work the same way we did with required actions. Ignore the raise and lower apis during creation, and just set the priority at creation.

the problem is during update.
raising the priority only switches the priority with the execution that has a higher priority. You can not set any number.

```
resource "keycloak_authentication_flow" "browser-copy-flow" {
	alias    = "browserCopyFlow"
	realm_id = "${keycloak_realm.test.id}"
	description = "browser based authentication"
}

resource "keycloak_authentication_execution" "browser-copy-cookie" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
	authenticator = "auth-cookie"
	requirement = "ALTERNATIVE" 
        priority = 10
}

resource "keycloak_authentication_execution" "browser-copy-kerberos" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
	authenticator = "auth-spnego"
	requirement = "DISABLED"
        priority = 20
}
```
changing priority from 20 to 15 on `browser-copy-kerberos` and in logic calling lower-priority-api would result in `browser-copy-kerberos` having priority 10 and `browser-copy-cookie` have priority 20, thus what people would expect.....

tomrutsaert:
>...
> 
> changing priority from 20 to 15 on `browser-copy-kerberos` and in logic calling lower-priority-api would result in `browser-copy-kerberos` having priority 10 and `browser-copy-cookie` have priority 20, thus what people would expect.....

Even deleting and re-creating will not work to set the priority, because the keyloak overrides what you set during creation.
I just do not see a decent solution.
The only thing that seems to work for complete new flows is using depends_on,this way the priorities are correct because everything was created in correct order.

example:
```
resource "keycloak_authentication_flow" "browser-copy-flow" {
	alias    = "browserCopyFlow"
	realm_id = "${keycloak_realm.test.id}"
	description = "browser based authentication"
}

resource "keycloak_authentication_execution" "browser-copy-cookie" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
	authenticator = "auth-cookie"
	requirement = "ALTERNATIVE"
	depends_on = ["keycloak_authentication_execution.browser-copy-kerberos"]
}

resource "keycloak_authentication_execution" "browser-copy-kerberos" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
	authenticator = "auth-spnego"
	requirement = "ALTERNATIVE"
}

resource "keycloak_authentication_execution" "browser-copy-idp-redirect" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
	authenticator = "identity-provider-redirector"
	requirement = "ALTERNATIVE"
	depends_on = ["keycloak_authentication_execution.browser-copy-cookie"]
}

resource "keycloak_authentication_subflow" "browser-copy-flow-forms" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_flow.browser-copy-flow.alias}"
	alias    = "browser-copy-flow-forms"
	requirement = "ALTERNATIVE"
	depends_on = ["keycloak_authentication_execution.browser-copy-idp-redirect"]
}

resource "keycloak_authentication_execution" "browser-copy-auth-username-password-form" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_subflow.browser-copy-flow-forms.alias}"
	authenticator = "auth-username-password-form"
	requirement = "REQUIRED"
}

resource "keycloak_authentication_execution" "browser-copy-otp" {
	realm_id = "${keycloak_realm.test.id}"
	parent_flow_alias = "${keycloak_authentication_subflow.browser-copy-flow-forms.alias}"
	authenticator = "auth-otp-form"
	requirement = "OPTIONAL"
	depends_on = ["keycloak_authentication_execution.browser-copy-auth-username-password-form"]
}
```

tomrutsaert:
This a best effort implementation.
I do not see another way.
I am thinking about creating a pr with the keycloak project to allow to set priority directly.
@mrparkers what do you think?

mrparkers:
Yeah, it sounds like you're coming to the same conclusions I came to when attempting to do this originally.

I think a PR upstream would definitely help if you are up for the challenge.

My other idea for representing this in HCL is to have a data source representing a subflow and its executions, which could be used in the top level flow to accurately represent all subflows and their order - only modifying a single resource.

It would kinda look like this:

```hcl
resource "keycloak_authentication_flow" "top_flow" {
    alias = "foo"
    
    execution { // TypeList, so the order of executions actually matters
        provider = "foo"
    }

    execution {
        provider = "bar"
    }

    subflow {
        json = "data.keycloak_authentication_subflow.subflow.json"
    }
}

data "keycloak_authentication_subflow" "subflow" {
    alias = "subflow"

    execution {
        provider = "subflow_foo"
    }

    subflow {
        json = "data.keycloak_authentication_subflow.second_subflow.json"
    }
}
```

The data sources here don't actually interact with the Keycloak API in any way, they just serialize the attributes of a subflow and its executions into JSON so the top level flow resource can read them.  This will allow you to specify an infinite number of subflows and their own executions, and the order of these would be controlled by the top level resource.

The downside of this approach is that the plan diff would look awful.

tomrutsaert:
> Yeah, it sounds like you're coming to the same conclusions I came to when attempting to do this originally.
>  ...
> 
> What are your thoughts on this?

I agree, your solution will probably be the better solution, there will be a bit more logic in interpreting the input and translating it to API calls, figuring out the execution Ids and the sub flow ids might be a bit more cumbersome as well and the diff would indeed be awful.
But other than that it will support all expected features

In comparison current implementation does work, with the exception of reordering executions and subflows. This is because it seems the API were created solely tot accommodate the keycloak admin UI, which is a pity.

I do not really have the time right now to transform current solution to a solution with TypeList, as it will take a lot of time for me to figure it out, because I can not copy paste it from anywhere else in the code base.
But I agree, working with a TypeList will probably be worth investigating.
(In any case I might fork the project to go forward as I already lost a bit too much time on this feature as it is.) Still have to work on config of authenticators as well

mrparkers:
> This is because it seems the API were created solely to accommodate the keycloak admin UI, which is a pity.

Yeah, I have come to the same conclusion while working on this provider.  There is a lot of room for improvement for its API.

I agree with all of the points you have made about my suggested solution.  It won't be easy to implement, but I think it is the only way to get all of the needed features.  I think I will have more time after the vacation I'm taking this week to take another stab at it.

tomrutsaert:
> ...
> 
> I agree with all of the points you have made about my suggested solution. It won't be easy to implement, but I think it is the only way to get all of the needed features. I think I will have more time after the vacation I'm taking this week to take another stab at it.

@mrparkers, Were you able to take a stab at this?

tomrutsaert:
Any news on a solution with nested lists?

mrparkers:
Unfortunately I haven't had the time to work on this, it was on my todo list to take another look after finishing support for the `keycloak_role` feature.

darin-sai:
We're needing this now, so @prowe and I have started to look in to supporting Flows.

tomrutsaert:
> 
> 
> We're needing this now, so @prowe and I have started to look in to supporting Flows.

That is good news! 
Feel free to use what ever you want (or not want )from this pull request

anandxmj:
What is the status on this PR. When is the plan to merge it. Please let me know

tomrutsaert:
@mrparkers It seems nobody is willing or has the time to convert this PR to work with the nested lists.
@darin-sai or @anandxmj or mrparkers were you able to make any progress on this PR with nested lists?

In the mean time we have no way to support custom flows although this PR covers 99% of all usecases.
While this PR is open for almost 6 months and while in 1 year, perhaps 2 years we will have keycloak x, where this system might be overhauled completely, would it not be better (not perfect, but better) to include this capability/feature in current state?

tomrutsaert:
> Yeah, it sounds like you're coming to the same conclusions I came to when attempting to do this originally.
> 
> I think a PR upstream would definitely help if you are up for the challenge.
> 
> My other idea for representing this in HCL is to have a data source representing a subflow and its executions, which could be used in the top level flow to accurately represent all subflows and their order - only modifying a single resource.
> 
> It would kinda look like this:
> 
> ```terraform
> resource "keycloak_authentication_flow" "top_flow" {
>     alias = "foo"
>     
>     execution { // TypeList, so the order of executions actually matters
>         provider = "foo"
>     }
> 
>     execution {
>         provider = "bar"
>     }
> 
>     subflow {
>         json = "${data.keycloak_authentication_subflow.subflow.json}"
>     }
> }
> 
> data "keycloak_authentication_subflow" "subflow" {
>     alias = "subflow"
> 
>     execution {
>         provider = "subflow_foo"
>     }
> 
>     subflow {
>         json = "${data.keycloak_authentication_subflow.second_subflow.json}"
>     }
> }
> ```
> 
> The data sources here don't actually interact with the Keycloak API in any way, they just serialize the attributes of a subflow and its executions into JSON so the top level flow resource can read them. This will allow you to specify an infinite number of subflows and their own executions, and the order of these would be controlled by the top level resource.
> 
> The downside of this approach is that the plan diff would look awful.
> 
> What are your thoughts on this?

@mrparkers I was taking a look again at you proposed theoretical solution and I was trying to implement this. But I encounter some problems:

I do not see how working with data resource for the subflows would actually create the subflows? 
Wouldn't this mean that the subflows need to already exist? Otherwise terraform can not read them and be consequence the top flow can not reference them. Am I missing something?

I did some test with TypedList and re-ording of subflows and executions,
And it seems terraform is just switching the parameters of the elements within the typedList, It is not actually reordering the elements theme self, So you would still need to :
1. destroy and recreate the whole top flow.
or
2. interpret the given input list  and fetch every subflow and execution from keycloak api, and check if anything actually has changed.... + calculate for which subflow/execution you need to call lower-priority or raise-priority. after every changes fetch every execution again , and verify the order again, till everything is in its correct place.

This was terraform plan in pseudo test switching the position of a sub-flow(browser-copy-flow-forms) with an execution(auth-spnego) in typedlist:
`  ~ keycloak_authentication_flow_improved.browser-copy-flow3
      sub.1.alias:         "browser-copy-flow-forms" => ""
      sub.1.authenticator: "" => "auth-spnego"
      sub.1.requirement:   "ALTERNATIVE" => "DISABLED"
      sub.1.type:          "subflow" => "execution"
      sub.2.alias:         "" => "browser-copy-flow-forms"
      sub.2.authenticator: "auth-spnego" => ""
      sub.2.requirement:   "DISABLED" => "ALTERNATIVE"
      sub.2.type:          "execution" => "subflow"
`

In any case the more I am looking at this, this might be a too hard solution  to replace a good enough solution (with same features) which is already provided.
Perhaps I am missing again something or am I not completely understanding how terraform works.

Keep in mind, this is all going to be obsolete when keycloak X will be released.

What are your thoughts?

tomrutsaert:
@mrparkers any new thoughts on this PR or on my questions/remarks in the last comment?

mrparkers:
Apologies for not getting to this sooner, this slipped my mind.

My initial thoughts with the data source for subflows is not for the data sources themselves to create subflows, but for them to just store arbitrary JSON that the top level resource would be able to read and create subflows from.  So yes, the data sources would need to exist first before they could be referenced.  This isn't a perfect solution (or even a great solution), but it's the only thing I can come up with that will solve every use case.

Honestly, I agree with you that Keycloak.X will probably scrap the idea of flows / executions, and seeing as we are supposed to get a release candidate for that later this year, I'm wondering if it would be best to save your time until then.  I imagine that there will be quite a lot of work involved with updating the provider to support the new API and features.

tomrutsaert:
> Apologies for not getting to this sooner, this slipped my mind.
> 
> My initial thoughts with the data source for subflows is not for the data sources themselves to create subflows, but for them to just store arbitrary JSON that the top level resource would be able to read and create subflows from. So yes, the data sources would need to exist first before they could be referenced. This isn't a perfect solution (or even a great solution), but it's the only thing I can come up with that will solve every use case.
> 
> Honestly, I agree with you that Keycloak.X will probably scrap the idea of flows / executions, and seeing as we are supposed to get a release candidate for that later this year, I'm wondering if it would be best to save your time until then. I imagine that there will be quite a lot of work involved with updating the provider to support the new API and features.

I need the features supported in this PR,  In some way I need to be able to manage custom flows.
So if we are not implementing the json way you mentioned, my only options at this moment are:
1. convince you to accept this PR as is, and solve the issues as they appear.
2. fork the project, and keep merging new features from here into my fork + also solve issues when they occur

I prefer option 1. :-)

Could you remind me, which features I was missing in my solution? For some reason I can find the missing features anymore.

(today, I looked at the code base from keycloak x, not much has changed since the announcement)

mrparkers:
Okay, I'm fine with your PR being merged so it can provide value to you.  I'm not sure I remember what I thought was missing before - could you update your PR to fix merge conflicts, and I'll give the code another review?

tomrutsaert:
Great news, I am now on holiday, I will fix the merge conflicts in the beginning of next week.

tomrutsaert:
Closed as of above comment, code changes on top of latest master can be found in #215

svenstaro:
This could use some more complete docs with a complex example I think.

tomrutsaert:
> This could use some more complete docs with a complex example I think.

Yes it could. I will try to add more complex example in the future

svenstaro:
Ok I'll try to figure something out and hopefully add a complex example.

svenstaro:
For Arch Linux, I came up with this fairly complex real-world example: https://gitlab.archlinux.org/archlinux/infrastructure/-/blob/master/tf-stage2/keycloak.tf

PR #139, opened 2019-07-31
Realm Password policies

tomrutsaert:
Added Support for password policies.
It checks if the password policy provider is installed.
I did not change or wrap the syntax of the realm password policy string the keycloak API uses.

### DIFF
--- a/example/main.tf
+++ b/example/main.tf
@@ -51,7 +51,7 @@ resource "keycloak_realm" "test" {
 		}
 	}
 
-	password_policy = "upperCase(1) and length(8) and forceExpiredPasswordChange(365) and notUsernametwo()"
+	password_policy = "upperCase(1) and length(8) and forceExpiredPasswordChange(365) and notUsername"
 }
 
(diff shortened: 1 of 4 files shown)


tomrutsaert:
Some checks were not successful
1 failing and 1 successful checks:
ci/circleci: test-password — Your tests failed on CircleCI 

--> These test seem to be flaky, they fail often but not always, the failure has nothing to with my change.

mrparkers:
They are flaky, I can merge this regardless.  Thanks for the contribution!

timbmg:
could your provide an example for this?

tomrutsaert:
`resource "keycloak_realm" "test" {
  realm                  = "test"
  enabled              = true
  password_policy = "upperCase(1) and length(8) and forceExpiredPasswordChange(365) and notUsername"
}`

The provider uses the same password policy string that keycloak uses itself.


You should always find at least on example for each feature in the example files:
https://github.com/mrparkers/terraform-provider-keycloak/tree/master/example

I should add this to the documentation (I do tend to forget this)

PR #142, opened 2019-08-07
Set alias of oidc provider on import

nagytech:
https://github.com/mrparkers/terraform-provider-keycloak/issues/141#issuecomment-519306566


mrparkers:
Thanks for the fix!  Sorry for the flaky test there

PR #146, opened 2019-08-30
Add keycloak_default_groups resource for creating realm default groups

### LINKED ISSUE #144: Add support for default groups
AFAICT, there is no way to specify default groups at the moment. This would be fairly easy to implement, since it's a really basic resource. The relevant endpoints are:

    GET /{realm}/default-groups (returns a list of all default groups)
    PUT /{realm}/default-groups/{groupId} (add a group to the list of default groups)
    DELETE /{realm}/default-groups/{groupId} (delete a group from the list of default groups)

The first way that comes to mind for how this could be implemented is having a very basic resource for each default group that simply maps a realm id to a group id. This maps very well to the keycloak API. The HCL would look something look this:

```
resource "keycloak_group" "example_group" {
  realm_id = "${keycloak_realm.realm.id}"
  name     = "example-group"
}

resource "keycloak_default_group" "example_group" {
  realm_id = "${keycloak_realm.realm.id}"
  group_id = "${keycloak_group.example_group.id}"
}
```

We could also have a default groups property on realms, or have a default groups resource, or have a default group property on groups, but I don't think those solutions map very well to the API. Additionally, we could specify it on the realm, but then we get into a scenario where if a realm references groups that need to be created after the realm is then there's a potential deadlock.

I would be happy to take a crack at adding this feature.

### LINKED ISSUE #145: Errors aren't checked properly when removing a group membership user from a group
https://github.com/mrparkers/terraform-provider-keycloak/blob/0f1fe1bad1e107f0199bb93466b18003450e7a79/provider/resource_keycloak_group_memberships.go#L106

I think this code should be returning `err` rather than `nil`. As-is, if there's an error removing a user from a group nothing will happen.

fharding1:
## Summary

This PR adds a new `keycloak_default_groups` resource which allows you to manage default groups for a realm.

Closes #144 
Closes #145 

## TODO

- [x] Validate that a group membership resource doesn't exist or figure out some other solution for that (or just document that using a default group and a group membership resource will lead to funky results), see https://github.com/mrparkers/terraform-provider-keycloak/issues/144#issuecomment-526634858
- [x] Add tests
- [x] Add ability to import default groups
- [x] Add documentation
- [x] Add an example

### DIFF
--- a/provider/resource_keycloak_default_groups_test.go
+++ b/provider/resource_keycloak_default_groups_test.go
@@ -33,6 +33,28 @@ func TestAccKeycloakDefaultGroups_basic(t *testing.T) {
 	})
 }
 
+func TestAccKeycloakDefaultGroups_import(t *testing.T) {
+	realmName := "terraform-" + acctest.RandString(10)
+	groupName := "terraform-group-" + acctest.RandString(10)
+
+	resource.Test(t, resource.TestCase{
+		Providers: testAccProviders,
+		PreCheck:  func() { testAccPreCheck(t) },
+		Steps: []resource.TestStep{
+			{
+				Config: testKeycloakDefaultGroups_basic(realmName, groupName),
+				Check:  testAccCheckGroupsAreDefault("keycloak_default_groups.group_default", []string{groupName}),
+			},
+			{
+				ResourceName:      "keycloak_default_groups.group_default",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateId:     realmName,
+			},
+		},
+	})
+}
+
 func TestAccKeycloakDefaultGroups_updateInPlace(t *testing.T) {
 	realmName := "terraform-" + acctest.RandString(10)
 
--- a/provider/resource_keycloak_default_groups_test.go
+++ b/provider/resource_keycloak_default_groups_test.go
@@ -2,7 +2,6 @@ package provider
 
 import (
 	"fmt"
-	"strings"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/acctest"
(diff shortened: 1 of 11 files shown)


fharding1:
I changed the example to be formatted with the canonical `terraform fmt`. If this isn't desired, I can revert it.

fharding1:
Hmm, tests are failing on `TestAccKeycloakDataSourceOpenidClient_basic` and `TestAccKeycloakDataSourceOpenidClientAuthorizationPolicy_basic`, but I didn't touch those.

provider/resource_keycloak_default_groups_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	var defaultGroupResources []string
+	for _, defaultGroup := range defaultGroups {
+		defaultGroupResources = append(defaultGroupResources, fmt.Sprintf(`"${keycloak_group.%s.id}"`, defaultGroup))
+	}
+
+	out += fmt.Sprintf(`
+resource "keycloak_default_groups" "group_default" {
+	realm_id = "${keycloak_realm.realm.id}"
+	group_ids = [%s]
```
mrparkers: I have a helper function in the test package called `arrayOfStringsForTerraformResource` that does this for you, could you refactor this to use that function?

fharding1: Perfect! Can do.

fharding1:
Had to force push to get rid of an accidentally commited update to the `go.mod` file (thanks go1.13rc2 :smile: )

fharding1:
Thanks for the review and for merging! :tada:

PR #153, opened 2019-09-07
Update keycloak_group_roles.md

xiaoyang-connyun:
When I use `roles` as the field, `terraform` tells me that is wrong, and it should be `role_ids` instead.


mrparkers:
Thanks for catching this!

PR #155, opened 2019-09-11
Add option to configure the Keycloak client timeout

LoicAG:
Our Keycloak instance is sometimes a bit slow to respond:

```
module.release_environment.keycloak_realm.realm: Creating...

Error: Post https://keycloak.is4.local/auth/admin/realms: net/http: request canceled (Client.Timeout exceeded while awaiting headers)

  on .terraform/modules/release_environment/main.tf line 5, in resource "keycloak_realm" "realm":
   5: resource "keycloak_realm" "realm" {
```

The request is cancelled, but the realm does get created, which leads to a conflict on subsequent `terraform apply` calls; and since the realm is not in the Terraform state, a `terraform destroy` does not solve the issue.

This PR allows for the Keycloak client timeout to be configurable, while keeping the default value of 5 seconds.


LoicAG:
The tests are failing, but the failures don't seem related to my changes

LoicAG:
ping @mrparkers , what do you think about this?

mrparkers:
Sorry for the delayed review.  This looks great, I was actually starting to run into issues with this myself.

Thanks for the contribution!

PR #166, opened 2019-10-15
Fix input for keycloak_realm data source

aroemers:
When trying out the new `keycloak_realm` data source, I received the the following error:

> Error: "realm": this field cannot be set

I have the suspicion this is a bug, this PR tries to fix that.


mrparkers:
Yeah, you're right, good catch.  I am not sure how I missed this in my review.

I'll merge this and work on fixing the test case for this data source since it is also not correct and cut a release with the changes.

Thanks for the contribution!

aroemers:
Thank you Michael, for your quick response and release! 👍

PR #170, opened 2019-10-23
feat: Support PKCE on OIDC Clients

### LINKED ISSUE #169: Support PKCE switch in Advanced Settings of openID Client
Hi there,

perhaps I haven't found it yet, but I am missing the PKCE switch in the "Advanced Settings" - Tab of an openid client in keycloak (using v7). 
Bildschirmfoto 2019-10-21 um 11 35 20

As it is state of the art now also for SPAs to use authorization grant with PKCE enabled (source e.g. https://developer.okta.com/blog/2019/05/01/is-the-oauth-implicit-flow-dead) I'd hoped there was a possibility to add it via terraform provider, but I can't find it yet. Am I missing something? If not, it would be great to have it!

PSanetra:
This PR adds support for the advanced OIDC client setting `Proof Key for Code Exchange Code Challenge Method` on Keycloak >= 7.0.0

The setting is also applied on Keycloak versions < 7.0.0, but is ignored on those versions. (At least on Keycloak 6.0.1)

Fixes #169

### DIFF
--- a/provider/resource_keycloak_openid_client.go
+++ b/provider/resource_keycloak_openid_client.go
@@ -12,6 +12,7 @@ import (
 var (
 	keycloakOpenidClientAccessTypes                        = []string{"CONFIDENTIAL", "PUBLIC", "BEARER-ONLY"}
 	keycloakOpenidClientAuthorizationPolicyEnforcementMode = []string{"ENFORCING", "PERMISSIVE", "DISABLED"}
+	keycloakOpenidClientPkceCodeChallengeMethod            = []string{"plain", "S256"}
 )
 
 func resourceKeycloakOpenidClient() *schema.Resource {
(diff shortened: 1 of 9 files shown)


mrparkers:
Thanks for the PR!

Could you run `make fmt` and push up the changes?

provider/resource_keycloak_openid_client.go
```diff
@@ -12,6 +12,7 @@ import (
 var (
 	keycloakOpenidClientAccessTypes                        = []string{"CONFIDENTIAL", "PUBLIC", "BEARER-ONLY"}
 	keycloakOpenidClientAuthorizationPolicyEnforcementMode = []string{"ENFORCING", "PERMISSIVE", "DISABLED"}
+	keycloakOpenidClientPkceCodeChallengeMethod            = []string{"", "plain", "S256"}
```
mrparkers: Is the empty string needed here if the `pkce_code_challenge_method` attribute is optional?  I would just expect users to omit this entirely rather than provide an empty string.

PSanetra: I'm not sure about this. Would it still be possible to omit this value if the actual value is provided by a variable via interpolation syntax?

PSanetra: I have removed the empty string from the options.

PSanetra:
@mrparkers I have pushed your suggested changes.

PR #171, opened 2019-10-28
Fix identity provider update and service account role management

AndrewChubatiuk:
Hi @mrparkers 
Prepared a PR with fixes for 
 - OIDC identity provider: It doesn't keep a valid client_secret value during resource update
 - service account role: sometimes it doesn't save a full state and tries to recreate resource once again

### DIFF
--- a/provider/resource_keycloak_oidc_identity_provider.go
+++ b/provider/resource_keycloak_oidc_identity_provider.go
@@ -94,6 +94,8 @@ func resourceKeycloakOidcIdentityProvider() *schema.Resource {
 func getOidcIdentityProviderFromData(data *schema.ResourceData) (*keycloak.IdentityProvider, error) {
 	rec, _ := getIdentityProviderFromData(data)
 	rec.ProviderId = data.Get("provider_id").(string)
+	_, useJwksUrl := data.GetOk("jwks_url")
+	_, enableUserInfo := data.GetOk("user_info_url")
 
 	extraConfig := map[string]interface{}{}
 	if v, ok := data.GetOk("extra_config"); ok {
@@ -107,6 +109,7 @@ func getOidcIdentityProviderFromData(data *schema.ResourceData) (*keycloak.Ident
 		ValidateSignature:    keycloak.KeycloakBoolQuoted(data.Get("validate_signature").(bool)),
 		AuthorizationUrl:     data.Get("authorization_url").(string),
 		ClientId:             data.Get("client_id").(string),
+		ClientSecret:         data.Get("client_secret").(string),
 		HideOnLoginPage:      keycloak.KeycloakBoolQuoted(data.Get("hide_on_login_page").(bool)),
 		TokenUrl:             data.Get("token_url").(string),
 		LogoutUrl:            data.Get("logout_url").(string),
@@ -115,16 +118,8 @@ func getOidcIdentityProviderFromData(data *schema.ResourceData) (*keycloak.Ident
 		JwksUrl:              data.Get("jwks_url").(string),
 		UserInfoUrl:          data.Get("user_info_url").(string),
 		ExtraConfig:          extraConfig,
-	}
-	_, useJwksUrl := data.GetOk("jwks_url")
-	rec.Config.UseJwksUrl = keycloak.KeycloakBoolQuoted(useJwksUrl)
-	_, enableUserInfo := data.GetOk("user_info_url")
-	rec.Config.DisableUserInfo = keycloak.KeycloakBoolQuoted(!enableUserInfo)
-
-	if data.HasChange("client_secret") {
-		rec.Config.ClientSecret = data.Get("client_secret").(string)
-	} else {
-		rec.Config.ClientSecret = "**********"
+		UseJwksUrl:						keycloak.KeycloakBoolQuoted(useJwksUrl),
+		DisableUserInfo:      keycloak.KeycloakBoolQuoted(!enableUserInfo),
 	}
 
 	return rec, nil
--- a/provider/resource_keycloak_oidc_identity_provider.go
+++ b/provider/resource_keycloak_oidc_identity_provider.go
@@ -118,7 +118,7 @@ func getOidcIdentityProviderFromData(data *schema.ResourceData) (*keycloak.Ident
 		JwksUrl:              data.Get("jwks_url").(string),
 		UserInfoUrl:          data.Get("user_info_url").(string),
 		ExtraConfig:          extraConfig,
-		UseJwksUrl:						keycloak.KeycloakBoolQuoted(useJwksUrl),
+		UseJwksUrl:           keycloak.KeycloakBoolQuoted(useJwksUrl),
 		DisableUserInfo:      keycloak.KeycloakBoolQuoted(!enableUserInfo),
 	}
 
--- a/provider/resource_keycloak_oidc_identity_provider.go
+++ b/provider/resource_keycloak_oidc_identity_provider.go
@@ -123,6 +123,8 @@ func getOidcIdentityProviderFromData(data *schema.ResourceData) (*keycloak.Ident
 
 	if data.HasChange("client_secret") {
 		rec.Config.ClientSecret = data.Get("client_secret").(string)
+	} else {
+		rec.Confog.ClientSecret = "**********"
 	}
 
 	return rec, nil
--- a/provider/resource_keycloak_oidc_identity_provider.go
+++ b/provider/resource_keycloak_oidc_identity_provider.go
@@ -124,7 +124,7 @@ func getOidcIdentityProviderFromData(data *schema.ResourceData) (*keycloak.Ident
 	if data.HasChange("client_secret") {
 		rec.Config.ClientSecret = data.Get("client_secret").(string)
 	} else {
-		rec.Confog.ClientSecret = "**********"
+		rec.Config.ClientSecret = "**********"
 	}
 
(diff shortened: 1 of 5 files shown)


provider/resource_keycloak_oidc_identity_provider.go
```diff
@@ -123,6 +123,8 @@ func getOidcIdentityProviderFromData(data *schema.ResourceData) (*keycloak.Ident
 
 	if data.HasChange("client_secret") {
 		rec.Config.ClientSecret = data.Get("client_secret").(string)
+	} else {
+		rec.Config.ClientSecret = "**********"
```
mrparkers: Is this going to send the string `**********` to Keycloak on update?  I'm wondering if it would be better to remove the `if` statement on line 124 and always call `rec.Config.ClientSecret = data.Get("client_secret").(string)` so Keycloak is always getting the correct client secret on every call.

provider/resource_keycloak_openid_client_service_account_role.go
```diff
@@ -40,18 +40,32 @@ func resourceKeycloakOpenidClientServiceAccountRole() *schema.Resource {
...
+			ServiceAccountUserId: serviceAccountRoleId,
+		}, nil
 	}
+
+	return nil, nil
 }
 
 func setOpenidClientServiceAccountRoleData(data *schema.ResourceData, serviceAccountRole *keycloak.OpenidClientServiceAccountRole) {
-	data.SetId(serviceAccountRole.Id)
+	data.SetId(fmt.Sprintf("%s/%s", serviceAccountRole.ServiceAccountUserId, serviceAccountRole.Id))
```
mrparkers: This might be a breaking change since the ID for this resource is being changed.  Can you verify that these resources aren't being recreated because of this change?

provider/resource_keycloak_openid_client_service_account_role.go
```diff
@@ -40,18 +40,32 @@ func resourceKeycloakOpenidClientServiceAccountRole() *schema.Resource {
...
-		RealmId:              data.Get("realm_id").(string),
-		ServiceAccountUserId: data.Get("service_account_user_id").(string),
+func getOpenidClientServiceAccountRoleFromData(data *schema.ResourceData, keycloakClient *keycloak.KeycloakClient) (*keycloak.OpenidClientServiceAccountRole, error) {
+
+	containerId := data.Get("client_id").(string)
+	roleName := data.Get("role").(string)
+	realmId := data.Get("realm_id").(string)
+	serviceAccountRoleId := data.Get("service_account_user_id").(string)
+
+	if realmId != "" && containerId != "" {
```
mrparkers: `realm_id` and `client_id` are both required attributes, so I think this statement is always `true`.

provider/resource_keycloak_openid_client_service_account_role.go
```diff
@@ -60,8 +74,16 @@ func setOpenidClientServiceAccountRoleData(data *schema.ResourceData, serviceAcc
...
 func resourceKeycloakOpenidClientServiceAccountRoleCreate(data *schema.ResourceData, meta interface{}) error {
 	keycloakClient := meta.(*keycloak.KeycloakClient)
-	serviceAccountRole := getOpenidClientServiceAccountRoleFromData(data)
-	err := keycloakClient.NewOpenidClientServiceAccountRole(serviceAccountRole)
+	serviceAccountRole, err := getOpenidClientServiceAccountRoleFromData(data, keycloakClient)
+	if err != nil {
+		return err
+	}
+
+	if serviceAccountRole == nil {
```
mrparkers: Can this error ever happen?  I don't think it's possible for `getOpenidClientServiceAccountRoleFromData` to return `nil` without also returning an error.

provider/resource_keycloak_openid_client_service_account_role.go
```diff
@@ -72,12 +94,16 @@ func resourceKeycloakOpenidClientServiceAccountRoleCreate(data *schema.ResourceD
...
-	realmId := data.Get("realm_id").(string)
-	serviceAccountUserId := data.Get("service_account_user_id").(string)
-	clientId := data.Get("client_id").(string)
-	id := data.Id()
+	serviceAccountRole, err := getOpenidClientServiceAccountRoleFromData(data, keycloakClient)
+	if err != nil {
+		return err
+	}
+
+	if serviceAccountRole == nil {
```
mrparkers: Same comment as above

provider/resource_keycloak_openid_client_service_account_role.go
```diff
@@ -90,23 +116,27 @@ func resourceKeycloakOpenidClientServiceAccountRoleRead(data *schema.ResourceDat
...
-	serviceAccountUserId := data.Get("service_account_user_id").(string)
-	clientId := data.Get("client_id").(string)
-	id := data.Id()
+	serviceAccountRole, err := getOpenidClientServiceAccountRoleFromData(data, keycloakClient)
+	if err != nil {
+		return err
+	}
 
-	return keycloakClient.DeleteOpenidClientServiceAccountRole(realmId, serviceAccountUserId, clientId, id)
+	if serviceAccountRole != nil {
```
mrparkers: Same comment as above, I don't think this can ever be false

provider/resource_keycloak_oidc_identity_provider.go
AndrewChubatiuk: This is how actually browser is working

provider/resource_keycloak_openid_client_service_account_role.go
AndrewChubatiuk: Just new resources will be created. In context of service account roles it'll try to apply existing changes once to add service account roles with ned IDs but POST action to service account roles endpoint always returns 204 even if role was already added

provider/resource_keycloak_oidc_identity_provider.go
AndrewChubatiuk: Anyway I've updated it

provider/resource_keycloak_openid_client_service_account_role.go
AndrewChubatiuk: Had another problem. And tested locally this change, forgot to remove

provider/resource_keycloak_openid_client_service_account_role.go
AndrewChubatiuk: They can be empty if you're using ```realm_id = keycloak_realm.realm.id ``` and ```client_id  = keycloak_openid_client.client.id``` that do not exist yet

AndrewChubatiuk:
@mrparkers Could you please take a look once again?

provider/resource_keycloak_openid_client_service_account_role.go
```diff
@@ -40,18 +40,32 @@ func resourceKeycloakOpenidClientServiceAccountRole() *schema.Resource {
...
-		ServiceAccountUserId: data.Get("service_account_user_id").(string),
+func getOpenidClientServiceAccountRoleFromData(data *schema.ResourceData, keycloakClient *keycloak.KeycloakClient) (*keycloak.OpenidClientServiceAccountRole, error) {
+	containerId := data.Get("client_id").(string)
+	roleName := data.Get("role").(string)
+	realmId := data.Get("realm_id").(string)
+	serviceAccountRoleId := data.Get("service_account_user_id").(string)
+
+	role, err := keycloakClient.GetRoleByName(realmId, containerId, roleName)
+	if err != nil {
+		if handleNotFoundError(err, data) == nil {
```
mrparkers: This is going to call `data.SetId("")` if the error is s 404, which removes this resource from state.  Is this what you want?

AndrewChubatiuk: yes, cause if role doesn't exist yet it cannot be attached to a client

AndrewChubatiuk: There's a get request before delete action
https://github.com/mrparkers/terraform-provider-keycloak/blob/master/keycloak/openid_client_service_account_role.go#L33

mrparkers: What I mean is that the `keycloak_openid_client_service_account_role` resource will be removed from state if the `GetRoleByName` call fails with a 404.  That seems problematic since ultimately this function won't return an error and the creation will appear to be successful.

If the `keycloak_openid_client_service_account_role`resource isn't valid without an existing role, shouldn't create just fail?

AndrewChubatiuk: When terraform plans resources creation it uses empty values for attributes of resource which is not yet created. Don't know how it should work, but at least have noticed such a behaviour for this provider. It'll cause errors during every plan when resources are not created yet. Without this check tests are always failing
https://circleci.com/gh/mrparkers/terraform-provider-keycloak/1152?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

mrparkers: Okay, your logic makes sense, thanks for the explanation.  I think, however, that all you really want to do is check if the error is a 404 and not necessarily remove the resource from state.
```suggestion
		if keycloak.ErrorIs404(err) {
```

AndrewChubatiuk: yes :)

AndrewChubatiuk: @mrparkers just updated

PR #175, opened 2019-11-16
Added condition for fetching roles having forward slash(/) in its name

Shubhammathur22:
Issue: Terraform giving 404 error while fetching client roles having forward slash(/) in its name.

Resolution: Added Replace string where ever required

Example Resource Name: Test/123

### DIFF
--- a/keycloak/role.go
+++ b/keycloak/role.go
@@ -3,6 +3,7 @@ package keycloak
 import (
 	"fmt"
 	"log"
+	"strings"
 )
 
 type Role struct {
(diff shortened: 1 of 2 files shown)


Shubhammathur22:
Hi @mrparkers , Could you please check this PR ?

mrparkers:
Hi @Shubhammathur22, thanks for the PR!

The code itself looks good.  Would you mind adding a test case for the change you made?

You can copy/paste [this test case](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/resource_keycloak_role_test.go#L12-L33) and update the `roleName` variable to include a forward slash, and you should be good to go!

Shubhammathur22:
Sure @mrparkers , I will add the test case for this PR shortly

Shubhammathur22:
@mrparkers , added the test case, please check

Shubhammathur22:
Thanks @mrparkers for managing this very useful provider code. Also, could you please share the version number which would be having the above changes? and when it will be released?

mrparkers:
My pleasure, I'm glad you find it useful!

CI is in the middle of building and releasing 1.13.1 right now. It should be available in about 10 minutes.

Shubhammathur22:
Great !!! Thanks again :)

PR #179, opened 2019-11-24
Fetch More than default 100 members for a group

Shubhammathur22:
Issue: Keycloak's ** GET /{realm}/groups/{id}/members** API by default returns maximum 100 members for any group.

Resolution: max parameter passed with the API call.

Ref Issue: https://github.com/mrparkers/terraform-provider-keycloak/issues/178#issue-527021618

### DIFF
--- a/provider/resource_keycloak_group_memberships_test.go
+++ b/provider/resource_keycloak_group_memberships_test.go
@@ -34,6 +34,23 @@ func TestAccKeycloakGroupMemberships_basic(t *testing.T) {
 	})
 }
 
+func TestAccKeycloakGroupMemberships_moreThan100members(t *testing.T) {
+	realmName := "terraform-" + acctest.RandString(10)
+	groupName := "terraform-group-" + acctest.RandString(10)
+	username := "terraform-user-101"
+
+	resource.Test(t, resource.TestCase{
+		Providers: testAccProviders,
+		PreCheck:  func() { testAccPreCheck(t) },
+		Steps: []resource.TestStep{
+			{
+				Config: testKeycloakGroupMemberships_moreThan100members(realmName, groupName),
+				Check:  testAccCheckUserBelongsToGroup("keycloak_group_memberships.group_members", username),
+			},
+		},
+	})
+}
+
 func TestAccKeycloakGroupMemberships_updateGroupForceNew(t *testing.T) {
 	realmName := "terraform-" + acctest.RandString(10)
 
@@ -409,6 +426,38 @@ resource "keycloak_group_memberships" "group_members" {
 	]
 }
 	`, realm, group, username)
+}
+
+func testKeycloakGroupMemberships_moreThan100members(realm, group string) string {
+	count := 200
+	var usersInGroupInterpolated []string
+	for i := 0; i < count; i++ {
+		usersInGroupInterpolated = append(usersInGroupInterpolated, fmt.Sprintf("${keycloak_user.users.%d.username}", i))
+	}
+	return fmt.Sprintf(`
+resource "keycloak_realm" "realm" {
+        realm = "%s"
+}
+
+resource "keycloak_group" "group" {
+        name     = "%s"
+        realm_id = "${keycloak_realm.realm.id}"
+}
+
+resource "keycloak_user" "users" {
+	count = %d
+        realm_id = "${keycloak_realm.realm.id}"
+        username = "terraform-user-${count.index}"
+}
+
+resource "keycloak_group_memberships" "group_members" {
+        realm_id = "${keycloak_realm.realm.id}"
+        group_id = "${keycloak_group.group.id}"
+
+        members = %s
+}
+
+        `, realm, group, count, arrayOfStringsForTerraformResource(usersInGroupInterpolated))
 }
 
 func testKeycloakGroupMemberships_updateGroupForceNew(realm, groupOne, groupTwo, username, currentGroup string) string {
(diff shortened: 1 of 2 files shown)


Shubhammathur22:
Hi @mrparkers , please check this PR. I am not sure what is the best way to achieve this but it looked fine to me.

mrparkers:
Hi @Shubhammathur22, apologies for not getting to this sooner.

I think this approach works, but it might be a good idea to use the pagination features that are provided for this API instead so Keycloak isn't loading potentially tens of thousands of users into memory at once.

Looks like we can use the `first` query param to specify the index we should start at ([API docs](https://www.keycloak.org/docs-api/8.0/rest-api/index.html#_getmembers)).  The first request can omit this query param entirely.  If the amount of users returned from the first request is 100, then we can continue to send follow up requests with `first=100`, `first=200`, etc.  Once a request returns less than 100 users, we can stop and return the final list of >100 users.

For a simple test, the following HCL should result in a failing test in `master`:

```hcl
  realm   = "test"
  enabled = true
}

resource "keycloak_user" "users" {
  count    = 200

  realm_id = keycloak_realm.test.id
  username = "user-${count.index}"
}

resource "keycloak_group" "test_group" {
  realm_id = keycloak_realm.test.id
  name     = "test_group"
}

resource "keycloak_group_memberships" "test_group_members" {
  realm_id = keycloak_realm.test.id
  group_id = keycloak_group.test_group.id

  members  = keycloak_user.users.*.username
}
```

This test can be added without any assertions and it will still fail, since a following plan after applying this HCL will not be empty, which will automatically fail the test.

Let me know if you have any questions or would like some help with this.  Thanks!

Shubhammathur22:
Thanks @mrparkers for sharing another approach to this but i doubt if pagination will help us in memory allocation as we will be fetching all the users belonging to that group.

Yes, pagination will help in loading many users at once but as a bigger picture, at the end the memory usage would be same but in this case API calls will also increase.

Please suggest which approach to use, will work on modifying the code accordingly.

mrparkers:
Yeah I suppose it doesn't matter all that much.

If we're fetching all members at once, I think you can pass `-1` for the `max` query parameter to get all of a group's members, so you don't need the extra API call.

Can you try that and add the test case I mentioned as well?  Thanks!

Shubhammathur22:
Sure @mrparkers ,i will check with max=-1 parameter and add the test case. Thanks

Shubhammathur22:
Thanks @mrparkers , this max=-1 thing looks to be working fine. Could you also share the location where should i add the test case?

provider/resource_keycloak_group_memberships_test.go
```diff
@@ -411,6 +428,38 @@ resource "keycloak_group_memberships" "group_members" {
 	`, realm, group, username)
 }
 
+func testKeycloakGroupMemberships_moreThan100members(realm, group string) string {
+	count := 200
```
mrparkers: On second thought, let's update this number to something like `110` - just enough for the test to fail, so we don't need to create an additional 90 users.
```suggestion
	count := 110
```

provider/resource_keycloak_group_memberships_test.go
```diff
@@ -411,6 +428,38 @@ resource "keycloak_group_memberships" "group_members" {
...
+	count = %d
+        realm_id = "${keycloak_realm.realm.id}"
+        username = "terraform-user-${count.index}"
+}
+
+resource "keycloak_group_memberships" "group_members" {
+        realm_id = "${keycloak_realm.realm.id}"
+        group_id = "${keycloak_group.group.id}"
+
+        members = %s
```
mrparkers: I think you should be able to use a [splat expression](https://www.terraform.io/docs/configuration/expressions.html#splat-expressions) here instead of building a list of usernames manually.

```suggestion
        members = keycloak_user.users.*.username
```

provider/resource_keycloak_group_memberships_test.go
```diff
@@ -34,6 +34,23 @@ func TestAccKeycloakGroupMemberships_basic(t *testing.T) {
...
+	groupName := "terraform-group-" + acctest.RandString(10)
+	username := "terraform-user-101"
+
+	resource.Test(t, resource.TestCase{
+		Providers: testAccProviders,
+		PreCheck:  func() { testAccPreCheck(t) },
+		Steps: []resource.TestStep{
+			{
+				Config: testKeycloakGroupMemberships_moreThan100members(realmName, groupName),
+				Check:  testAccCheckUserBelongsToGroup("keycloak_group_memberships.group_members", username),
```
mrparkers: The `Check` doesn't hurt but it actually isn't needed for this test.  Terraform tests will automatically perform a `plan` after applying the config, and if the plan shows a change, it will automatically fail the test.

Before your changes, this test (without the `Check`) would fail because the old code would only fetch 100 members from Keycloak, and the Terraform state thinks there are >100 members after applying, so Terraform will think it needs to add the missing members.

The changes you introduced will correctly fetch all of the group's members, so the subsequent plan after applying will show no changes.

Anyways, I'd feel better about removing this since it seems strange to only check a single member.

provider/resource_keycloak_group_memberships_test.go
Shubhammathur22: done

provider/resource_keycloak_group_memberships_test.go
Shubhammathur22: Removed check clause

Issue #174, opened 2019-11-13
A user should be able to set RefreshTokenMaxReuse to 0 to enable RevokeRefreshToken


tomrutsaert:
A user should be able to set RefreshTokenMaxReuse to 0 to enable RevokeRefreshToken
This value is about the amount of times a refresh can be reused, in which 0 is most often chosen value.
But in the code (see below) disables RevokeRefreshToken when you set RefreshTokenMaxReuse to 0.

```
	if refreshTokenMaxReuse := data.Get("refresh_token_max_reuse").(int); refreshTokenMaxReuse > 0 {
		realm.RevokeRefreshToken = true
		realm.RefreshTokenMaxReuse = refreshTokenMaxReuse
	} else {
		realm.RevokeRefreshToken = false
	}
```

PR #183, opened 2019-12-06
brute-force-detection-settings + ssl_required + custom-realm-attributes + fix 174 + k8.0.0

tomrutsaert:
support for brute-force-detection-settings
support for ssl_required
support for custom-realm-attributes
bugfix for #174
upgrade to keycloak 8.0.0

### DIFF
--- a/keycloak/realm.go
+++ b/keycloak/realm.go
@@ -27,14 +27,15 @@ type Realm struct {
 	DisplayName string `json:"displayName"`
 
 	// Login Config
-	RegistrationAllowed         bool `json:"registrationAllowed"`
-	RegistrationEmailAsUsername bool `json:"registrationEmailAsUsername"`
-	EditUsernameAllowed         bool `json:"editUsernameAllowed"`
-	ResetPasswordAllowed        bool `json:"resetPasswordAllowed"`
-	RememberMe                  bool `json:"rememberMe"`
-	VerifyEmail                 bool `json:"verifyEmail"`
-	LoginWithEmailAllowed       bool `json:"loginWithEmailAllowed"`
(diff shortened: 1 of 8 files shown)


tomrutsaert:
The change of behavior of revoke refresh tokens settings should be mentioned.

tomrutsaert:
@mrparkers Did you have a chance to look at this PR? Do you have any remarks?

keycloak/realm.go
```diff
@@ -65,8 +66,19 @@ type Realm struct {
 	SupportLocales              []string `json:"supportedLocales"`
 	DefaultLocale               string   `json:"defaultLocale"`
 
-	//extra attributes of a realm, contains security defenses browser headers and brute force detection parameters(those still nee to be added)
-	Attributes Attributes `json:"attributes,omitempty"`
+	//extra attributes of a realm
+	Attributes map[string]interface{} `json:"attributes"`
+
+	BrowserSecurityHeaders BrowserSecurityHeaders `json:"browserSecurityHeaders"`
```
mrparkers: Is this a breaking change in Keycloak's API?  It looks like we were using `_browser_header` before - is that going to break this provider's compatibility with older versions of Keycloak?

tomrutsaert: The Realm tests stills work if I run on ac test with 'make local' and docker-compose pointing to keycloak version 4.8.3.Final
There 2 ways of setting these parameters, The UI is using the way above
I have no idea why there are 2 ways to do this. If you set both ways at the same time with different values, then setting it directly as a real parameters wins from setting it through the attributes.
Thus this seems to be the safer bet and a more correct way to do it through this method.

PR #191, opened 2019-12-30
Add option ExcludeSessionStateFromAuthResponse on OpenIdClient

daviddesre:
Allows to set exclude_session_state_from_auth_response flag on Keycloak OpenID clients

### DIFF
--- a/keycloak/openid_client.go
+++ b/keycloak/openid_client.go
@@ -48,7 +48,8 @@ type OpenidClient struct {
 }
 
 type OpenidClientAttributes struct {
-	PkceCodeChallengeMethod string `json:"pkce.code.challenge.method"`
+	PkceCodeChallengeMethod             string `json:"pkce.code.challenge.method"`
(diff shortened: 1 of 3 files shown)


keycloak/openid_client.go
```diff
@@ -48,7 +48,8 @@ type OpenidClient struct {
 }
 
 type OpenidClientAttributes struct {
-	PkceCodeChallengeMethod string `json:"pkce.code.challenge.method"`
+	PkceCodeChallengeMethod             string `json:"pkce.code.challenge.method"`
+	ExcludeSessionStateFromAuthResponse bool   `json:"exclude.session.state.from.auth.response,string"`
```
mrparkers: We typically use `KeycloakBoolQuoted` for these annoying boolean attributes that are actually strings.  https://github.com/mrparkers/terraform-provider-keycloak/blob/master/keycloak/util.go#L10

daviddesre: Thanks, i fix it :)

PR #195, opened 2020-01-07
Implement keycloak_ldap_hardcoded_role_mapper resource

### LINKED ISSUE #109: Support hardcoded-ldap-role-mapper LDAP Mapper
This would grant a specified Keycloak role to each Keycloak user linked with LDAP.  
See [documentation](https://www.keycloak.org/docs/latest/server_admin/index.html#_ldap_mappers)

It could look like this:
```
resource "keycloak_ldap_hardcoded_role_mapper" "role_mapper" {
  realm_id                = "${keycloak_realm.my-realm.id}"
  ldap_user_federation_id = "${keycloak_ldap_user_federation.my-ldap.id}"
  name                    = "my-role-mapper"
  role                    = "my-role"
}
```

madddi:
This adds a resource for the hardcoded-ldap-role-mapper. Fixes #109

### DIFF
--- a/docs/resources/keycloak_ldap_hardcoded_role_mapper.md
+++ b/docs/resources/keycloak_ldap_hardcoded_role_mapper.md
@@ -1,0 +1,55 @@
+# keycloak_ldap_hardcoded_role_mapper
+
+This mapper will grant a specified Keycloak role to each Keycloak user linked with LDAP.
+
+### Example Usage
+
+```hcl
+resource "keycloak_realm" "realm" {
+    realm   = "test"
+    enabled = true
+}
+
+resource "keycloak_ldap_user_federation" "ldap_user_federation" {
+    name                    = "openldap"
+    realm_id                = keycloak_realm.realm.id
+
+    username_ldap_attribute = "cn"
+    rdn_ldap_attribute      = "cn"
+    uuid_ldap_attribute     = "entryDN"
+    user_object_classes     = [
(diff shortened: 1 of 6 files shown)


madddi:
Hey @mrparkers, I tested the new resource with a real KeyCloak test setup and it works fine - the tests are failing though. I can't really figure out why, can you give me a hint?

mrparkers:
Sure - you can use the `TF_LOG=DEBUG` environment variable to get extra logging to see why the request failed:

```
TF_LOG=DEBUG make testacc TESTARGS="-run TestAccKeycloakLdapHardcodedRoleMapper_basic"
```

From that, I saw the response payload for the bad request:

```json
{"errorMessage":"There is no role corresponding to configured value"}
```

It looks like your test assumes the role `admin` exists in the test realm.  Perhaps the version of Keycloak you're testing against creates that role by default on new realms, but version 8.0.0 (the version CI is testing against) does not?

madddi:
Oh... :man_facepalming: Sure, this role only exists in the `master` realm by default. I'll just update the tests to create a role and use this. Thank you!

By the way: I noticed that KeyCloak, Terraform, Go and openldap are out of date in your test suite. I'd update them in a separate PR if you think that makes sense.

mrparkers:
Also, if you'd like to submit a PR to update versions and switch the usage of HCL to HCL2 that would be great!

Issue #111, opened 2019-05-09
support multiple keycloak versions + acceptance tests


mrparkers:
#106 upgrades the provider to support Keycloak 6.0.1.  While I don't believe the provider itself needs to be changed to support each version of Keycloak, it would be nice for the acceptance tests to test against all supported versions to give confidence to the users who are not ready to upgrade yet.

PR #196, opened 2020-01-08
Update CI tools and images

madddi:
This PR updates the tool versions used in CircleCI and local acceptance tests:

- Golang 1.11 -> 1.13
- KeyCloak 8.0.0 -> 8.0.1
- `osixia/openldap:1.2.2` -> `osixia/openldap:1.3.0`
- `postgres:9.6` -> `postgres:12`

Also fixes a small typo in an error message.

This PR intentionally leaves Terraform at version 0.12.1. I tried upgrading it to 0.12.18 but this breaks a lot of tests. Error messages look different, validation is different and it seems to trigger race conditions in KeyCloak... I suggest to do this upgrade in a separate PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,8 @@ workflows:
   version: 2
   test:
     jobs:
-      - test
+      - test-7.0.1
+      - test-8.0.1
   release:
     jobs:
       - test:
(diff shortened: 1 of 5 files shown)


madddi:
@mrparkers I've modified the CI config to test multiple versions (#111)... Unfortunately CircleCI seems to be somewhat limited for this kind of use case. Maybe you can have a look at the config and tell me what you think

mrparkers:
Nice work!  I think the approach you used is solid.  Testing the latest two major versions is probably fine, I think it's reasonable to assume that most users will be using one of them.

PR #199, opened 2020-01-17
Add support for group attributes

aromeyer:
This PR adds support for group attributes following user attribute implementation.


AdrienFromToulouse:
closes https://github.com/mrparkers/terraform-provider-keycloak/issues/194

PR #202, opened 2020-01-23
Service account realm roles support

waldemarschmalz:
This PR adds support to add a realm-role to a service account. 

With `resource_keycloak_openid_client_service_account_role` it was not possible to add realm-roles. Maybe this is not the best solution, but it works. 

I added `resource_keycloak_openid_client_service_account_realm_role`.
Currently it is only possible to add a single role per resource. There are no docs yet. 

Maybe `resource_keycloak_openid_client_service_account_role` and `resource_keycloak_openid_client_service_account_realm_role` could be merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ workflows:
       - test-8.0.1
   release:
     jobs:
-      - test:
+      - test-8.0.1:
           filters:
             tags:
               only: /\d+\.\d+\.\d+(-rc.\d+)?/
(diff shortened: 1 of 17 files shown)


mrparkers:
Yeah, I agree that these two resources should eventually be merged, but there really isn't a clean way to do that right now.  Unfortunately, the way I decided to handle resource IDs when I first started writing this provider makes it difficult to use them them practically.  In this case, they could have be used to determine the difference between a realm role and a client role, so another resource wouldn't be needed.

At some point this year I plan to release a 2.0 version that completely reworks how resource IDs are treated which should simplify the code a lot, and allow for this to be condensed into a single resource.

For now I'll merge this for you so you can get the value you need.  Thanks for the PR!

waldemarschmalz:
Thanks!

PR #203, opened 2020-01-23
Support admin_url attribute in openid clients

adresan:
This PR adds support for the `admin_url` attribute when creating or updating openid clients


mrparkers:

PR #204, opened 2020-01-24
feat: support attribute access_token_lifespan for openid_client

dlechevalier:
Adding access_token_lifespan for openid_client


dlechevalier:
There's a bug with using this feature. I'll fix it and make a new PR when it will be OK.

PR #207, opened 2020-01-29
Added Consent Required support for openid client

SvenHamers:


provider/resource_keycloak_openid_client.go
```diff
@@ -146,6 +146,11 @@ func resourceKeycloakOpenidClient() *schema.Resource {
 				Optional: true,
 				Default:  true,
 			},
+			"consent_required": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
```
LoicAG: Defaulting this to false should fix the tests

PR #208, opened 2020-01-29
Added support to add client role on service account user

zikphil:
This is a gross copy/paste of PR #202 but to also add the capability of adding Client Role to Client Service Account.

Example Use-Case: Grant realm-management roles on another client's service account.

Apologies for the dirtyness of code, this is my first Go coding experience.

```
resource "keycloak_openid_client_service_account_client_role" "client_service_account_roles" {
  realm_id = keycloak_realm.test-realm.id
  client_id  = data.keycloak_openid_client.realm_management.id
  service_account_user_id  = keycloak_openid_client.test_client.service_account_user_id
  role = "manage-realm"
}
```

### DIFF
--- a/keycloak/openid_client_service_account_client_role.go
+++ b/keycloak/openid_client_service_account_client_role.go
@@ -1,0 +1,62 @@
+package keycloak
+
+import (
+	"fmt"
+)
+
+type OpenidClientServiceAccountClientRole struct {
+	Id                   string `json:"id"`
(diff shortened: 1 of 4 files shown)


waldemarschmalz:
It is already possible to add client roles to a service account user. See in example/main.tf :-)

zikphil:
Omg you are absolutely right

PR #215, opened 2020-02-11
Support for flows, subflows and executions

tomrutsaert:
support for flows, subflows and executions
(+ added missing fields in data_source_keycloak_realm)

This are the changes form PR [#138](https://github.com/mrparkers/terraform-provider-keycloak/pull/138) + adaptations needed for keycloak 8.0.x 

See comments in #138 for more info

### DIFF
--- a/keycloak/authentication_execution.go
+++ b/keycloak/authentication_execution.go
@@ -1,0 +1,144 @@
+package keycloak
+
+import (
+	"fmt"
+)
+
+// this is only used when creating an execution on a flow.
+// other fields can be provided to the API but they are ignored
+// POST /realms/${realmId}/authentication/flows/${flowAlias}/executions/execution
+type authenticationExecutionCreate struct {
+	Provider string `json:"provider"` //authenticator of the execution
+}
+
+type authenticationExecutionRequirementUpdate struct {
+	RealmId         string `json:"-"`
+	ParentFlowAlias string `json:"-"`
+	Id              string `json:"id"`
+	Requirement     string `json:"requirement"`
+}
+
+// this type is returned by GET /realms/${realmId}/authentication/flows/${flowAlias}/executions
+type AuthenticationExecution struct {
+	Id                   string `json:"id"`
+	RealmId              string `json:"-"`
+	ParentFlowAlias      string `json:"-"`
+	Authenticator        string `json:"authenticator"` //can be any authenticator from GET realms/{realm}/authentication/authenticator-providers OR GET realms/{realm}/authentication/client-authenticator-providers OR GET realms/{realm}/authentication/form-action-providers
+	AuthenticationConfig string `json:"authenticationConfig"`
+	AuthenticationFlow   bool   `json:"authenticationFlow"`
+	FlowId               string `json:"flowId"`
+	ParentFlowId         string `json:"parentFlow"`
+	Priority             int    `json:"priority"`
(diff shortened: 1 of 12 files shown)


mrparkers:
Thanks for re-opening this, I'll have some time later this week to review this again :smile:

provider/data_source_keycloak_realm.go
```diff
@@ -60,6 +60,10 @@ func dataSourceKeycloakRealm() *schema.Resource {
 				Type:     schema.TypeBool,
 				Computed: true,
 			},
+			"ssl_required": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
tomrutsaert: Changes in this file are not really related to the PR, but I saw they were missing, so I added them. see also #218

keycloak/authentication_subflow.go
```diff
@@ -0,0 +1,162 @@
...
+func (keycloakClient *KeycloakClient) getExecutionId(authenticationSubFlow *AuthenticationSubFlow) (string, error) {
+	list, err := keycloakClient.ListAuthenticationExecutions(authenticationSubFlow.RealmId, authenticationSubFlow.ParentFlowAlias)
+	if err != nil {
+		return "", err
+	}
+
+	var executionId = ""
+	for _, ex := range list {
+		if ex.FlowId == authenticationSubFlow.Id {
+			executionId = ex.Id
```
mrparkers: you could return here instead of breaking

keycloak/authentication_subflow.go
```diff
@@ -0,0 +1,162 @@
...
+			break
+		}
+	}
+	if executionId == "" {
+		return "", errors.New("no execution id found for subflow")
+	}
+	return executionId, nil
```
mrparkers: if you return in the for loop, then you can assume this fails if the code reaches this point

keycloak/authentication_execution.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (keycloakClient *KeycloakClient) RaiseAuthenticationExecutionPriority(realmId, id string) error {
+	_, _, err := keycloakClient.post(fmt.Sprintf("/realms/%s/authentication/executions/%s/raise-priority", realmId, id), nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (keycloakClient *KeycloakClient) LowerAuthenticationExecutionPriority(realmId, id string) error {
```
mrparkers: are these two functions used?  if not, how should users raise and lower the priority for executions?

tomrutsaert: There are not used, but they do work.
I left in here in the case some one worked out a better solution, and could use these methods....

Should I remove them?

keycloak/authentication_subflow.go
tomrutsaert: I will change this.

keycloak/authentication_execution.go
mrparkers: No, I was just curious how the priority was determined, but then I realized in your example that you used `depends_on` in order to ensure subflows with higher priorities were created before subflows with lower priorities.

This approach makes sense, although it would be impossible to update this aftwards.  But if that's good enough for you then I'm okay with this.  I'm hoping that this is good enough for most users until Keycloak X is released, in which case I'll probably have to rewrite a lot of this provider anyways.

PR #220, opened 2020-02-12
handle both scope and resource permission types

yspotts:
This PR allows for handling all types of permissions - resources as well as scopes


Amad27:
@mrparkers 
Hey, we updated some tests, and the example. Could you please let us know what you think?

rfahsel3:
Wow! This would be really helpful. Can we get this merged? It looks like the tests have been completed.

mrparkers:
Thanks for adding the tests, @Amad27!

I pulled this down locally and tested the changes you made to `example/main.tf`.  Everything looks good, except that running `terraform apply` and then running a following `terraform destroy` results in a failure.  This appears to only happen on this branch and not on master.  Would you or @yspotts mind taking a look at this?

Amad27:
@mrparkers I just retested it locally and was able to get it to apply and destroy without errors after running the `make user-federation-example` command. Could you please show me what error are you receiving?

mrparkers:
Hmm, I must have had something weird going on with my local setup, because I'm not running into those errors anymore.

I probably just messed something up locally.

PR #221, opened 2020-02-12
Add client authz policies

yspotts:
This PR adds all the client authorization policies keycloak provides (except for rule based which is deprecated).


Amad27:
I'll take over the tests!

Amad27:
@mrparkers I just finished the tests for the new providers, and they all pass locally. 
But for some reason the JS Policy is failing on the pipeline. We are having trouble figuring this out. 
Could you please help us diagnose this failure.

yspotts:
@mrparkers @Amad27 I was able to get more info from the server response. Here is the error:

`"error":"Script upload is disabled"` 

This is because JS policies are in feature preview and thus need to be enabled when keycloak runs. We do run that way in production, however, the image in the test pipeline does not (as that is the default). Would you prefer that we run with preview features on, or comment out the tests for the JS policy?

Amad27:
@yspotts @mrparkers
I got the tests to pass in CircleCi! 
just needed to toggle `-Dkeycloak.profile.feature.upload_scripts`

Amad27:
Examples are up too! @mrparkers could you please take another look at this pr when you get a chance.

mrparkers:
Thanks for adding the tests!  I don't have a problem with passing that flag to keycloak during tests, nice work.

I'm trying to look at the diff that GitHub is showing me, but a lot of the changes that are showing are already in master, and it's making this difficult to review.  Would you be able to fix the git history for this branch, or make a new branch and cherry-pick your commits to it so the diff only displays what files were changed?

Thanks!

Amad27:
@mrparkers i'm making a new pr to remove the merged code to make it easier to read

Amad27:
@mrparkers  just moved it to a new pr hopefully that will make it easier to read.
https://github.com/mrparkers/terraform-provider-keycloak/pull/246

mrparkers:
Closed via #246

Issue #152, opened 2019-09-05
allow for mapping roles to users and service accounts


mrparkers:
Discussion started in #143.

For mapping roles to a service account, it makes sense to make to have a data source to get the ID of that service account, and use a resource like `keycloak_user_roles` to map roles to it.  It might look something like this:

```hcl
resource "keycloak_openid_client" "client" {
  realm_id  = "${keycloak_realm.realm.id}"
  client_id = "client"
  name      = "client"

  enabled = true

  access_type   = "CONFIDENTIAL"
  client_secret = "secret"
  
  service_accounts_enabled = true
}

data "keycloak_service_account_user" "client_service_account" {
  client_id = "${keycloak_openid_client.client.id}"
}

resource "keycloak_role" "role" {
  realm_id = "${keycloak_realm.realm.id}"
  name     = "my-role"
}

resource "keycloak_user_roles" "client_service_account_roles" {
  realm_id = "${keycloak_realm.realm.id}"
  user_id  = "${data.keycloak_service_account_user.client_service_account.id}"

  roles = [
    "${keycloak_role.role.id}"
  ]
}
```

thoughts?  cc @rlewan @camelpunch

rlewan:
Totally makes sense, I like the simplicity 👍

camelpunch:
LGTM!

dimsedane:
I am looking into this a little as my first try with go and Terraform development since we need it for at project. 

While my need is also for a service account, would it be better for the datasource to be for a generic keycloak user, not just a service account? You could still get the service account user from the naming convention, but it would be possible to apply it broader.

mrparkers:
Hi @dimsedane, thanks for wanting to contribute!  Feel free to post any questions you may have here along the way, I'll be happy to help.

I had initially considered a more generic `keycloak_user` data source as well, but my concern with using that for service accounts is that we would be relying on that naming convention which isn't guaranteed to remain consistent.  I think ultimately there will be a generic `keycloak_user` data source regardless, and I suppose there would be nothing stopping people from using it that way, but I would probably still add a `keycloak_service_account_user` data source just for the extra safety.

totobest:
Thank for the idea @mrparkers.

Would it better for the resource "keycloak_openid_client" to output an attribute `service_account_user_id` if `service_accounts_enabled` is set to `true`?

languitar:
Any chance to get this going? We would need this.

mrparkers:
[...]

PR #222, opened 2020-02-13
adding user realm role assignment to user resource

becjon:
Hey, Thank you for this keycloak provider. I missed the feature to assign realm roles to user and implemented it. It would be great if you could review the PR and Merge it

I found the following issue that also references this desire.
https://github.com/mrparkers/terraform-provider-keycloak/issues/152

### DIFF
--- a/keycloak/user.go
+++ b/keycloak/user.go
@@ -189,3 +189,27 @@ func (keycloakClient *KeycloakClient) RemoveUsersFromGroup(realmId, groupId stri
 
 	return nil
 }
+
+func (keycloakClient *KeycloakClient) AddRealmLevelRoleMapping(user *User, roles []*Role) error {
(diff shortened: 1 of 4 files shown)


mrparkers:
Hi @becjon, thanks for the PR!

Did you have any thoughts around the example HCL I posted in #152 to solve this problem, specifically the `keycloak_user_roles` resource?  I liked that approach because it would allow you to map roles to Keycloak users that were federated from external sources, such as an LDAP directory.  I think that would cover more use cases than using the `keycloak_user` resource for this, since it will allow you to assign roles to users regardless of how they were created.

becjon:
Hi @mrparkers, yes you are right. An implementation in an own resource has more use-cases. Your example looks good to me. I think the state management for assigned roles would also be easier to handle in an own resource.

mrparkers:
Would you be okay with updating your implementation so that its moved to its own resource?  I think most of the code in the `keycloak` package will remain the same.  Let me know your thoughts on this.

mrparkers:
This functionality has been implemented in #315

PR #229, opened 2020-02-19
Minor cleanups

BernhardBerbuir:
- use only tabs at release-targets.json
- document missing environment variable for local test execution

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Once the environment is ready, you can run the acceptance tests after setting th
 ```
 KEYCLOAK_CLIENT_ID=terraform \
 KEYCLOAK_CLIENT_SECRET=884e0f95-0f42-4a63-9b1f-94274655669e \
+KEYCLOAK_CLIENT_TIMEOUT=5 \
 KEYCLOAK_REALM=master \
 KEYCLOAK_URL="http://localhost:8080" \
 make testacc
--- a/scripts/release-targets.json
+++ b/scripts/release-targets.json
@@ -10,7 +10,7 @@
 	{
 		"os": "linux",
 		"platform": "amd64",
-    "static": true
+		"static": true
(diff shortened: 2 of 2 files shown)


Trois-Six:
Please revert the change on GetGenericClientProtocolMappers, that breaks my terraformer provider.

BernhardBerbuir:
@Trois-Six : Could you please provide more information why this patch is problematic. Do you have a test case which currently fails, but should work?

Trois-Six:
The only place where it's used (I created that function with that goal) is here: https://github.com/GoogleCloudPlatform/terraformer/blob/master/providers/keycloak/openid_client.go#L157

BernhardBerbuir:
I'm not sure, if I understand the problem. The function `GetGenericClientProtocolMappers(realmId string, clientId string)` has two shortcomings:
* it returns only one `GenericClientProtocolMapper` instead of an array of `GenericClientProtocolMapper` (there could be more than one GenericClientProtocolMapper)
* the function does not return `GenericClientProtocolMapper`s which are bound to a `client_scope` instead of a `client`

If this is the problem, then I can create a PR, but this would change the signature of the function (one additional parameter, return value becomes an array)
=> this breaks all current usages of the function

Trois-Six:
I do not agree, it returns a OpenidClientWithGenericClientProtocolMappers, with an array of mappers associated to the oidc client.

BernhardBerbuir:
Ok, I see the array of mappers is hidden in the struct `OpenidClientWithGenericClientProtocolMappers` (side note: the name is not optimal as also SAML clients can use `GenericClientProtocolMappers`)
=> no change of the return value is required

I have seen that not a complete revert of this pull request has been done but a bug fix instead. So I'm absolutely o with the current situation.

PR #233, opened 2020-02-21
Add access token lifespan to keycloak oidc client

### LINKED ISSUE #232: Openid Client is missing the Advanced section
The resource keycloak_openid_client is missing the Advanced Settings, which are needed to set Access Token Lifespan per client. 

Did i oversaw something or is this missing?

moritz31:
This pull request fixes #232

### DIFF
--- a/provider/resource_keycloak_openid_client_test.go
+++ b/provider/resource_keycloak_openid_client_test.go
@@ -232,6 +232,32 @@ func TestAccKeycloakOpenidClient_updateInPlace(t *testing.T) {
 	})
 }
 
+func TestAccKeycloakOpenidClient_AccessToken_basic(t *testing.T) {
+	realmName := "terraform-" + acctest.RandString(10)
+	clientId := "terraform-" + acctest.RandString(10)
+
+	accessTokenLifespan := "1800"
+
+	resource.Test(t, resource.TestCase{
+		Providers:    testAccProviders,
+		PreCheck:     func() { testAccPreCheck(t) },
+		CheckDestroy: testAccCheckKeycloakOpenidClientDestroy(),
+		Steps: []resource.TestStep{
+			{
+				Config: testKeycloakOpenidClient_AccessToken_basic(realmName, clientId, accessTokenLifespan),
+				Check:  testAccCheckKeycloakOpenidClientExistsWithCorrectLifespan("keycloak_openid_client.client", accessTokenLifespan),
+			},
+			{
+				ResourceName:            "keycloak_openid_client.client",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateIdPrefix:     realmName + "/",
+				ImportStateVerifyIgnore: []string{"exclude_session_state_from_auth_response"},
+			},
+		},
+	})
+}
+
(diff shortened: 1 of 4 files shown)


provider/resource_keycloak_openid_client_test.go
```diff
@@ -859,3 +889,27 @@ resource "keycloak_openid_client" "client" {
 }
 	`, realm, clientId, standardFlowEnabled, implicitFlowEnabled, directAccessGrantsEnabled, serviceAccountsEnabled)
 }
+
+func TestAccKeycloakOpenidClient_AccessToken_basic(t *testing.T) {
```
mrparkers: nitpick: typically the actual test cases are at the top of the file.  Could you move this near the existing test functions?

provider/resource_keycloak_openid_client_test.go
```diff
@@ -622,6 +637,21 @@ resource "keycloak_openid_client" "client" {
...
+	return fmt.Sprintf(`
+resource "keycloak_realm" "realm" {
+	realm = "%s"
+}
+
+resource "keycloak_openid_client" "client" {
+	client_id   		  = "%s"
+	realm_id    		  = "${keycloak_realm.realm.id}"
+	access_type 		  = "CONFIDENTIAL"
+	access_token_lifespan = "1800"
```
mrparkers: Since this test HCL is specifically asserting against the access token lifespan, could you pass this in as an argument?

provider/resource_keycloak_openid_client_test.go
```diff
@@ -444,6 +444,21 @@ func testAccCheckKeycloakOpenidClientExistsWithCorrectProtocol(resourceName stri
...
 }
 
+func testAccCheckKeycloakOpenidClientExistsWithCorrectLifespan(resourceName string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		client, err := getOpenidClientFromState(s, resourceName)
+		if err != nil {
+			return err
+		}
+
+		if client.Attributes.AccessTokenLifespan != "1800" {
```
mrparkers: Could you pass this value as an argument?

provider/resource_keycloak_openid_client_test.go
```diff
@@ -444,6 +444,21 @@ func testAccCheckKeycloakOpenidClientExistsWithCorrectProtocol(resourceName stri
...
+			return fmt.Errorf("expected openid client to have openid-connect protocol, but got %s", client.Protocol)
```
mrparkers: Could you fix this error message?  it doesn't look like this applies here

provider/resource_keycloak_openid_client_test.go
moritz31: Done :)

provider/resource_keycloak_openid_client_test.go
moritz31: Done :)

provider/resource_keycloak_openid_client_test.go
moritz31: Done

provider/resource_keycloak_openid_client_test.go
moritz31: Sorry copy and fail

PR #237, opened 2020-02-24
feature: Allows ignoring insecure certificates

Filirom1:
Hello,

Disabling security check is dangerous and should be avoided but in development, self signed certificates are quite frequent.

Usage:
```
provider "keycloak" {
    client_id     = "admin-cli"
    username      = "keycloak"
    password      = "password"
    url           = "http://localhost:8080"
    verify_tls    = false
}
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -57,6 +57,7 @@ The following provider attributes are supported:
 - `realm` (Optional) - The realm used by the provider for authentication. Defaults to environment variable `KEYCLOAK_REALM`, or `master` if the environment variable is not specified.
 - `initial_login` (Optional) - Optionally avoid Keycloak login during provider setup, for when Keycloak itself is being provisioned by terraform. Defaults to true, which is the original method.
 - `client_timeout` (Optional) - Sets the timeout of the client when addressing Keycloak, in seconds. Defaults to environment variable `KEYCLOAK_CLIENT_TIMEOUT`, or 5 is the environment variable is not specified.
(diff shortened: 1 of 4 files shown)


docs/index.md
```diff
@@ -57,6 +57,7 @@ The following provider attributes are supported:
 - `realm` (Optional) - The realm used by the provider for authentication. Defaults to environment variable `KEYCLOAK_REALM`, or `master` if the environment variable is not specified.
 - `initial_login` (Optional) - Optionally avoid Keycloak login during provider setup, for when Keycloak itself is being provisioned by terraform. Defaults to true, which is the original method.
 - `client_timeout` (Optional) - Sets the timeout of the client when addressing Keycloak, in seconds. Defaults to environment variable `KEYCLOAK_CLIENT_TIMEOUT`, or 5 is the environment variable is not specified.
+- `verify_tls` (Optional) - Allows ignoring insecure certificates when set to false. Defaults to true. Disabling security check is dangerous and should be avoided.
```
mrparkers: Would you mind renaming this to `tls_insecure_skip_verify` so it's clear that this should be avoided if possible?  Then we can update the default value to `false`

Filirom1: Done in https://github.com/mrparkers/terraform-provider-keycloak/pull/237/commits/60df991d9aa63891ef8c86f6a11a18f575fa1ed5

Filirom1:
I have tested the provider after the changes and the merge. It works well.

Issue #231, opened 2020-02-20
resource_keycloak_oidc_identity_provider does not support setting up a Google IDP


tomrutsaert:
It is not possible to set up a google IDP with the resource_keycloak_oidc_identity_provider.

The _kc oidc idp resource_ assumes certain things that are not valid for most Identity Provider implementations that extend from IdentityProvider.java interface in the keycloak code base.
Example: GoogleIdentityProvider.java

In case of the _google Idp_ I found that:
============================
- authorization_url: is required by the _oidc idp resource_, but not used by _google idp_ in UI
- token_url: is required by the _oidc idp resource_, but not used by _google idp_ in UI
- user_info_url: is required by the _oidc idp resource_, but not used by _google idp_ in UI, It does need to be set to "https://openidconnect.googleapis.com/v1/userinfo" otherwise GoogleIdentityProvider code will not work as expected. + You need to set it, as _oidc idp resource_ uses this field to determine the value of disableUserInfo. For the google idp you can set this field independently.
- jwks_url: this field needs to be set, If you want to set _use_jwks_url to be true. But I google idp does not need to set a jwks_url parameter.
- Many other fields have default values, where google idp does not expect them.
For some of these fields I did not find a workaround with as result that: I can not make my _google idp_ config work through terraform.

Possible solutions:
==============
1. make a separate _google idp resource_. (downside: someone else still need to make one for each social provider)
2. Make the generic idp resource more lax. 
     - Do not assume settings based on other settings, ex: use_jwks_url and jwks__url
     - Remove all default values(except provider_id)
     - Remove all required fields(except for realm, provider_id, and alias)
(downside: backwards compatibility for people already using this resource and expect the default values)

@mrparkers and @AndrewChubatiuk, What are your opinions?
Would it best to make a PR for solution 1 or 2?

@nagytech I saw in another issue(#141) your were setting u a google idp, Did you got it working, if yes how? Do you have an example?

mrparkers:
I don't really have any experience using this particular feature of Keycloak, but generally I am in favor of having separate resources for this so we can be more explicit.  I agree that making changes to the generic idp resource might introduce changes that affect current users, so I like the idea of making a separate google idp resource.

nagytech:
@tomrutsaert 

[...]

PR #240, opened 2020-02-27
Support for google idp in a separate resource

tomrutsaert:
Support for google idp in a separate resource


tomrutsaert:
See  #231

tomrutsaert:
See  #231

mrparkers:
The code looks good to me!  I'll trust you on this working as expected, since I don't have the means to test it out myself.

Thanks for the PR!

PR #241, opened 2020-02-29
Add support for authentication execution configuration

ndrpnt:
Building on #215, this PR adds support for configuring executions.

I doubt that this is the best implementation as I am a keycloak/terraform newbie, but it seems good enough for my use case.

### DIFF
--- a/docs/resources/keycloak_authentication_execution_config.md
+++ b/docs/resources/keycloak_authentication_execution_config.md
@@ -1,0 +1,53 @@
+# keycloak_authentication_execution_config
+
+Allows for managing an authentication execution configuration.
+
+### Example Usage
+
+```hcl
+resource "keycloak_realm" "realm" {
+	realm   = "my-realm"
+	enabled = true
+}
+
+resource "keycloak_authentication_flow" "flow" {
+	realm_id = "${keycloak_realm.realm.id}"
+	alias    = "my-flow-alias"
+}
+
+resource "keycloak_authentication_execution" "execution" {
+	realm_id          = "${keycloak_realm.realm.id}"
+	parent_flow_alias = "${keycloak_authentication_flow.flow.alias}"
+	authenticator     = "identity-provider-redirector"
+}
+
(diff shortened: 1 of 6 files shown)


mrparkers:
Hi @ndrpnt, thanks for the PR!  My only thought here is that it may be more user-friendly to define `config` as a new attribute on the `keycloak_authentication_execution` resource instead of having its own resource.  What are your thoughts on this?

ndrpnt:
I could see that being more elegant for the user, but slightly more complex in the code as it doesn't stick to the Keycloak API (one Terraform resource would manage two Keycloak resources). I can give it a try.

ndrpnt:
Hi @mrparkers, getting back to you after trying to merge `execution` and `execution config` in a single Terraform resource. It is more complex than I anticipated, and I would rather not waste more time on it, as I don't think this is the best solution.

My unfinished attempt is [here](https://github.com/ndrpnt/terraform-provider-keycloak/tree/merge_execution_config). I may be missing something, but I feel like it makes the code more complicated than it needs to be. I ended up in some sort of typecasting/mapping hell (or is it what any provider looks like?) which makes me wonder if I am heading in the right direction.
This is to be compared with the separate resources implementation which was super simple.

Some problems I faced:
- CRUD operation can be successful on one resource and fail on the other (implies the use of partial state ?)
- Lifecycle of the resources differ: we should ideally be able to add/update/delete an `execution config` without affecting the parent `execution`. This is some logic that we delegate to the Terraform framework when splitting the resources
- The client offers two separate datastructures (`AuthenticationExecution` and `AuthenticationExecutionConfig`) that I have trouble mapping to a single `ResourceData` struct (maybe using an intermediate datastructure would help?)

What do you think?

mrparkers:
Thanks for your thoughts!  I didn't think my suggestion would bring that many challenges, but I understand that it's difficult to manage this in one terraform resource with the Keycloak API presents them as two separate resources.

I'm fine with merging your PR as-is.

Thanks for the contribution!

PR #246, opened 2020-03-04
Add client auth policies

Amad27:
moved @yspotts  pr https://github.com/mrparkers/terraform-provider-keycloak/pull/221


example/client_authorization_policys.tf
```diff
@@ -0,0 +1,174 @@
+resource keycloak_realm test {
```
mrparkers: I think most of these examples are going to conflict with the pre-existing `main.tf` file in the `example` folder.  Could you rename the resources in this file so they don't conflict with `main.tf`?  It's okay to hardcode any strings you want here, these files really just serve as an example of how to use the provider.

keycloak/keycloak_client.go
```diff
@@ -280,6 +281,7 @@ func (keycloakClient *KeycloakClient) sendRequest(request *http.Request) ([]byte
 	}
 
 	if response.StatusCode >= 400 {
+		fmt.Println(string(body), "Error with the response")
```
mrparkers: We're already printing the body on line 280, is there a reason this is needed?

In case you weren't aware, you can set the environment variable `TF_LOG` to `DEBUG` to enable debug logging which should print out a bunch of http request/response information.

provider/resource_keycloak_openid_client_authorization_aggregate_policy_test.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+func TestResourceKeycloakOpenidClientAuthorizationAggregatePolicy(t *testing.T) {
+	realmName := "terraform-" + acctest.RandString(10)
+	clientId := "terraform-" + acctest.RandString(10)
+
+	resource.Test(t, resource.TestCase{
+		Providers: testAccProviders,
+		PreCheck:  func() { testAccPreCheck(t) },
+		// CheckDestroy: testResourceKeycloakOpenidClientAuthorizationAggregatePolicyDestroy(),
```
mrparkers: is there a reason this is commented out?

provider/resource_keycloak_openid_client_authorization_aggregate_policy_test.go
```diff
@@ -0,0 +1,127 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testResourceKeycloakOpenidClientAuthorizationAggregatePolicy_basic(realmName, clientId),
+				Check:  testResourceKeycloakOpenidClientAuthorizationAggregatePolicyExists("keycloak_openid_client_aggregate_policy.test"),
+			},
+		},
+	})
+}
+
+func testResourceKeycloakOpenidClientAuthorizationAggregatePolicy_basic(realm, clientId string) string {
```
mrparkers: nitpick: could you move this function to the bottom of this file?

provider/resource_keycloak_openid_client_authorization_aggregate_policy_test.go
```diff
@@ -0,0 +1,127 @@
...
+func testResourceKeycloakOpenidClientAuthorizationAggregatePolicyDestroy() resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		for _, rs := range s.RootModule().Resources {
+			if rs.Type != "keycloak_openid_client_aggregate_policy" {
+				continue
+			}
+
+			realm := rs.Primary.Attributes["realm"]
+			resourceServerId := rs.Primary.Attributes["resourceServerId"]
+			policyId := rs.Primary.Attributes["policyId"]
```
mrparkers: these attributes should be the same as the attributes available on the resource.  so `resourceServerId` should be `resource_server_id`.  and it doesn't look like `policyId` is a valid attribute, perhaps you're looking for `rs.Primary.ID` here

example/client_authorization_policys.tf
mrparkers: You are also free to reuse the resources that have already been created in `main.tf`, meaning that you can just leave them in that file and use variable interpolation to depend on them (like you are on line 12).

provider/resource_keycloak_openid_client_authorization_aggregate_policy.go
```diff
@@ -0,0 +1,185 @@
...
+			},
+			"decision_strategy": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"owner": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"logic": {
```
mrparkers: Looks like this can only be `POSITIVE` or `NEGATIVE`, could you add some validation here to enforce this?  I think you can do this for all of the policy resources as well.

provider/resource_keycloak_openid_client_authorization_aggregate_policy.go
```diff
@@ -0,0 +1,185 @@
...
+	keycloakClient := meta.(*keycloak.KeycloakClient)
+
+	realmId := data.Get("realm_id").(string)
+	resourceServerId := data.Get("resource_server_id").(string)
+	id := data.Id()
+
+	return keycloakClient.DeleteOpenidClientAuthorizationAggregatePolicy(realmId, resourceServerId, id)
+}
+
+func resourceKeycloakOpenidClientAuthorizationAggregatePolicyImport(d *schema.ResourceData, _ interface{}) ([]*schema.ResourceData, error) {
```
mrparkers: It looks like this import function is basically the same as all of the other import functions for the policy resources - could you reuse the same function for all of them if possible?

Take a look at how I did this with protocol mappers: https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/generic_protocol_mapper_helpers.go

provider/resource_keycloak_openid_client_authorization_aggregate_policy.go
```diff
@@ -0,0 +1,185 @@
...
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"decision_strategy": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"owner": {
```
mrparkers: I'm a bit confused on what the `owner`, `resources`, and `scopes` attributes are supposed to represent.  I see these attributes exist for all of the policy resources but I can't seem to find them in the Keycloak GUI, nor do I see these being used in the `client_authorization_policies.tf` file that you added.  Could you help me understand why these are needed?

provider/resource_keycloak_openid_client_authorization_client_policy.go
```diff
@@ -0,0 +1,200 @@
...
+			},
+			"owner": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"logic": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"policies": {
```
mrparkers: Is the `policies` attribute needed for this policy?  I thought it was only needed for the aggregate policy resource.

provider/resource_keycloak_openid_client_authorization_group_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testResourceKeycloakOpenidClientAuthorizationGroupPolicy_basic(realmName, clientId),
+				Check:  testResourceKeycloakOpenidClientAuthorizationGroupPolicyExists("keycloak_openid_client_group_policy.test"),
+			},
+		},
+	})
+}
+
+func testResourceKeycloakOpenidClientAuthorizationGroupPolicy_basic(realm, clientId string) string {
```
mrparkers: nitpick: can you move this to the bottom of the file?

provider/resource_keycloak_openid_client_authorization_js_policy_test.go
```diff
@@ -0,0 +1,111 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testResourceKeycloakOpenidClientAuthorizationJSPolicy_basic(realmName, clientId),
+				Check:  testResourceKeycloakOpenidClientAuthorizationJSPolicyExists("keycloak_openid_client_js_policy.test"),
+			},
+		},
+	})
+}
+
+func testResourceKeycloakOpenidClientAuthorizationJSPolicy_basic(realm, clientId string) string {
```
mrparkers: nitpick: can you move this to the bottom of the file?

provider/resource_keycloak_openid_client_authorization_role_policy.go
```diff
@@ -0,0 +1,193 @@
...
+			},
+			"type": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"role": {
```
mrparkers: I can see in the GUI that this policy supports realm roles and client roles, which is this attribute supposed to represent?  And it looks like I can add more than one role to the policy, is there a reason you specified `MinItems` to be 1?

provider/resource_keycloak_openid_client_authorization_time_policy_test.go
```diff
@@ -0,0 +1,123 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testResourceKeycloakOpenidClientAuthorizationTimePolicy_basic(realmName, policyName, clientId),
+				Check:  testResourceKeycloakOpenidClientAuthorizationTimePolicyExists("keycloak_openid_client_time_policy.test"),
+			},
+		},
+	})
+}
+
+func testResourceKeycloakOpenidClientAuthorizationTimePolicy_basic(realm, policyName, clientId string) string {
```
mrparkers: nitpick: can you move this to the bottom of the file?

provider/resource_keycloak_openid_client_authorization_user_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testResourceKeycloakOpenidClientAuthorizationUserPolicy_basic(realmName, clientId),
+				Check:  testResourceKeycloakOpenidClientAuthorizationUserPolicyExists("keycloak_openid_client_user_policy.test"),
+			},
+		},
+	})
+}
+
+func testResourceKeycloakOpenidClientAuthorizationUserPolicy_basic(realm, clientId string) string {
```
mrparkers: nitpick: can you move this to the bottom of the file?

provider/resource_keycloak_openid_client_authorization_role_policy.go
RyanAtWare2Go: I just checked the Chrome network inspector when creating a role policy, and I see that while bifurcated in the UI, the API call combines all the roles into one object:
image

Note the request above had one realm role and one client role.

RyanAtWare2Go: I think the reason for min items equaling 1 is because a role policy essentially does nothing if there are no roles specified.

Amad27: Thanks Ryan! 
@mrparkers I can remove the `MinItems` if you'd like.

Amad27:
@mrparkers I made the requested changes! Could you please take another look.

provider/resource_keycloak_openid_client_authorization_role_policy.go
mrparkers: My bad, I misread `MinItems` to be `MaxItems`.  Thanks for the clarification on the roles.

mrparkers:
The changes you made so far look good, although there are still 5 other comments I left that I didn't see changes for.

I think GitHub collapses comments when there are too many, I had to click "Load More" on the part of the page that says "5 hidden conversations"

provider/resource_keycloak_openid_client_authorization_aggregate_policy.go
Amad27: They're not! I'll remove them.

Amad27:
Woops! Just made the last 5 changes. Let me know what you think. 
@mrparkers

provider/resource_keycloak_openid_client_authorization_aggregate_policy.go
mrparkers: Sorry, I should have been more clear with this.  I saw these attributes on all of the policy resources you added, but I only left the comment on this file.  Could you take a look at all of the other policy resources and see if these attributes are needed or used?  If not, could you remove them and also remove them from the corresponding struct in the `keycloak` package?

provider/validation_helpers.go
```diff
@@ -0,0 +1,13 @@
+package provider
+
+import (
+	"fmt"
+)
+
+func logicKeyValidation(val interface{}, key string) (warns []string, errs []error) {
```
mrparkers: Thanks for adding this!  Although I probably should have let you know about the `validation` package which gives you the `StringInSlice` function that could have made this implementation easier.  Take a look at this example: https://github.com/mrparkers/terraform-provider-keycloak/blob/f4770b0a3726a82d0fd0176cba328b4b2dda1389/provider/resource_keycloak_ldap_user_federation.go#L65

Could you change this to use this function?

Amad27: Oh cool! Just replaced that with the validation package. 👍

mrparkers:
I left another comment that might be hard to see since it's in the collapsed conversation: https://github.com/mrparkers/terraform-provider-keycloak/pull/246#discussion_r389742260

provider/resource_keycloak_openid_client_authorization_aggregate_policy.go
Amad27: No worries! I removed all the unnecessary attributes.

PR #247, opened 2020-03-07
WIP: User roles mapping

moritz31:
Hey mrparkers,

this is a draft of a user-roles-mapping resource which is based on @becjon works.
I've currently faced the issue that i'm not really sure what data.SetId does and if i set it to something combined it will try to retrive the user with that name instead of using user_id. Maybe you can tell why this happen. If i remove the read Function all is working like expected.


mrparkers:
Hey @moritz31, nice progress so far.

`data.SetId` is used to give the resource a unique identifier which can be used to query for the resource in the future.  In theory, the ID should contain all of the information that is needed for the provider to fetch this resource from Keycloak, so typically the ID is some strings separated by a forward slash.  The ID should also be unique, meaning that no two instances of the same resource can use the same ID.

In most cases, a resource's ID will at least contain the unique ID of the resource on the Keycloak side, which is typically a guid.  In your case, however, you aren't actually creating a new resource in Keycloak, but mapping two existing resources together.  So you have a bit more flexibility in terms of what your ID can be.

If you'd like, you can take a look at the [`keycloak_group_roles`](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/resource_keycloak_group_roles.go) resource for some inspiration.   
This resource is very similar to the resource you're implementing - its associating roles with a group instead of a user. The ID I used here was `${realmId}/${groupId}`, since this resource will only ever be used once per group, so I could assume that this combination would be unique for each instance of the resource, and it contains all of the information I need to perform a read operation.

Hopefully that clears some things up for you.  Since the `keycloak_group_roles` resource will behave similarly to the resource you're trying to implement, feel free to borrow as much code / tests as you'd like for this resource.

languitar:
Any chance to get this rolling? We would need this ;-)

mrparkers:
This functionality has been implemented in #315

PR #253, opened 2020-03-16
Role Scopes: Allow specifying client scope instead of client

dmeyerholt:
My first shot at extending this great provider.
Following https://github.com/mrparkers/terraform-provider-keycloak/pull/242 In addition to clients, I want to extend the keycloak_generic_client_role_mapper resource to also manage client roles for client scopes. This allows consolidation of client roles in dedicated client scopes.
Im no go developer and still need to extend the 
* validation
* test cases
* documentation

### DIFF
--- a/example/roles.tf
+++ b/example/roles.tf
@@ -15,6 +15,13 @@ resource "keycloak_openid_client" "pet_api" {
   access_type = "BEARER-ONLY"
 }
 
+// Optional client scope for mapping additional client role
+resource "keycloak_openid_client_scope" "extended_pet_details" {
+  realm_id  = "${keycloak_realm.roles_example.id}"
+  name      = "extended-pet-details"
+  description = "Optional scope offering additional information when getting pets"
+}
+
 resource "keycloak_role" "pet_api_create_pet" {
   name        = "create-pet"
   realm_id    = "${keycloak_realm.roles_example.id}"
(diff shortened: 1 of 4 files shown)


mrparkers:
Hey @dmeyerholt, thanks for opening this PR.  Your changes look good so far.

I could be wrong, but I don't think you need to add any more validation on top of what you already have -  all I would have expected to see was the `ConflictsWith` on the client id and client scope id attributes.

For the test cases, you should be able to copy/paste a lot of code in the `resource_keycloak_generic_client_role_mapper_test.go` file and replace the `client_id` attribute with `client_scope_id` in the HCL snippets, and add new assertions to make sure the correct value is set after `tf apply` runs.  I'd also appreciate it if you could add an example HCL configuration within `example/main.tf`.

Let me know if you need any assistance with this.

dmeyerholt:
Thanks @mrparkers,
* validation using `ConflictsWith` should be sufficient, right
* will provide more testcases
* additional HCL example coming up as well
Hoping to finish in the next days :)

dmeyerholt:
Looks okay so far. for documentation I extended the pet examples in roles.tf. For that to work properly those changes were done
* on the app client: disable full scope
* Now map all pet_api client roles into the app client (realm role missing though, maybe another pr ;) )
* new optional client scope makes it possible to map the (new) client role `read-pet-with-details` into the tokens only, when that client scope is requested in the token request.

PR #256, opened 2020-03-19
Allow changing the email verification state

languitar:
Add the email_verified flag for users.


languitar:
Now also with correct formatting ;-)

PR #259, opened 2020-03-23
fix: revert patch on GetGenericClientProtocolMappers

Trois-Six:
Revert the patch on that function, as it changes its behaviour, and my terraform provider is not working anymore. I will support ClientScopes someday, but for the moment my code is not prepared for that, and it's not the good way to support it.

### DIFF
--- a/keycloak/generic_client_protocol_mapper.go
+++ b/keycloak/generic_client_protocol_mapper.go
@@ -33,10 +33,10 @@ func (keycloakClient *KeycloakClient) NewGenericClientProtocolMapper(genericClie
 	return nil
 }
 
-func (keycloakClient *KeycloakClient) GetGenericClientProtocolMappers(realmId string, clientId string, clientScopeId string, mapperId string) (*OpenidClientWithGenericClientProtocolMappers, error) {
+func (keycloakClient *KeycloakClient) GetGenericClientProtocolMappers(realmId string, clientId string) (*OpenidClientWithGenericClientProtocolMappers, error) {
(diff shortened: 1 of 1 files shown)


Trois-Six:
broken by https://github.com/mrparkers/terraform-provider-keycloak/pull/229

mrparkers:
Hi @Trois-Six

I'm sorry that this change broke your dependency on the `keycloak` package within this repo, but I can't revert this PR since it has already been included in a release.  Reverting this would be a breaking change to the provider, and I'm not willing to do that for a dependency that's out of my control.

If you're using Go Modules to import the `keycloak` package, you can try to import the `1.16.0` tag, which does not contain the code you're attempting to revert here.

Alternatively, you may submit a PR that corrects the issue you're having within the `keycloak` package, as long as it doesn't cause any breaking changes to the provider schema.

Unfortunately, I can't guarantee that the `keycloak` package will never be updated with breaking changes in the future.  It was never intended to be public.  However, I'd be willing to accept a PR that adds a test in CI to ensure that the dependency you have on this package doesn't break with future PRs.

Trois-Six:
but that function is not used in the provider ?! and the modification makes it completely unusable.

Trois-Six:
that PR is not a revert, but a patch, and only on the function, I am not reverting everyting.

mrparkers:
Apologies!  For some reason the diff that GitHub was showing me had 300+ changes, but after refreshing this PR it showed your real changes.

Since this function is apparently not used and tests are still green I'm okay with this PR.

Sorry again for the confusion here.

PR #263, opened 2020-03-26
feat: add saml client installation provider

Filirom1:
Hello,

I need to retrieve the SAML installation XML to inject it in AWS SAML Provider

```hcl
    realm   = "my-realm"
    enabled = true
}

resource "keycloak_saml_client" "saml_client" {
    realm_id                = "${keycloak_realm.realm.id}"
    client_id               = "test-saml-client"
    name                    = "test-saml-client"

    sign_documents          = false
    sign_assertions         = true
    include_authn_statement = true

    signing_certificate = "${file("saml-cert.pem")}"
    signing_private_key = "${file("saml-key.pem")}"
}

data "keycloak_saml_client_installation_provider" "saml_idp_descriptor" {
  realm_id    = "${keycloak_realm.realm.id}"
  client_id   = "${keycloak_saml_client.saml_client}"
  provider_id = "saml-idp-descriptor"
}

resource "aws_iam_saml_provider" "default" {
  name                   = "myprovider"
  saml_metadata_document = data.keycloak_saml_client_installation_provider.saml_idp_descriptor.value
}
```

I started this Pull Request that is working, but I will need help if you need tests or refactoring because I am not fluent in golang yet :-)

Cheers
Romain

### DIFF
--- a/docs/data_sources/keycloak_saml_client_installation_provider.md
+++ b/docs/data_sources/keycloak_saml_client_installation_provider.md
@@ -45,7 +45,7 @@ resource "aws_iam_saml_provider" "default" {
 The following arguments are supported:
 
 - `realm_id` - (Required) The realm this group exists within.
-- `client_id` - (Required) The name of the saml client
+- `client_id` - (Required) The name of the saml client. Not the id of the client.
 - `provider_id` - (Required) Could be one of `saml-idp-descriptor`, `keycloak-saml`, `saml-sp-descriptor`, `keycloak-saml-subsystem`, `mod-auth-mellon`
 
 ### Attributes Reference
--- a/docs/data_sources/keycloak_saml_client_installation_provider.md
+++ b/docs/data_sources/keycloak_saml_client_installation_provider.md
@@ -5,6 +5,9 @@ of a SAML Client.
 
 ### Example Usage
 
+In the example below, we extract the SAML metadata IDPSSODescriptor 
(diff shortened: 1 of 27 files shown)


mrparkers:
Hey @Filirom1, thanks for the PR!

I think the code you've submitted looks good so far.  I really like the example you posted in this PR, that's a good use case for this data source.  If you'd like, you could include that example in the docs to show off a real use case for this.

I've never personally used the "Installation" tab for a SAML client before, but based on the code you've submitted, it looks like you're expecting an XML document to be returned in the `value` attribute.  I think a good test for this data source would create a realm, a saml client, and a saml installation provider data source (just like your example), and grab the `value` attribute from the data source and use golang's `xml` package to parse it.  You'll have to create a basic `struct` with xml tags, but you don't have to include the entire structure of the xml payload if you don't want to.  All I want to assert is that the `value` attribute is an xml document.

You can look at examples for parsing xml in the `xml` package godoc: https://golang.org/pkg/encoding/xml/.  Feel free to model your test similarly to the other data source tests.

If you have any specific questions about any of this please feel free to let me know.  Thanks!

Filirom1:
Is it OK for you @mrparkers ?

jurgen-weber-deltatre:
so for setting up a SAML client for AWS...

>     signing_certificate = "${file("saml-cert.pem")}"
>    signing_private_key = "${file("saml-key.pem")}"

what are these files? Where do I get them?

PR #265, opened 2020-04-01
feat: add ldap role mapper

Trois-Six:
add ldap role mapper


Trois-Six:
Hi, what do you think about this PR ?

PR #267, opened 2020-04-03
Added ldap_user_federation attribute trust_email

### LINKED ISSUE #234: Missing attribute "trustEmail" in keycloak_ldap_user_federation
The "keycloak_ldap_user_federation" resource is missing the attribute for the GUI option "Trust Email". See https://mrparkers.github.io/terraform-provider-keycloak/resources/keycloak_ldap_user_federation/

A change in the GUI puts the following JSON snippet in the "config" map:

        "trustEmail": [
          "true"
        ],

lathspell:
Something strange happened with Pull Request 251, one of my git-rebase actions seems to have closed it?

The problem with Keycloak 7.0.1 is that the GUI does not support the "trust_email" attribute in this version. It is not in the Admin Webinterface nor in the representation that is reported back in the API.

How do you usually cope with such attributes? Is there a good way to throw an error depending on the version? If you can give me an example, I could try to enhance the patch.

  

original Text from Pull Request 251:
---------------------------------------------------------------------
fixes #234

It's a simple boolean value, I copy & paste fixed it whereever I found ValidatePasswordPolicy/validate_password_policy.

Verified that it can in fact turn the setting on and off in Keycloak 8.0.1.

### DIFF
--- a/provider/resource_keycloak_ldap_user_federation_test.go
+++ b/provider/resource_keycloak_ldap_user_federation_test.go
@@ -226,6 +226,9 @@ func TestAccKeycloakLdapUserFederation_basicUpdateAll(t *testing.T) {
 	firstValidatePasswordPolicy := randomBool()
 	firstPagination := randomBool()
 
+	firstTrustEmail := randomBool()
+	secondTrustEmail := !firstTrustEmail
+
 	firstConnectionTimeout, _ := keycloak.GetDurationStringFromMilliseconds(strconv.Itoa(acctest.RandIntRange(1, 3600) * 1000))
 	secondConnectionTimeout, _ := keycloak.GetDurationStringFromMilliseconds(strconv.Itoa(acctest.RandIntRange(1, 3600) * 1000))
 	firstReadTimeout, _ := keycloak.GetDurationStringFromMilliseconds(strconv.Itoa(acctest.RandIntRange(1, 3600) * 1000))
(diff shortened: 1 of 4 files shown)


mrparkers:
That makes sense, thanks for the explanation.

I went ahead and updated CI to pass a `KEYCLOAK_VERSION` environment variable when running the tests.  Could you update the failing test to check this using `os.Getenv` and skip the test if the version is set to `7.0.1`?  Also, if you update the docs for this new attribute to say that it only works on Keycloak 8+ that would be great.

Everything else LGTM.

lathspell:
I added a utility function that extracts the Keycloak major version from the new environment variable KEYCLOAK_VERSION.

The test sets the first and second value for "trustEmail" to false if the Keycloak version is <8. This should work as the variable is fetched using `parseBoolAndTreatEmptyStringAsFalse()`.

The test still fails though and I cannot figure out why. In the output I see changes in `batch_size_for_sync`, `user_object_classes`and `validate_password_policy` which are all totally unrelated to `trustEmail`.

I could not figure out how to do single step debugging into the test (IntellIJ just stopped) or where exactly this output comes from. So you need to help a bit here.

provider/resource_keycloak_ldap_user_federation_test.go
```diff
@@ -128,6 +129,15 @@ func TestAccKeycloakLdapUserFederation_basicUpdateAll(t *testing.T) {
 	firstValidatePasswordPolicy := randomBool()
 	firstPagination := randomBool()
 
+	var firstTrustEmail, secondTrustEmail bool
+	if keycloakMajorVersion(t) >= 8 {
+		firstTrustEmail = randomBool()
+		secondTrustEmail = !firstTrustEmail
+	} else {
+		firstTrustEmail = false
+		secondTrustEmail = false
```
mrparkers, line 138: Looks like this is still failing.

It might be best to remove the `trustEmail` stuff from this test and make a separate test specifically for testing this attribute.  Then we can skip that test if the version of Keycloak doesn't support it.

lathspell:
I rebased the patch to the current master branch.
As Keycloak 7.0.1 is no longer tested, the test case now runs perfectly :-)

mrparkers:
Sorry for just now getting back to this @lathspell.

Since we don't need to look at the `KEYCLOAK_VERSION` environment variable anymore, would you mind updating this PR to remove the checking for this?

lathspell:
I rebased the patch. Is it ok to merge now?

mrparkers (approved):

Issue #216, opened 2020-02-11
Realm id


Trois-Six:
Hi,

Let's say that you create a realm with the Keycloak UI, and then you change it's name from the UI. You will have in the API two different values for realm.Id and realm.realm (in my case Foo vs foo).

In provider/resource_keycloak_realm.go you use the realm attribute to populate the Id. I think that it's not good.

```
--- resource_keycloak_realm.go.orig	2020-02-11 15:24:45.328010093 +0100
+++ resource_keycloak_realm.go	2020-02-11 15:25:05.500152234 +0100
@@ -446,7 +446,7 @@
 	}
 
 	realm := &keycloak.Realm{
-		Id:          data.Get("realm").(string),
+		Id:          data.Get("id").(string),
 		Realm:       data.Get("realm").(string),
 		Enabled:     data.Get("enabled").(bool),
 		DisplayName: data.Get("display_name").(string),
```

I imported that resource and I am using the resulting variable "keycloak_realm.realm_foo.id" to set the realm_id attribute of a keycloak_ldap_user_federation resource, and it's not working (I don't see the resource in the GUI, although I don't have any error in terraform).

Regards,

Pierre

Trois-Six:
I know that sometimes in the API we have to use the realm.Id and sometimes the realm.realm, and I am not sure about the impacts of such a change. I have seen that because I am using my keycloak terraformer provider, and the realm.Id didn't match any "real" Realm Id.

dmeyerholt:
Hi i also observed related behaviour when importing a custom user federation. during import, the realm_id attribute is correctly used inside the imported federation resource (using the id of the realm). however trying to apply that (again) the resource is not found and will be recreated with another realm-id (the realm name) and even does not show up in keycloaks ui (as the provider is bound to another (non existing) realm.

mrparkers:
When you create realms in Keycloak, the ID for the realm is the same as its name, so typically those two attributes are the same and can be used interchangeably.

Honestly, I have never tried to change a realm's name before, because I wasn't sure what kind of impact that would have.  For example, the issuer value that Keycloak uses in the OIDC well-known docs uses the name of the realm, so changing this would break authentication for any applications that are hooked up the the realm.

There's a good chance I made some incorrect assumptions when originally writing this, as I was a little confused by Keycloak's decision to use the realm name as its unique ID as opposed to a GUID which is used for every other resource's unique ID.

Anyways, there's two things going on here:

[...]

PR #270, opened 2020-04-08
Clarify realm id / name and expose realm id as new realm attribute

dmeyerholt:
Coming from discussions here #216 my shot at clarifying usage of realm name and id.
Main points are:
* the requests to the API require the usage of the realm name, not the id. this is a rather cosmetic change in `keycloak/realm.go`
* in the realm resource, now the "real" realm id is exposed as "internal_id" attribute
* we retain the default behaviour of creating a realm with realmId == realmName so nothing breaks new/existing realms created by the provider
* This offers the possibility to import existing realms which have a realm id that differs from the realm's name. This is relevant for possible keycloak component which use the realm id for the parent_id attributes. Additional PR coming for the custom_user_federation if this PR gets through. I think, the ldap stuff is also affected by possible id and name mismatches (through the use of parent_id attributes)

### DIFF
--- a/docs/resources/keycloak_realm.md
+++ b/docs/resources/keycloak_realm.md
@@ -174,6 +174,12 @@ The `brute_force_detection` block supports the following attributes:
 
 Map, can be used to add custom attributes to a realm. Or perhaps influence a certain attribute that is not supported in this terraform-provider
 
+### Attributes Reference
+
+In addition to the arguments listed above, the following computed attributes are exported:
+
+- `internal_id` - When importing realms created outside of this terraform provider, they could use generated arbitrary IDs for the technical realm id. Realms created by this provider always use the realm's name for its technical id.
+
 ### Import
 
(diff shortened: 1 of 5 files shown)


mrparkers:
This looks good so far.  Here are a few changes I'd also like to see before I'm ready to merge this:

- Add a test that creates a `keycloak_realm` and asserts that its `internal_id` attribute is not the same as its name.
- Add this property to the `keycloak_realm` data source.
- Document this new attributes on the doc pages for the `keycloak_realm` resource and data source.

Let me know if you need help with any of these things.

Thanks!

dmeyerholt:
hey @mrparkers , 
had time to finish this pr with your suggestions. Skipped documentation of data source, as that doc references the realm doc for all supported attributes.
Hope you are fine with this now.
will start new pr for improving the custom_userfederation resource to leverage this new internalId attribute.

PR #272, opened 2020-04-08
add Aggregate attribute values  to User Attribute mapper

arminfelder:
adds support for the Aggregate Attribute Values setting


provider/resource_keycloak_openid_user_attribute_protocol_mapper.go
```diff
@@ -84,6 +84,12 @@ func resourceKeycloakOpenIdUserAttributeProtocolMapper() *schema.Resource {
 				Default:      "String",
 				ValidateFunc: validation.StringInSlice([]string{"JSON", "String", "long", "int", "boolean"}, true),
 			},
+			"aggregate_attrs": {
```
mrparkers, line 87: this is a nitpick, but I'd prefer if this was called `aggregate_attributes`

arminfelder: ok, done

arminfelder:
don't mind, thanks for merging

PR #275, opened 2020-04-19
support for deployed js policies

alevit33:
With this little change, it's possible to create js policies from already deployed policies. The field "code" of the resource "keycloak_openid_client_js_policy" can now be 2 things:
- javascript code, so that a brand new policy is created
- a javascript file already deployed, so that a new policy is created starting from this file.
This is recommended since in the last version of Keycloak uploading js code is disabled by default, but it's still possible for backwards compatibility.

Tested with Keycloak v9.0.3 and Terraform v0.12.24


alevit33:
Also added support for User Managed Access at realm level

mrparkers:
Awesome, can you update the docs for the `keycloak_openid_client_js_policy` resource to explain this change, and run `make fmt` on your branch please?

alevit33:
Ok, `make fmt` done.
The docs for `keycloak_openid_client_js_policy` doesn't exist yet, as well as the docs for all the client policies. They should be created, but honestly I'm not sure to know all the fields... For now, I've added a note in the example file `example/client_authorization_policys`

Issue #238, opened 2020-02-25
keycloak_realm is missing defaultSignatureAlgorithm


lathspell:
They keycloak_realm resource is missing the attribute "defaultSignatureAlgorithm" from https://www.keycloak.org/docs-api/5.0/rest-api/index.html#_realmrepresentation that is needed to change the Algo from HS256 to RS256 as some framekworks like Micronaut do not understand HS256 (at least in the version I've encountered).

lathspell:
This issue can probably be closed as you merged https://github.com/mrparkers/terraform-provider-keycloak/pull/282 in June?

mrparkers:
You're correct, thanks for helping me clean up :smile:

PR #282, opened 2020-05-08
GH-238 Added 'default_signature_algorithm' to Realm config

lathspell:
Please have a look. I verified that "RS256" appears in the Keycloak Web GUI and that "" resets it to the default of an empty HTML Select input element.


tomrutsaert:
Fmt check is failing.

Please run 'make fmt' and commit.

Issue #291, opened 2020-05-18
creating resource keycloak_user with federated_identity fails with keycloak >= 9


zauberpony:
Hey, first of all thanks for your work!

Creating users with federated_identity set fails for keycloak > 9 with the following message:

```
Error: error sending POST request to /auth/admin/realms/my-realm/users/df4e64ba-5fbc-4842-b160-79b2f13d2e51/federated-identity/google: 409 Conflict

  on main.tf line 11, in resource "keycloak_user" "my_user":
  11: resource "keycloak_user" "my_user" {
```

Basically https://github.com/mrparkers/terraform-provider-keycloak/blob/master/keycloak/user.go#L43-L46 fails for keycloak >= 9, since https://github.com/mrparkers/terraform-provider-keycloak/blob/master/keycloak/user.go#L36 already creates the federated_identity, thus the conflict.

Terraform snippet that reproduces the issue:
```
resource "keycloak_realm" "my_realm" {
  realm = "my-realm"
}

resource "keycloak_oidc_google_identity_provider" "realm_identity_provider" {
  realm         = keycloak_realm.my_realm.realm
  client_id     = "random-id.apps.googleusercontent.com"
  client_secret = "random-secret"
}

resource "keycloak_user" "my_user" {
  realm_id = keycloak_realm.my_realm.id
  username = "my_user"

  federated_identity {
    identity_provider = "google"
    user_id           = 123456789
    user_name         = "my_user@example.com"
  }
}
```

tomrutsaert:
Thanks for the report, good find.
Temp workaround, first create the user, then add the federate identity....

Good that keycloak changed this behavior so drastically.... :confused: 
So before version 9, It ignored the federated identity when you did a POST to create a new user. You need to do an explicit call to add the federated identities.
Starting version 9 it does look at the federated identities and creates them. So you may not do the extra calls any more --> they return a 409 conflict.

Anyway I think, I will need to fall back on first deleting the federating identities after creating the user, to add them immediately again.
Otherwise we need to build in some check on keycloak version, to know how to create these federated identities in all cases.

zauberpony:
Well, instead of deleting the federating identity, you could check if they have been created/are available.
Or, remove the `federatedIdentities` from the `user` when doing the initial post.

Both solutions (and some others one could think of like e.g., just ignore 409 etc.) feel hacky.

[...]

PR #300, opened 2020-05-26
Fixes for tests that are failing on keycloak version >= 9

tomrutsaert:
- fix issue 291: #291 
- Disable customIdp test and example:
The custom idp example and test do not work in keycloak 10, because the interfaces that our customIdp implements, have changed in the keycloak latest version.
We need to make decide which keycloak version we going to support and test for the customIdp
- Fix for google-custom-idp not existing. this scenario was already covered by the above point so unneeded, the custom config is still checked

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,7 +25,7 @@ services:
     environment:
     - KEYCLOAK_USER=keycloak
     - KEYCLOAK_PASSWORD=password
-    - KEYCLOAK_LOGLEVEL=DEBUG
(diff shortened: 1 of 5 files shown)


hawknewton:
Ideally we'd support both and detect the server version at runtime.

tomrutsaert:
If you are talking about customidp, yes i agree.
This is about starting the discussion en in mean time helping 10.0.1 version pass in #294.

If you are talking about #291 fix, this fix does support both and does not need runtime version check.

PR #301, opened 2020-05-27
Fix incorrect default value in docs

pascal-hofmann:
keycloak_openid_user_realm_role_protocol_mapper.multivalued defaults to false (not true).


tomrutsaert:
It is indeed false in our code.
Which is a bit strange as the default in keycloak is true.

PR #306, opened 2020-05-28
Introduction of testUtil function keycloakVersionIsHigherOrEqualThan + plus fix tests

tomrutsaert:
test_utils function keycloakVersionIsHigherOrEqualThan is used to fix the number of events and micorprofilejwt scope test failures on older keycloaks
see #294


tomrutsaert:
@mrparkers @hawknewton
What do you guys think?
KEYCLOAK_VERSION ENV VAR is always available in our ci test sets, So I used it to make a little function that can be used in tests.
I think this PR will make everything green in #294

provider/resource_keycloak_openid_client_optional_scopes_test.go
```diff
@@ -429,6 +440,37 @@ resource "keycloak_openid_client_optional_scopes" "optional_scopes" {
...
+	client_id      = "${keycloak_openid_client.client.id}"
+	optional_scopes = [
+		"address",
+		"phone",
+		"offline_access",
+		"${keycloak_openid_client_scope.client_scope.name}"
+	]
+}
+	`, realm, client, clientScope)
+	}
```
tomrutsaert, line 473: I am aware we could work with smaller yaml string snippets and concat them. But I went for clarity and use the whole yml snippet as it would be used in that version of keycloak

provider/test_utils.go
```diff
@@ -75,3 +76,14 @@ func TestCheckResourceAttrNot(name, key, value string) resource.TestCheckFunc {
 		return nil
 	}
 }
+
+func keycloakVersionIsHigherOrEqualThan(keycloakMajorVersion int) (bool, error) {
```
mrparkers, line 80: nitpick: could you rename this to `keycloakVersionIsGreaterThanOrEqualTo`?

README.md
```diff
@@ -35,9 +35,12 @@ KEYCLOAK_CLIENT_SECRET=884e0f95-0f42-4a63-9b1f-94274655669e \
 KEYCLOAK_CLIENT_TIMEOUT=5 \
 KEYCLOAK_REALM=master \
 KEYCLOAK_URL="http://localhost:8080" \
+KEYCLOAK_VERSION=8.0.1
```
hawknewton, line 38: Why not pull the version from the `serverinfo` endpoint?

tomrutsaert: Yes that would be a good idea, but it seems that is not always possible.

You need the keycloakClient to be able to fetch the serverInfo
And AFAIK this can only be done through `keycloakClient := testAccProvider.Meta().(*keycloak.KeycloakClient)` 
This Meta code documentation states:
```
// Meta returns the metadata associated with this provider that was
// returned by the Configure call. It will be nil until Configure is called.
```
Thus in a test where you need differentiate in the input of the "Configure Call", you can not use this. See most of the optionalScopesTests, as example: `TestAccKeycloakOpenidClientOptionalScopes_basic`

To be complete I did add a method to gets the keycloak version via /serverInfo endpoint and I have used it in 'TestAccKeycloakRealmEvents_unsetEnabledEventTypes'

If I am missing something please let me know. This is also all new for me....
@mrparkers , Do you know a way to get a keycloakClient before 'Configure' is called

provider/test_utils.go
tomrutsaert: done

README.md
mrparkers: During the tests, `testAccProvider.Meta()` should always be a pointer to `keycloak.KeycloakClient`.  Unless I'm missing something, you should always have access to this.

This is configured in the `init` function of `provider_test.go`: https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/provider_test.go#L21 - so it should be available as soon as the package is loaded.

The `Configure` call that you're talking about is this: https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/provider.go#L140

Meaning that any time a provider is instantiated, it's already been "configured" with an interface, which is really just a pointer to a `keycloak.KeycloakClient`.  We should be able to use a type assertion to convert it from the interface value any time we want.

tomrutsaert: I always got following type of errors when I tried to access `testAccProvider.Meta()' before the `Config:` part of a test was finished.

```tom@tom-HP-EliteBook-850-G3:~/projects/go/src/github.com/tomrutsaert/terraform-provider-keycloak$ KEYCLOAK_CLIENT_ID=terraform KEYCLOAK_CLIENT_SECRET=884e0f95-0f42-4a63-9b1f-94274655669e KEYCLOAK_REALM=master KEYCLOAK_URL="http://localhost:8080" TF_ACC=1 go test github.com/mrparkers/terraform-provider-keycloak/provider -v --run TestAccKeycloakOpenidClientOptionalScopes_basic
=== RUN   TestAccKeycloakOpenidClientOptionalScopes_basic
--- FAIL: TestAccKeycloakOpenidClientOptionalScopes_basic (0.00s)
panic: interface conversion: interface {} is nil, not *keycloak.KeycloakClient [recovered]
        panic: interface conversion: interface {} is nil, not *keycloak.KeycloakClient

goroutine 7 [running]:
testing.tRunner.func1.1(0x123dd20, 0xc00011e2d0)
        /snap/go/current/src/testing/testing.go:940 +0x2f5
testing.tRunner.func1(0xc00065a120)
        /snap/go/current/src/testing/testing.go:943 +0x3f9
panic(0x123dd20, 0xc00011e2d0)
        /snap/go/current/src/runtime/panic.go:969 +0x166
github.com/mrparkers/terraform-provider-keycloak/provider.TestAccKeycloakOpenidClientOptionalScopes_basic(0xc00065a120)
        /home/tom/projects/go/src/github.com/tomrutsaert/terraform-provider-keycloak/provider/resource_keycloak_openid_client_optional_scopes_test.go:34 +0x583
testing.tRunner(0xc00065a120, 0x1448f18)
        /snap/go/current/src/testing/testing.go:991 +0xdc
created by testing.(*T).Run
        /snap/go/current/src/testing/testing.go:1042 +0x357
FAIL    github.com/mrparkers/terraform-provider-keycloak/provider       0.014s
FAIL
```

Probably doing something wrong,
Might take a look tomorrow....

mrparkers: Interesting.  I'd be happy to take a look if you update your PR to include the code that's causing the panic.

tomrutsaert: I have created 3 `TestAccKeycloakOpenidClientOptionalScopes_basic_fail` tests, that all three have the same panic. 
this not working anywhere where i call it.
Strange although it did work in the `TestAccKeycloakRealmEvents_unsetEnabledEventTypes` test

tomrutsaert: apparently the 'fail' tests do work in an CI env :confused:

mrparkers: Ahh, yeah I see now.

So it looks like what's happening here is that the provider [`Configure` function](https://github.com/hashicorp/terraform-plugin-sdk/blob/79b4af5fffc8a33bcff6183efc99f3006d7fde5c/helper/schema/provider.go#L245) isn't called until the middle of the test execution.  So those panics in tests 1 and 2 make sense - when you're running only a single test, the provider hasn't been configured yet at the point where you're attempting a type assertion against the meta interface.

It passes in CI because CI runs all of the tests sequentially, so by the time your tests are reached, the provider has already been configured.

Interestingly enough, tests 1 and 2 fail for me locally due to a panic, and test 3 fails, but because of an unexpected diff.  But it seems to pass in CI.

I went ahead and pushed a change that adds a new variable `keycloakClient` that can be used during tests when the meta interface isn't ready yet.

tomrutsaert: Thanks, 
I have removed the testUtil method that uses the KEYCLOAK_VERSION env var
Everything is now using serverInfo endpoint version as suggested by hawknewton

provider/test_utils.go
```diff
@@ -75,3 +77,29 @@ func TestCheckResourceAttrNot(name, key, value string) resource.TestCheckFunc {
...
+	if err != nil {
+		return false, fmt.Errorf("/serverInfo endpoint retuned an error, server Keycloak version could not be determined")
+	}
+	keycloakVersion := serverInfo.SystemInfo.ServerVersion
+	if len(keycloakVersion) > 0 {
+		i, err := strconv.Atoi(string(keycloakVersion[0]))
+		if err == nil {
+			return i >= keycloakMajorVersion, nil
+		}
+	}
```
hawknewton, line 96: This breaks when  ` version >= 10`.

What do you think of making use of hashicorp's [go-version](https://github.com/hashicorp/go-version) library?  It'd be nice to be able to target specific minor/patch versions as well.

tomrutsaert: Good point, and i used the go-version lib, thanks for the tip.
I am learning new things

PR #309, opened 2020-05-29
add user session note protocol mapper

dlechevalier:


dlechevalier:
Adding user session note protocol mapper.
Waiting for your feedback.

Issue #276, opened 2020-04-23
User-Agent uses the Go user agent instead of the Terraform user-agent


awilliamsOM1:
Terraform uses a user agent that looks like this:

`User-Agent: Terraform/0.12.6 TF_APPEND_USER_AGENT` where `TF_APPEND_USER_AGENT` is an optional environment variable.

The keycloak provider in terraform does not specify the user agent when making requests so the default GO user agent is used `User-Agent: Go-http-client/1.1`

This causes problems for because we cannot properly identify traffic as coming from terraform. The standard terraform user agent should be used possibly with other provider information added

PR #311, opened 2020-06-02
Use the terraform-plugin-sdk user agent string with running tf version

awilliamsOM1:
Http requests now using the user-agent built by the terraform-plugin-sdk using the running terraform version string instead of the default Go user agent.

Before: `Go-http-client/1.1`
After: `HashiCorp Terraform/0.12.25 (+https://www.terraform.io) Terraform Plugin SDK/1.6.0`

Starting with Terraform version v.0.12 the running version Terraform can be obtained from the provider. However this field is not included in prior versions of Terraform and the running version cannot be discovered. As a way to support older versions, the version string of "0.11+compatible" is used. This was inspired by the way the aws provider plugin works, https://github.com/terraform-providers/terraform-provider-aws/commit/0e183c1c5919a6cc1727b829e7a6bf1968f8b265

Fix for issue #276.

### DIFF
--- a/keycloak/keycloak_client.go
+++ b/keycloak/keycloak_client.go
@@ -17,7 +17,6 @@ import (
 	"time"
 
 	"golang.org/x/net/publicsuffix"
-	"github.com/hashicorp/terraform-plugin-sdk/httpclient"
 )
 
 type KeycloakClient struct {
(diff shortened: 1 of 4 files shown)


mrparkers:
thanks for this, this is a good change.  I made one small update to compute the user agent in the `provider` package, since I want to keep the `keycloak` package free of anything terraform related so it can be used outside of the context of this provider.

awilliamsOM1:
Thanks for making the change and fixing the tests. Moving the user agent building to prevent a `terraform` dependency in the `keycloak` package makes sense. Thanks!

PR #314, opened 2020-06-09
Fix number of parts in the OpenID client service account role import

jgrgt:
Small change, but it's keeping me from importing.


mrparkers:
Thank you for the PR!

I went ahead and backfilled the missing test for this, and found out that the import function was broken in more than one way, so it wouldn't have worked even with your fix.  I pushed the other needed fix to your branch.

jgrgt:
@mrparkers Indeed, the import does (did) not work. But adding a test and fixing the other bug was above my level. Thanks for looking into this and fixing it!

PR #315, opened 2020-06-09
Add resource keycloak_user_roles

elmarx:
Hey,

I know there are already two open PRs to add user-role-support, but they both look stale and not ready…

This one is done, it's based heavily upon the `keycloak_group_roles`, although in some details the implementation differs, since the group-endpoint returns mapped roles, but the user-endpoint not.

On the other hand, explicitly querying the user's role-mapping endpoints returns full roles (opposed to role-names), thus the implementation is simpler, once the role-mapping endpoint has been queried.

I also tested the resource manually, so as far as I know it's done/it works, but I'm looking forward to your feedback.

### DIFF
--- a/provider/resource_keycloak_user_roles.go
+++ b/provider/resource_keycloak_user_roles.go
@@ -1,0 +1,347 @@
+package provider
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/mrparkers/terraform-provider-keycloak/keycloak"
+	"strings"
+)
+
+func resourceKeycloakUserRoles() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceKeycloakUserRolesCreate,
+		Read:   resourceKeycloakUserRolesRead,
+		Update: resourceKeycloakUserRolesUpdate,
+		Delete: resourceKeycloakUserRolesDelete,
+		// This resource can be imported using {{realm}}/{{userId}}.
+		Importer: &schema.ResourceImporter{
+			State: resourceKeycloakUserRolesImport,
+		},
+		Schema: map[string]*schema.Schema{
+			"realm_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"user_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"role_ids": {
+				Type:     schema.TypeSet,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+				Required: true,
+			},
+		},
+	}
+}
+
+func userRolesId(realmId, userId string) string {
+	return fmt.Sprintf("%s/%s", realmId, userId)
+}
+
+func addRolesToUser(keycloakClient *keycloak.KeycloakClient, clientRolesToAdd map[string][]*keycloak.Role, realmRolesToAdd []*keycloak.Role, user *keycloak.User) error {
+	if len(realmRolesToAdd) != 0 {
+		err := keycloakClient.AddRealmRolesToUser(user.RealmId, user.Id, realmRolesToAdd)
+		if err != nil {
+			return err
+		}
+	}
+
+	for k, roles := range clientRolesToAdd {
+		if len(roles) != 0 {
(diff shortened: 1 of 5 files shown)


n0rig:
This looks great; this is something I was trying to solve. Can we get more eyes on this?

provider/keycloak_role_helpers.go
```diff
@@ -0,0 +1,47 @@
...
+	for _, roleId := range roleIds {
+		role, err := keycloakClient.GetRole(realmId, roleId)
+		if err != nil {
+			return nil, err
+		}
+
+		if role.ClientRole {
+			roles[role.ClientId] = append(roles[role.ClientId], role)
+		} else {
+			roles["realm"] = append(roles["realm"], role)
```
tomrutsaert, line 17: I see you use "realm" as a keyword in the map of roles. But what if there is a client called "realm"?
Is there a reason for not using the UserRoleMapping struct instead of this map of roles?

provider/resource_keycloak_user_roles.go
```diff
@@ -0,0 +1,253 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"user_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"role_ids": {
```
tomrutsaert, line 31: Do I understand correctly that "role_ids" contains both realm role ids and client role ids?

provider/keycloak_role_helpers.go
zauberpony: Correct, I moved this function to reuse it. And I reused this "pattern" with 'realm' as a special-keywork for clients without challenging it.
But looking at the code, using a dedicated struct might simplify the handling/code, and would avoid the edge-case with clients named "realm".

I will look into this and see if I can change this.

provider/resource_keycloak_user_roles.go
zauberpony: Yes. That's once again adopted from `keycloak_group_roles`.

provider/keycloak_role_helpers.go
mrparkers: Yeah feel free to blame me for this one.  Using a dedicated struct is probably better, but I won't force you to fix this since it was my decision to do this in the first place.

provider/resource_keycloak_user_roles.go
mrparkers: Yup, Keycloak doesn't actually care if its a realm or client role, so in this sense, the provider doesn't either.  So if you have multiple `keycloak_role` resources, you can add each `keycloak_role.*.id` to this set without having to worry about whether the roles are attached to realms or clients.

mrparkers:
I think this can be merged as-is, but I'll give you the opportunity to make the improvement that @tomrutsaert suggested if you're interested.  If not, we can merge this as-is.

Thanks for the PR!

zauberpony:
Thanks for your feedback!

Well I played around with the code, and yes, using a dedicated struct is much cleaner and should be simple. I will update this PR in the next few days :)

elmarx:
OK, PR has been updated. 

Short summary of changes:

The map is gone and I replaced it with a dedicated structure `roleMapping`, which is similar to Keycloak's `UserRoleMapping`, but not the same (the map for client-roles uses the uuid instead of the clientid which keycloak uses).
Thus keycloak `UserRoleMapping` needs to be transformed into the internal `roleMapping`.

This of course means the previous "diff-function" `removeDuplicationRoles` from `resource_keycloak_group_roles` can't be reused, so there's `calculateRoleMappingUpdates` to figure out the updates.

elmarx:
Awesome, thanks for merging and thanks for adding documentation ;)

ryanmac8:
Any idea when the documentation will be updated? Would like to try this out but don't know what variables the resource requires.

elmarx:
mrparkers already updated the documentation: https://github.com/mrparkers/terraform-provider-keycloak/blob/master/docs/resources/keycloak_user_roles.md

But please be aware that since the merge of this PR no version has been released, yet. So you need to compile the provider on your own.

ryanmac8:
Ah ok. Thanks! I was reading the documentation from https://mrparkers.github.io/terraform-provider-keycloak and couldn't find it. I did end up compelling and testing it out. Works perfectly! :D

PR #318, opened 2020-06-12
resource to enable the token exchange idp permission

tomrutsaert:
resource to enable the token exchange idp permission  and auto create of client policy

This is part of a preview keycloak feature. You need to enable this feature to be able to use this resource.
More information about enabling the preview feature can be found here: https://www.keycloak.org/docs/latest/securing_apps/index.html#_token-exchange

This implies that I enabled this feature in the docker-compose file.
This also implies that the tests will fail in CIRCLE-CI

@mrparkers  What is your opinion on this?
Should I enable this preview feature in CIRCLE_CI or do I add a check in test_utils based on a env variable?
I did not see a way to detect if this feature is enabled via server_info, thus we can not use this server_info approach.


tomrutsaert:
Strange all the tests passed. I did not expect this....

docs/resources/keycloak_identity_provider_token_exchange_scope_permission.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Argument Reference
+
+The following arguments are supported:
+
+- `realm_id` - (Required) The realm this group exists in.
+- `provider_alias` - (Required) Alias of the identity provider.
+- `policy_type` - (Optional) Defaults to "client" This is also the only value policy type supported by this provider.
+- `clients` - (Required) Ids of the clients for which a policy will be created and set on scope based token exchange permission.
+- `policy_id` - (Computed) Policy id that will be set on the scope based token exchange permission automatically created by enabling permissions on the reference identity provider.
```
mrparkers, line 67: typically we document computed attributes under a section called "Attributes Reference". example: https://github.com/mrparkers/terraform-provider-keycloak/blob/master/docs/resources/keycloak_group.md#attributes-reference

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
```diff
@@ -0,0 +1,292 @@
...
+			"clients": {
+				Type:        schema.TypeSet,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Required:    true,
+				Description: "Ids of the clients for which a policy will be created and set on scope based token exchange permission",
+			},
+			"policy_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				ForceNew:    true,
```
mrparkers, line 48: why is `ForceNew` needed for a computed attribute?

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
```diff
@@ -0,0 +1,292 @@
...
+	} else if len(permission.Policies) == 1 {
+		openidClientAuthorizationClientPolicy, err := keycloakClient.GetOpenidClientAuthorizationClientPolicy(realmId, realmManagementClient.Id, permission.Policies[0])
+		if err != nil {
+			return err
+		}
+		openidClientAuthorizationClientPolicy.Clients = clients
+		return keycloakClient.UpdateOpenidClientAuthorizationClientPolicy(openidClientAuthorizationClientPolicy)
+
+	} else {
+		return fmt.Errorf("only supports 1 client policy is supporterd on the identity provider token exchange scope based permission, but %d were found", len(permission.Policies))
```
mrparkers, line 108:
```suggestion
		return fmt.Errorf("only one client policy is supported, but %d were found", len(permission.Policies))
```

cleaned up the wording a bit. the error message will already contain the resource name, so I think this is enough info for the user.

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
```diff
@@ -0,0 +1,292 @@
...
+		ResourceServerId: realmManagementClientId,
+		Name:             realmId + "_" + providerAlias + "_client_policy",
+		DecisionStrategy: "UNANIMOUS",
+		Logic:            "POSITIVE",
+		Type:             "client",
+		Clients:          clients,
+	}
+	err := keycloakClient.NewOpenidClientAuthorizationClientPolicy(openidClientAuthorizationClientPolicy)
+	if err != nil {
+		if strings.Contains(err.Error(), "409 Conflict") {
```
mrparkers, line 124: let's make a helper function to detect 409 errors, similar to the one we use for 404 errors: https://github.com/mrparkers/terraform-provider-keycloak/blob/766d7b6ded815cfe5db986a053af2bdc5b78c344/keycloak/error.go#L17-L21

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
```diff
@@ -0,0 +1,292 @@
...
+			index := 1
+			for index <= 10 && err != nil {
```
mrparkers, line 126: can you help me understand why we're trying to do this 10 times?  it looks like we're avoiding an error due to the policy name already being in use - can we let users decide what name they want to use instead?

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+func resourceKeycloakIdentityProviderTokenExchangeScopePermissionUpdate(data *schema.ResourceData, meta interface{}) error {
+	keycloakClient := meta.(*keycloak.KeycloakClient)
+	realmId := data.Get("realm_id").(string)
+	providerAlias := data.Get("provider_alias").(string)
+	policyType := data.Get("policy_type").(string)
+	var clients []string
+
+	if v, ok := data.GetOk("clients"); ok {
```
mrparkers, line 187: the `clients` attribute is required, so I don't think we need to use `data.GetOk` here

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
```diff
@@ -0,0 +1,292 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"provider_alias": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"policy_type": {
```
mrparkers, line 33: you could add some validation on this attribute if only `client` is supported, similar to https://github.com/mrparkers/terraform-provider-keycloak/blob/5d0c60bac0759dbdf26e1aee6693346b4b598fc9/provider/resource_keycloak_openid_client.go#L54

docs/resources/keycloak_identity_provider_token_exchange_scope_permission.md
tomrutsaert: I will change this

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
tomrutsaert: Mistake, I will remove it

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
tomrutsaert: :+1:  I will take look on Monday for this

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
tomrutsaert: well, I did not know how to solve this. But I a user manually disables the Idp permissions,
The policy stays behind. so you would never be able to recreate the policy. So even If you let the user choose the policy name. the CreateAfterManualDestroyTest would fail.
Furthermore If an user can choose their own policyName there might be confusion that they have to create a policy outside of this resource and link to it.....
I know, it still not bullet proof

mrparkers:
Nice work on this @tomrutsaert!  I think it's pretty funny that the feature flag only toggles its visibility in the UI.  We should probably just leave that toggle enabled for example purposes.

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
tomrutsaert: true, But the idea Is that this later extended for different type of policies, at that moment this will be come optional attribute. It is only required when policy_type 'client' is selected

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
tomrutsaert: Yes I will take a look

provider/resource_keycloak_identity_provider_token_exchange_scope_permission.go
tomrutsaert: I am gonna remove the index and 10 times loop and just add random string in case the generated name random already existed
What do you think?

mrparkers: Yeah this is probably fine.  My understanding is that this can only happen when users manually make changes outside of Terraform, so I am not too worried about it.

tomrutsaert: Yes exactly

PR #320, opened 2020-06-15
Improve client scope

wadahiro:
I have made the following improvements regarding client scope.

- Added `include_in_token_scope` and `gui_order` arguments into `keycloak_openid_client_scope` resource
- Added `keycloak_saml_client_scope` resource
- Added `keycloak_saml_client_default_scopes` resource


provider/resource_keycloak_openid_client_scope.go
```diff
@@ -35,6 +37,15 @@ func resourceKeycloakOpenidClientScope() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"include_in_token_scope": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"gui_order": {
+				Type:     schema.TypeString,
```
mrparkers, line 46: Any reason for not using `schema.TypeInt` for this attribute?

keycloak/openid_client_scope.go
```diff
@@ -13,6 +13,8 @@ type OpenidClientScope struct {
 	Attributes  struct {
 		DisplayOnConsentScreen string `json:"display.on.consent.screen"` // boolean in string form
 		ConsentScreenText      string `json:"consent.screen.text"`
+		GuiOrder               string `json:"gui.order"`
+		IncludeInTokenScope    string `json:"include.in.token.scope"` // boolean in string form
```
mrparkers, line 17: You can set this attribute's type to `KeycloakBoolQuoted`, which is just a bool that understands how to marshal and unmarshal as a string.  Then you won't have to worry about converting it back and forth between those two types.

keycloak/saml_client_scope.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+type SamlClientScope struct {
+	Id          string `json:"id,omitempty"`
+	RealmId     string `json:"-"`
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Protocol    string `json:"protocol"`
+	Attributes  struct {
+		DisplayOnConsentScreen string `json:"display.on.consent.screen"` // boolean in string form
```
mrparkers, line 14: Same comment as above, let's use `KeycloakBoolQuoted` here.

provider/resource_keycloak_openid_client_scope.go
wadahiro: Keycloak REST API treats it as string type. Also, to remove the attribute, empty string ("") is used in admin console. That's why I use `schema.TypeString` here.

keycloak/openid_client_scope.go
wadahiro: OK, I'll use `KeycloakBoolQuoted`.

keycloak/saml_client_scope.go
wadahiro: I see.

provider/resource_keycloak_openid_client_scope.go
mrparkers: I understand why you made that decision, but I think it would be better to change the type of this attribute to a `schema.TypeInt` anyways.  Unfortunately, there is a lot of areas for improvement for the Keycloak REST API, so I don't want to exactly model provider resources based on the types that the API uses.

We can convert the integer from the provider to a string to make the API call, and we can treat 0 as an empty string for the purpose of omitting the attribute to reset the order.

wadahiro: All right. I'll change the code.

wadahiro:
@mrparkers I modified the code and rebased onto the latest master. Could you check it again?

PR #321, opened 2020-06-15
Added docs for service account role.

Useurmind:
Hi,

the docs for the service account roles of openid clients was missing. I only discovered that it existed when browsing through some old issues. 

I added the documentation so that other users can more easily find it.

Kind Regards


Useurmind:
I dont know why the tests fail. I just added a new markdown file. That should not influence the test success. :(

tomrutsaert:
I reran the tests, looks good.
Thx for the contribution.

PR #332, opened 2020-06-30
Improve provider connection: allow ommited "/auth" and improve login-error-messages

elmarx:
Hey,

Strictly speaking this PR addresses two distinct issues, but both are rather small.

First, the provider previously did not check the status-code of the initial login, so if some reverse-proxy returns an error and a non-JSON-error-message, the error-message is just about a parsing failure. By explicitly checking the HTTP-status-code, the provider can show more precise error messages.

The second commit enables to use keycloak at another web-context then "/auth" (e.g. to leave out "/auth" completely). 
Maybe it would be nicer to just use the setting `baseUrl` and require users to include "/auth" if necessary, but that would introduce a breaking change, so I opted to add an additional optional setting to override the "/auth" part.

### DIFF
--- a/keycloak/keycloak_client.go
+++ b/keycloak/keycloak_client.go
@@ -40,11 +40,11 @@ type ClientCredentials struct {
 }
 
 const (
-	apiUrl   = "/auth/admin"
-	tokenUrl = "%s/auth/realms/%s/protocol/openid-connect/token"
+	apiUrl   = "/admin"
+	tokenUrl = "%s/realms/%s/protocol/openid-connect/token"
 )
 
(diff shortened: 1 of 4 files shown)


keycloak/keycloak_client.go
```diff
@@ -122,6 +122,9 @@ func (keycloakClient *KeycloakClient) login() error {
 	if err != nil {
 		return err
 	}
+	if accessTokenResponse.StatusCode != 200 {
```
mrparkers, line 125: nitpick: let's use `http.StatusOk` here

```suggestion
	if accessTokenResponse.StatusCode != http.StatusOk {
```

elmarx: Good point, I updated the PR.

PR #340, opened 2020-07-08
Add store_token and trust_email to google identitiy provider

sj14:


mrparkers:
Hey @sj14, thanks for the PR.

I'm a bit confused about the problem you're trying to solve with this change.  These attributes can already be found within the [generic identity provider schema](https://github.com/mrparkers/terraform-provider-keycloak/blob/d8b44cd1fd15c23f95b6b29ff47ac4f9398d23ca/provider/generic_keycloak_identity_provider.go#L19), which is [merged](https://github.com/mrparkers/terraform-provider-keycloak/blob/d8b44cd1fd15c23f95b6b29ff47ac4f9398d23ca/provider/resource_keycloak_oidc_google_identity_provider.go#L84) into the schema you're modifying here.

Could you help me understand a bit more about the problem you're trying to solve?

sj14:
Ah, then this PR makes zero sense. Sorry, I haven't seen that it's embedded. Quite funny because I haven't tried the same matching keywords without the patch.

PR #350, opened 2020-07-22
[Draft] Add role attributes

sj14:
This PR adds role attributes.

### DIFF
--- a/keycloak/role.go
+++ b/keycloak/role.go
@@ -7,15 +7,16 @@ import (
 )
 
 type Role struct {
-	Id          string `json:"id,omitempty"`
-	RealmId     string `json:"-"`
-	ClientId    string `json:"-"`
-	RoleId      string `json:"-"`
-	Name        string `json:"name"`
-	Description string `json:"description"`
-	ClientRole  bool   `json:"clientRole"`
-	ContainerId string `json:"containerId"`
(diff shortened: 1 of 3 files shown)


sj14:
@mrparkers I'm not sure what I'm missing in the test. When you think this PR is a good idea in general, could you help me with the test please?

tomrutsaert:
Hey sj14,

I would advice not adapt the existinmg test but rather create new tests besides the existing tests. Because changing the existing tests might give wrong perception when they fail. We want the basic tests plus a test for a specific feature.
You can copy paste a basic existing test and apdat it to test your specific feature.

Specific for your case , you could take a look on how it is done on the realm resource and custom attributes test TestAccKeycloakRealm_customAttribute (line 625)

I assume the reason this test fails is because you put the key also between double quotes. It should not be between double quotes, only the value should.
it should look like this:
```
attributes   = {
	%s = "%s"
}
```

Did you run these test locally? The project readme explains how this can be done.

sj14:
Hey,
the change is probably not needed anymore for my company, hence I will probably not invest more time in it :-(
Feel free to close this PR when no one want's to continue on it.

mrparkers:
I'll go ahead and close this for now, feel free to reopen it if you'd like to work on this again.

PR #352, opened 2020-07-23
add docs for 'config' parameter in resource 'keycloak_custom_user_federation'

RomanNess:
It took me some time to realize that `keycloak_custom_user_federation` can already hand through provider configuration to the user federation provider. So I decided to add some docs in this pull request. 

I also extended the `Example Usage` to indicate that values in the config can contain various types such as strings and bools.


mrparkers:

PR #356, opened 2020-08-03
Add WebAuth{, Passwordwordless} Policy

klausenbusk:
Fix #355

PR #357, opened 2020-08-03
Add support for 11.0.0 and drop support for 8.0.0

klausenbusk:
#356 is currently blocked on this PR.

### DIFF
--- a/keycloak/ldap_group_mapper.go
+++ b/keycloak/ldap_group_mapper.go
@@ -27,6 +27,8 @@ type LdapGroupMapper struct {
 	MappedGroupAttributes       []string
 
 	DropNonExistingGroupsDuringSync bool
+
+	LdapGroupsPath string
 }
 
 func convertFromLdapGroupMapperToComponent(ldapGroupMapper *LdapGroupMapper) *component {
@@ -66,6 +68,9 @@ func convertFromLdapGroupMapperToComponent(ldapGroupMapper *LdapGroupMapper) *co
 		},
 		"drop.non.existing.groups.during.sync": {
 			strconv.FormatBool(ldapGroupMapper.DropNonExistingGroupsDuringSync),
+		},
+		"groups.path": {
+			ldapGroupMapper.LdapGroupsPath,
 		},
 	}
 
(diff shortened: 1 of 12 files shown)


klausenbusk:
I assume the error is caused by this change: https://github.com/keycloak/keycloak/commit/6005503a3dd7b65b79e8c3289738cb5d60779d8f
```
Error: error sending POST request to /auth/admin/realms/terraform-tduhynsuat/components: 400 Bad Request. Response body: {"errorMessage":"Missing configuration for Groups Path"}
```

klausenbusk:
... and now I broke all  <11.0.0.. How can we support 11.0.0 without breaking < 11.0.0?

mrparkers:
Yeah... I was afraid that something like this might happen due to a major release introducing breaking API changes.

I haven't dug too deep in this yet, but it looks like Keycloak 11 is expecting `groups.path` for the LDAP group mapper to always be provided, and all previous versions are failing when that is provided.

If this is the case, then we'll have to start doing version checking at runtime instead of just doing it during tests.  I can dig into this more when I have some free time.

klausenbusk:
>  I can dig into this more when I have some free time.

I would appreciate that! I just experimented a little in the "wip" commit, but it still lacks a `if version < 11.0.0 set ldap_groups_path = default`, and I'm not sure it is implemented the best way. Maybe you have a better idea/design in mind :)

klausenbusk:
@mrparkers Could you take a look? I just added `Computed: true` which is a much simpler solution.

klausenbusk:
> Hmm... Now 11.0.0 is broke..

Now all the CI is green. Could you take a look at the last commit?

keycloak/ldap_group_mapper.go
```diff
@@ -69,9 +69,10 @@ func convertFromLdapGroupMapperToComponent(ldapGroupMapper *LdapGroupMapper) *co
 		"drop.non.existing.groups.during.sync": {
 			strconv.FormatBool(ldapGroupMapper.DropNonExistingGroupsDuringSync),
 		},
-		"groups.path": {
-			ldapGroupMapper.LdapGroupsPath,
-		},
+	}
+
+	if KeycloakVersionIsGreaterThanOrEqualTo(keycloakClient, "11.0.0") {
```
klausenbusk, line 74: I don't think we need this, but maybe it worth it anyway as it is more explicit.

svenstaro:
@klausenbusk This needs a rebase.

klausenbusk:
I think this was <somehow> fixed as part of: https://github.com/mrparkers/terraform-provider-keycloak/pull/369

PR #360, opened 2020-08-07
Data source for user and authentication execution

rjmasikome:
Hi everyone,

I didn't see any contribution guidelines, so I will just go ahead and open this PR. This Pull Request will add data source for User and Authentication Execution. Currently we need that for our process.

If there's anything lacking please do let me know.

Cheers,
RJ


provider/data_source_keycloak_user.go
```diff
@@ -0,0 +1,38 @@
...
+
+	realmId := data.Get("realm_id").(string)
+	username := data.Get("username").(string)
+
+	user, err := keycloakClient.GetUserByUsername(realmId, username)
+	if err != nil {
+		return err
+	}
+
+	mapFromUserToData(data, user)
```
mrparkers, line 35: Once master is merged into this branch, this line will cause test failures, since this function will attempt to set attributes that aren't defined in the data source.

mrparkers (changes requested):
Apologies for the late review on this.

The only potential issue I'm noticing here is that the schemas for these new data sources don't match the schemas for their equivalent resources.  Typically we'd expect these schemas to match unless there is a good reason not to.

Could you update the data source schemas, or help me understand if there was particular reason to omit the other attributes?

provider/data_source_keycloak_user.go
rjmasikome: Changed as requested

mrparkers (approved):
I went ahead and pushed some changes to your branch.  let me know what you think.  otherwise, lgtm

rjmasikome:
Hey thanks for the change. LGTM.

PR #364, opened 2020-08-19
Add Keycloak_openid_client_permission for admin_fine_grained_authz

jermarchand:
Add on `keycloak_openid_client` the optional option `permissions_enabled` to enable Fine grained permissions on the client.

Add resource `keycloak_openid_client_permission` to manage the policies to apply to each scopes.

### DIFF
--- a/provider/resource_keycloak_openid_client.go
+++ b/provider/resource_keycloak_openid_client.go
@@ -3,10 +3,11 @@ package provider
 import (
 	"errors"
 	"fmt"
+	"strings"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	"github.com/mrparkers/terraform-provider-keycloak/keycloak"
-	"strings"
 )
 
 var (
--- a/docs/resources/openid_client_permission.md
+++ b/docs/resources/openid_client_permission.md
@@ -1,0 +1,93 @@
+# keycloak_openid_client_permissions
+
+Allows you to manage all openid client Scope Based Permissions.
+
+This is part of a preview keycloak feature. You need to enable this feature to be able to use this resource.
+More information about enabling the preview feature can be found here: https://www.keycloak.org/docs/latest/securing_apps/index.html#_token-exchange
+
+When enabling Openid Client Permissions, Keycloak does several things automatically:
+1. Enable Authorization on build-in realm-management client
+1. Create scopes "view", "manage", "configure", "map-roles", "map-roles-client-scope", "map-roles-composite", "token-exchange"
+1. Create a resource representing the openid client
+1. Create all scope based permission for the scopes and openid client resource
+
+If the realm-management Authorization is not enable, you have to ceate a dependency (`depends_on`) with the policy and the openid client.
+
+### Example Usage
+
+```hcl
+resource "keycloak_realm" "realm" {
+	realm  = "realm"
+}
+
+resource "keycloak_openid_client" "my_openid_client" {
+  realm_id              = keycloak_realm.realm.id
+  name                  = "my_openid_client"
+  client_id             = "my_openid_client"
+  client_secret         = "secret"
+  access_type           = "CONFIDENTIAL"
+  standard_flow_enabled = true
+  valid_redirect_uris   = [
+    "http://localhost:8080/*",
+  ]
+}
+
+data "keycloak_openid_client" "realm_management" {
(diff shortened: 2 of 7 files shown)


mrparkers:
@tomrutsaert could you take a look at this when you have a chance?  I'm very unfamiliar with the permissions / policy features within Keycloak.

provider/resource_keycloak_openid_client_permission.go
```diff
@@ -0,0 +1,300 @@
...
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mrparkers/terraform-provider-keycloak/keycloak"
+)
+
+func resourceKeycloakOpenidClientPermission() *schema.Resource {
```
tomrutsaert, line 12: Should this not plural resourceKeycloakOpenidClientPermissions instead of resourceKeycloakOpenidClientPermission?

provider/resource_keycloak_openid_client.go
```diff
@@ -180,6 +181,11 @@ func resourceKeycloakOpenidClient() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"permissions_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
```
tomrutsaert, line 188: Would Having resourceKeycloakOpenidClientPermissions pointing a client not make this obsolete?
I would let the resourceKeycloakOpenidClientPermissions handle the enable of disable of the permissions on the client.
Even in keycloak I think the permissions are located in the wrong location. It is really a resource in it own right
What do you think?

provider/resource_keycloak_openid_client.go
```diff
@@ -377,6 +383,20 @@ func resourceKeycloakOpenidClientCreate(data *schema.ResourceData, meta interfac
...
+			if err != nil {
+				return err
+			}
+		} else {
+			err := keycloakClient.DisableOpenidClientPermissions(client.RealmId, client.Id)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
tomrutsaert, line 398: If you agree this would be a part of resourceKeycloakOpenidClientPermissions create method

tomrutsaert (commented):
Looks good.
Better than what I created: resourceKeycloakIdentityProviderTokenExchangeScopePermission
By looking at this, mine does to many things automatically.
The policy should indeed be created in a separate resource....
I do not recall why I did not it like this....

I would call the resource in plural: keycloak_openid_client_permission**s** as that is what it is called in the keycloak UI + it handles more then 1 permission

That being said: I would remove the permission toggle on the keycloak openid client resource and let the fact you have a keycloak_openid_client_permission resource handle the enable/disable call. Or with an enable flag on the resource itself.
@jermarchand what do you think?

provider/resource_keycloak_openid_client_permission_test.go
```diff
@@ -0,0 +1,135 @@
...
+	realmName := "tf_view-" + acctest.RandString(10)
+	clientId := "tf-" + acctest.RandString(10)
+
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: testAccProviderFactories,
+		PreCheck:          func() { testAccPreCheck(t) },
+		Steps: []resource.TestStep{
+			{
+				Config: testKeycloakOpenidClientPermission_basic(realmName, clientId),
+				Check:  testAccCheckKeycloakOpenidClientPermissionExists("keycloak_openid_client_permission.my_permission"),
```
tomrutsaert, line 23: Could you add a test that checks that the policy is actually set on the permission?

provider/resource_keycloak_openid_client_permission.go
jermarchand: Yep, I'll rename this resource.

provider/resource_keycloak_openid_client.go
jermarchand: I added `permissions_enabled` in` resourceKeycloakOpenidClient` to be able to enable / disable permissions independent of the permissions configuration themselves.

We can indeed consider that the presence of `resourceKeycloakOpenidClientPermissions` activates the permissions on the client.
To be able to manage its deactivation, I will keep the `enabled` parameter in` resourceKeycloakOpenidClientPermissions` which with the value `false` will delete the resource

jermarchand:
@tomrutsaert could you take a look at my updates ?

tomrutsaert:
Looks good to me

PR #365, opened 2020-08-20
Add field session_note to user session protocol mapper. This field ne…

hcl31415:
Add field session_note to user session protocol mapper. This field needs to be written additionally to session_note_label if the protocol mapper is intended to work correctly, e.g. an Apache mod_auth_openidc client.

### DIFF
--- a/docs-old/resources/keycloak_openid_user_session_note_protocol_mapper.md
+++ b/docs-old/resources/keycloak_openid_user_session_note_protocol_mapper.md
@@ -31,6 +31,7 @@ resource "keycloak_openid_user_session_note_protocol_mapper" "user_session_note_
 	client_id          = keycloak_openid_client.openid_client.id
 	claim_name         = "foo"
 	claim_value_type   = "String"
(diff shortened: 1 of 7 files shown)


mrparkers:
Huh, I swear this attribute in the protocol mapper used to be called "session note label", but it appears to have disappeared completely from the UI.  I think it might have been silently renamed in one of the releases since I originally created this resource.

What do you think about simply renaming the existing `session_note_label` attribute to `session_note`?  If you're fine with this, I can push the change to your branch, or you could take a stab at it.

hcl31415:
That's perfectly fine with me. I just kept the original session note label field because I didn't want to introduce unintended behaviour in Keycloak. It seems though that the value of the session note label is still stored in the database, when PUT or POST against the admin API but in the end I couldn't find out what it really does.

hcl31415:
I removed the session-note-label field from the commit.

mrparkers (approved):
LGTM :+1:

PR #368, opened 2020-08-21
Fix delete endpoint group policy

jermarchand:


mrparkers:
hey @jermarchand, thanks for the PR!

I'm not too familiar with these particular features within Keycloak, so could you provide me with a simple example in HCL that demonstrates the problem that this PR fixes?

I'll write a test that uses this HCL so we don't have to worry about something like this happening again.

jermarchand:
I can't create tests because the DELETE endpoint seem's to not return an error :/

However, with regard to the other Endpoints called in this file, I think we should use "group" and not "roles".

mrparkers:
That's okay, the API itself doesn't necessarily need to return an error, there are plenty of ways that we can assert that something happened during the execution of this code.  We could check that the Terraform state has a specific value, or if further API calls to Keycloak can verify something we did.

In any case, I really don't know enough about this specific Keycloak feature to know whether or not this is a good change.  Perhaps if you could describe to me what differences you see in the GUI after making this change, I may be able to create a test for that.

mrparkers:
I ended up adding this text and fixing this in #667.  Turns out it actually doesn't matter whether the endpoint uses `group` or `role` here, it will work in either case.  But I made the change anyways for consistency.

Thanks again for this PR!

PR #372, opened 2020-08-27
Fix group import

jermarchand:
After import a Group, the`ParentId`is not set in the tfstate. So on 'terraform plan` it show a delete+creation.
The `groupParentId(group *Group)` function always set the `parentId` with empty string (https://github.com/mrparkers/terraform-provider-keycloak/blob/master/keycloak/group.go#L25)

I propose to keep the `findParentGroup` recursion function, but instead of just return a boolean, return the `parentId` found.

### DIFF
--- a/keycloak/group.go
+++ b/keycloak/group.go
@@ -35,27 +35,27 @@ func (keycloakClient *KeycloakClient) groupParentId(group *Group) (string, error
 	}
 
 	var parentGroup Group
-	if findParentGroup(*group, groups, parentGroup) {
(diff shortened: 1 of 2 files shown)


mrparkers:
I experimented with this a bit myself - you're right that it doesn't look like the group parent ID is set on import, and this code change fixes that.  However, the tests aren't actually checking this (they pass without your fix).  I tried for a bit to get a failing test to verify this but I was not successful, so I'll just approve and merge this anyways.

Thanks for the PR!

PR #380, opened 2020-09-04
Added keycloak_openid_client_service_account_realm_role to docs-old/

lathspell:
The object keycloak_openid_client_service_account_realm_role exists but was not documented.
I found documentation in docs/ but not in docs-old/ which still seems to be the source for the Web documentation.

I couldn't test the result as I don't know how to process this mkdocs.yml


mrparkers:
I'm going to close this because I'm not too worried about keeping the contents of `docs-old` up to date.  I'll be keeping the GitHub pages site up for a little while longer, but v1 of the provider won't be maintained anymore going forward.

PR #385, opened 2020-09-12
Register required action if not reqistered

### LINKED ISSUE #354: Support registering a new required action
https://github.com/mrparkers/terraform-provider-keycloak/pull/131 implemented so we can update existing required action, but we can't add new required actions (ex: WebAuthn).

Registering a new required action is done by posting to [`/{realm}/authentication/register-required-action`](https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_registerrequiredaction), ex:
```sh
$ curl 'http://localhost:8080/auth/admin/realms/master/authentication/register-required-action' -H 'Content-Type: application/json' -H 'Authorization: Bearer eyJhbGc...' --data-raw '{"providerId":"webauthn-register","name":"Webauthn Register"}'
```

Unregistered required actions can be GET from [`unregistered-required-actions`](https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_getunregisteredrequiredactions).

klausenbusk:
Fix #354

### DIFF
--- a/example/main.tf
+++ b/example/main.tf
@@ -107,6 +107,13 @@ resource "keycloak_required_action" "custom-configured_totp" {
   priority       = keycloak_required_action.custom-terms-and-conditions.priority + 15
 }
 
+resource "keycloak_required_action" "required_action" {
+  realm_id  = keycloak_realm.test.realm
+  alias     = "webauthn-register"
+  enabled   = true
+  name      = "Webauthn Register"
+}
+
 resource "keycloak_group" "foo" {
   realm_id = keycloak_realm.test.id
   name     = "foo"
--- a/keycloak/required_action.go
+++ b/keycloak/required_action.go
@@ -7,6 +7,7 @@ type RequiredAction struct {
 	RealmId       string              `json:"-"`
 	Alias         string              `json:"alias"`
 	Name          string              `json:"name"`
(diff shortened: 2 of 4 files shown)


klausenbusk:
I'm not sure the failure is related to the changes?

klausenbusk:
> I'm not sure the failures is related to the changes?

I just rebease with the circleci test timeout from master (https://github.com/mrparkers/terraform-provider-keycloak/commit/000b2485152725eab5ba598bc69947db036d84d3) and now CI is green.

mrparkers:
this looks good, just a couple of thoughts:

- could you add the example action you're using to one of the existing test cases to verify that this will work going forward
- in terms of the lifecycle of a newly registered required action, do you think it should be "unregistered" on delete?  just want to hear your thoughts on this.

klausenbusk:
> could you add the example action you're using to one of the existing test cases to verify that this will work going forward

Sure, will do.

> in terms of the lifecycle of a newly registered required action, do you think it should be "unregistered" on delete? just want to hear your thoughts on this.

You can't "unregister" a required action, you can only delete it (at which point you would need to register it again).

mrparkers (approved):
thanks for adding that test, LGTM 👍

PR #389, opened 2020-09-25
OpenId and SAML Client's attributes

slavko-vega:
Supported are attributes for OpenId and SAML clients


slavko-vega:
@mrparkers please review this PR, it represents solution for https://github.com/mrparkers/terraform-provider-keycloak/issues/387 . Basically, to keep existing strongly typed attributes and use their benefits, I implemented a custom (un)marshaling for those types.

mrparkers:
Hey @slavko-vega, sorry for taking so long to review this.

I understand the approach for doing this within the `IdentityProviderMapperConfig`, but I'm less confident in the approach for the OpenID and SAML clients.  Those approaches seem to define reserved keys to avoid overwriting, but otherwise is similar to the first approach.  That seems brittle to me, because now we'd have to make sure that we come back to this code every time a new first class attribute is added.

Do you think we can still solve this problem for OpenID and SAML clients without defining this list of reserved keys?  Even if it could potentially cause errors for an end user, I think that's fine as long as there's some sort of warning about this attribute to use it at your own risk.

slavko-vega:
Hey @mrparkers I was thinking about the problem and have a solution based on reflection. Namely, we can easily read tags of all predefined properties. What do you think about solution based on the reflection?

mrparkers:
I guess I'm not completely understanding the problem we're trying to solve here.  I understand that your goal is to add user-defined attributes to API calls for resources such as openid and saml clients, but is there any reason that we need to use reflection or define reserved keys?

My thinking is that we could just allow users to use this `attributes` map to define any properties they want on the model that is sent to the Keycloak API, and it's up to them to make sure that they aren't overwriting other attributes or sending attributes that the Keycloak API will reject.  As long as we know what field we're trying to add, we know what field to look for in the response we get back so the state can be updated accordingly.

This appears to be what we're doing with the `IdentityProviderMapperConfig`, right?  Can we take the same approach for openid and saml clients as well?

slavko-vega:
Hey @mrparkers , you're totally right. I wasn't be aware of that implementation in IdentityProviderMapperConfig. I'll use the same approach for OpenId/SAML client.

slavko-vega:
Hi @mrparkers, client (OpenId and SAML) attributes are implemented as extra_config.

civ0:
Hi all, are there any updates on this PR?

Setting custom attributes on clients would be very useful for me. Please let me know, if there is something I can do to help progressing this PR.

Tokynet:
BUMP @slavko-vega and @mrparkers

PR #391, opened 2020-09-28
Added issuer and sync_mode attributes to r/keycloak_oidc_identity_provider

spirius:
Added `issuer` and `sync_mode` attributes for `keycloak_oidc_identity_provider` resource.

### DIFF
--- a/provider/resource_keycloak_oidc_identity_provider.go
+++ b/provider/resource_keycloak_oidc_identity_provider.go
@@ -62,6 +62,17 @@ func resourceKeycloakOidcIdentityProvider() *schema.Resource {
 			Required:    true,
 			Description: "Token URL.",
 		},
+		"issuer": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "Issuer Identifier",
+		},
+		"sync_mode": {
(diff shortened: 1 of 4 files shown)


provider/resource_keycloak_oidc_identity_provider.go
```diff
@@ -62,6 +62,18 @@ func resourceKeycloakOidcIdentityProvider() *schema.Resource {
 			Required:    true,
 			Description: "Token URL.",
 		},
+		"issuer": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "Issuer Identifier",
+			Default:     "",
```
mrparkers, line 69: nitpick: This will always default to the zero-value for strings, which is an empty string.  So this line can be removed.

mrparkers (commented):
LGTM besides the one comment, thanks for the PR!

provider/resource_keycloak_oidc_identity_provider.go
spirius: thanks for the review! fixed as you suggested.

mrparkers:
Closing in favor of #746

PR #392, opened 2020-10-01
Add the decision strategy setting to the openid client

randomswdev:
This merge request is intended to allow overriding the decision strategy at the client level. If not implemented, the client always defines "Unanimuous" as the Global decision strategy, which can prevent some use caes to be imeplemented.


mrparkers (approved):
LGTM, thanks for the PR!

PR #397, opened 2020-10-12
Allow option IMPORT in LDAP group mapper

hcl31415:
There exists a third option for the LDAP group mapper sync mode: IMPORT. This commit adds this value to the enum.


mrparkers (approved):
Thanks for the PR!

PR #400, opened 2020-10-21
Add Keycloak_users_permission for admin_fine_grained_authz

jermarchand:
Add resource keycloak_users_permission to manage the policies to apply to each scopes.

### DIFF
--- a/docs/resources/users_permission.md
+++ b/docs/resources/users_permission.md
@@ -1,0 +1,133 @@
+---
+page_title: "keycloak_users_permissions Resource"
+---
+
+# keycloak_users_permissions
+
+Allows you to manage all users Scope Based Permissions https://www.keycloak.org/docs/latest/server_admin/#_users-permissions
+
+This is part of a preview keycloak feature `admin_fine_grained_authz` (see https://www.keycloak.org/docs/latest/server_admin/#_fine_grain_permissions)
+
+You need to enable this feature to be able to use this resource.
+More information about enabling the preview feature `admin_fine_grained_authz` can be found here: https://www.keycloak.org/docs/latest/server_installation/#profiles
+
+When enabling Users Permissions, Keycloak does several things automatically:
+1. Enable Authorization on build-in realm-management client (if not already enabled)
+1. Create a resource representing the users permissions
+1. Create scopes "view", "manage", "map-roles", "manage-group-membership", "impersonate", "user-impersonated"
+1. Create all scope based permission for the scopes and users resource
+
+If the realm-management Authorization is not enable, you have to ceate a dependency (`depends_on`) with the policy and the openid client.
+
+### Example Usage
+
+```hcl
+resource "keycloak_realm" "realm" {
+	realm  = "realm"
+}
+
+data "keycloak_openid_client" "realm_management" {
+	realm_id  = keycloak_realm.realm.id
+	client_id = "realm-management"  
+}
+
+resource keycloak_openid_client_permissions "realm-management_permission" {
+	realm_id   = keycloak_realm.realm.id
+	client_id  = data.keycloak_openid_client.realm_management.id
+	enabled = true
+}
+
+resource keycloak_user test {
+	realm_id = keycloak_realm.realm.id
+	username = "test-user"
+
+	email      = "test-user@fakedomain.com"
+	first_name = "Testy"
+	last_name  = "Tester"
+}
+
+resource keycloak_openid_client_user_policy test {
+	resource_server_id = "${data.keycloak_openid_client.realm_management.id}"
+	realm_id = keycloak_realm.realm.id
+	name = "client_user_policy_test"
+	users = [keycloak_user.test.id]
+	logic = "POSITIVE"
+	decision_strategy = "UNANIMOUS"
+	depends_on = [
+		keycloak_openid_client_permissions.realm-management_permission,
+	]
+}
+
+resource "keycloak_users_permissions" "my_permission" {
+	realm_id                                = keycloak_realm.realm.id
+	
+	view_scope {
+		policies = [ keycloak_openid_client_user_policy.test.id ]
+		description = "description"
+		decision_strategy = "UNANIMOUS"
+	}
+	manage_scope {
+		policies = [ keycloak_openid_client_user_policy.test.id ]
+		description = "description"
+		decision_strategy = "UNANIMOUS"
+	}
+	map_roles_scope {
+		policies = [ keycloak_openid_client_user_policy.test.id ]
+		description = "description"
+		decision_strategy = "UNANIMOUS"
+	}
+	manage_group_membership_scope {
+		policies = [ keycloak_openid_client_user_policy.test.id ]
+		description = "description"
+		decision_strategy = "UNANIMOUS"
+	}
+	impersonate_scope {
+		policies = [ keycloak_openid_client_user_policy.test.id ]
+		description = "description"
+		decision_strategy = "UNANIMOUS"
+	}
+	user_impersonated_scope {
+		policies = [ keycloak_openid_client_user_policy.test.id ]
+		description = "description"
+		decision_strategy = "UNANIMOUS"
+	}
+}
+```
+
+### Argument Reference
+
+The following arguments are supported:
+
+- `realm_id` - (Required) The realm this users permissions exists in.
+- `view_scope` - (Optional) When specified, set the scope based view permission.
+  - `policies` - (Optional) Assigned policies to the permission 
+  - `description` - (Optional) Description of the permission 
+  - `decision_strategy` - (Optional) Decision strategie of the permission 
+- `manage_scope` - (Optional) When specified, set the scope based manage permission.
+  - `policies` - (Optional) Assigned policies to the permission 
+  - `description` - (Optional) Description of the permission 
+  - `decision_strategy` - (Optional) Decision strategie of the permission 
+- `map_roles_scope` - (Optional) When specified, set the scope based map_roles permission.
+  - `policies` - (Optional) Assigned policies to the permission 
+  - `description` - (Optional) Description of the permission 
+  - `decision_strategy` - (Optional) Decision strategie of the permission 
+- `manage_group_memberip_scope` - (Optional) When specified, set the scope based manage_group_memberip permission.
+  - `policies` - (Optional) Assigned policies to the permission 
+  - `description` - (Optional) Description of the permission 
+  - `decision_strategy` - (Optional) Decision strategie of the permission 
+- `impersonate_scope` - (Optional) When specified, set the scope based impersonate permission.
+  - `policies` - (Optional) Assigned policies to the permission 
+  - `description` - (Optional) Description of the permission 
+  - `decision_strategy` - (Optional) Decision strategie of the permission 
+- `user_impersonated_scope` - (Optional) When specified, set the scope based user_impersonated permission.
+  - `policies` - (Optional) Assigned policies to the permission 
+  - `description` - (Optional) Description of the permission 
+  - `decision_strategy` - (Optional) Decision strategie of the permission 
+
+### Attributes Reference
+
+In addition to the arguments listed above, the following computed attributes are exported:
+
+- `enabled` - User permissions are Enabled (true) 
+- `authorization_resource_server_id` - Resource server id representing the realm management client on which this permission is managed.
+
--- a/docs/resources/users_permission.md
+++ b/docs/resources/users_permission.md
@@ -4,93 +4,113 @@ page_title: "keycloak_users_permissions Resource"
 
 # keycloak_users_permissions
 
-Allows you to manage all users Scope Based Permissions https://www.keycloak.org/docs/latest/server_admin/#_users-permissions
+Allows you to manage fine-grained permissions for all users in a realm: https://www.keycloak.org/docs/latest/server_admin/#_users-permissions
 
-This is part of a preview keycloak feature `admin_fine_grained_authz` (see https://www.keycloak.org/docs/latest/server_admin/#_fine_grain_permissions)
+This is part of a preview Keycloak feature: `admin_fine_grained_authz` (see https://www.keycloak.org/docs/latest/server_admin/#_fine_grain_permissions).
+This feature can be enabled with the Keycloak option `-Dkeycloak.profile.feature.admin_fine_grained_authz=enabled`. See the
+example [`docker-compose.yml`](https://github.com/mrparkers/terraform-provider-keycloak/blob/898094df6b3e01c3404981ce7ca268142d6ff0e5/docker-compose.yml#L21) file for an example.
 
-You need to enable this feature to be able to use this resource.
-More information about enabling the preview feature `admin_fine_grained_authz` can be found here: https://www.keycloak.org/docs/latest/server_installation/#profiles
+When enabling fine-grained permissions for users, Keycloak does several things automatically:
+1. Enable Authorization on built-in `realm-management` client (if not already enabled).
+1. Create a resource representing the users permissions.
+1. Create scopes `view`, `manage`, `map-roles`, `manage-group-membership`, `impersonate`, and `user-impersonated`.
+1. Create all scope based permission for the scopes and users resources.
 
-When enabling Users Permissions, Keycloak does several things automatically:
-1. Enable Authorization on build-in realm-management client (if not already enabled)
-1. Create a resource representing the users permissions
-1. Create scopes "view", "manage", "map-roles", "manage-group-membership", "impersonate", "user-impersonated"
-1. Create all scope based permission for the scopes and users resource
+~> This resource should only be created once per realm.
 
-If the realm-management Authorization is not enable, you have to ceate a dependency (`depends_on`) with the policy and the openid client.
-
-### Example Usage
+## Example Usage
 
 ```hcl
 resource "keycloak_realm" "realm" {
-	realm  = "realm"
+  realm  = "my-realm"
 }
 
 data "keycloak_openid_client" "realm_management" {
-	realm_id  = keycloak_realm.realm.id
-	client_id = "realm-management"  
+  realm_id  = keycloak_realm.realm.id
+  client_id = "realm-management"
 }
 
-resource keycloak_openid_client_permissions "realm-management_permission" {
-	realm_id   = keycloak_realm.realm.id
-	client_id  = data.keycloak_openid_client.realm_management.id
-	enabled = true
+// enable permissions for realm-management client
+resource "keycloak_openid_client_permissions" "realm_management_permission" {
+  realm_id   = keycloak_realm.realm.id
+  client_id  = data.keycloak_openid_client.realm_management.id
+  enabled    = true
 }
 
-resource keycloak_user test {
-	realm_id = keycloak_realm.realm.id
-	username = "test-user"
+// creating a user to use with the keycloak_openid_client_user_policy resource
+resource "keycloak_user" "test" {
+  realm_id = keycloak_realm.realm.id
+  username = "test-user"
 
-	email      = "test-user@fakedomain.com"
-	first_name = "Testy"
-	last_name  = "Tester"
+  email      = "test-user@fakedomain.com"
+  first_name = "Testy"
+  last_name  = "Tester"
 }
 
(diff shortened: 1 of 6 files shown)


docs/resources/users_permission.md
```diff
@@ -0,0 +1,91 @@
...
+	user_impersonated_scope_policy_id       = keycloak_openid_client_user_policy.test.id
+}
+```
+
+### Argument Reference
+
+The following arguments are supported:
+
+- `realm_id` - (Required) The realm this group exists in.
+- `client_id` - (Required) The id of the client that provides the role.
```
tomrutsaert, line 78: This can be removed?

docs/resources/users_permission.md
```diff
@@ -0,0 +1,91 @@
...
+	logic = "POSITIVE"
+	decision_strategy = "UNANIMOUS"
+	depends_on = [
+		keycloak_openid_client.my_openid_client
+	]
+}
+
+resource "keycloak_users_permissions" "my_permission" {
+	realm_id                                = keycloak_realm.realm.id
+	client_id                               = keycloak_openid_client.my_openid_client.id
```
tomrutsaert, line 63: This can be removed?

docs/resources/users_permission.md
```diff
@@ -0,0 +1,91 @@
...
+If the realm-management Authorization is not enable, you have to ceate a dependency (`depends_on`) with the policy and the openid client.
+
+### Example Usage
+
+```hcl
+resource "keycloak_realm" "realm" {
+	realm  = "realm"
+}
+
+resource "keycloak_openid_client" "my_openid_client" {
```
tomrutsaert, line 23: Is this client relevant in this documentation example?

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/mrparkers/terraform-provider-keycloak/keycloak"
+)
+
+func resourceKeycloakUsersPermissions() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceKeycloakUsersPermissionsCreate,
+		Read:   resourceKeycloakUsersPermissionsRead,
+		Delete: resourceKeycloakUsersPermissionsDelete,
+		Update: resourceKeycloakUsersPermissionsUpdate,
+		// This resource can be imported using {{realm}}/{{client_id}}. The Client ID is displayed in the GUI
```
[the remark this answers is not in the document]
jermarchand: Indeed, the comment does not describe the variable. 
I remove it because the import can be done with the realmId

provider/resource_keycloak_users_permissions_test.go
```diff
@@ -0,0 +1,211 @@
...
+resource "keycloak_realm" "realm" {
+  realm = "%s"
+}
+
+data keycloak_openid_client "realm_management" {
+  realm_id  = keycloak_realm.realm.id
+  client_id = "realm-management"  
+}
+
+resource keycloak_openid_client_permissions "realm-management_permission" {
```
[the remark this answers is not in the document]
jermarchand: As long as permission management is not enabled somewhere in the realm, the AuthZ function is not available on "realm-management" client. So it is not possible to add policies.

As the policy must exist before you can add it to user permissions, I therefore have 3 steps:

1: activate permissions on an existing element in the realm (here on the "realm-management" client)
2: create the policy after the AuthZ is available on "realm-management"
3: create the usere_permission with the policies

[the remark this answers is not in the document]
tomrutsaert: And now I remember why I had build in the policy creation within the idp_permission resource.
Any way your way is still better, As long as it is documented and stated why it is needed.
Somewhere in the future we should adapt the resource ipd_permission resource to be consistent with your implementations.

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,278 @@
...
+func resourceKeycloakUsersPermissionsImport(d *schema.ResourceData, _ interface{}) ([]*schema.ResourceData, error) {
+	parts := strings.Split(d.Id(), "/")
+	if len(parts) != 2 {
+		return nil, fmt.Errorf("Invalid import. Supported import formats: {{realmId}}")
+	}
+	d.Set("realm_id", parts[0])
+
+	d.SetId(parts[0])
+
+	return []*schema.ResourceData{d}, nil
```
tomrutsaert, line 277: You indicated that only realm ID is needed for import
so if there is only a realmId? Is it correct that the incoming Id needs to be split in 2 parts, or that there ever be 2 parts?
So I assume this will always return an error?

jermarchand: Doyou think I can use "ImportStatePassthrough" ?
https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_importer.go#L47

tomrutsaert: To be honest, I always just copy paste this from other resources. i always forget how this exactly works
lets ask @mrparkers ....

jermarchand: honstly, I do the same most of time ;) 

I ask for `ImportStatePassthrough` because it's used for realm import (https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/resource_keycloak_realm.go#L91)

mrparkers: `ImportStatePassthrough` is used when all you need to do during import is set the ID and nothing more.  The `keycloak_realm` resource uses this because it doesn't need to set any other attributes for the resource besides its ID.

As an example, look at the import function for the `keycloak_openid_client` resource: https://github.com/mrparkers/terraform-provider-keycloak/blob/ba25fb0d2e25f163784610aa28dd646386b8a322/provider/resource_keycloak_openid_client.go#L438-L447

This is not a good candidate for `ImportStatePassthrough` because the `keycloak_openid_client` resource needs both the ID and `realm_id` attributes in order to function correctly.

Does that make sense?

jermarchand: I add an import test :) 

With `schema.ImportStatePassthrough` it does'nt work.
With a dedicated function, it works.

mrparkers: Right, `ImportStatePassthrough` won't work because you're setting both the resource ID and the `realm_id` attribute.  You can only use `ImportStatePassthrough` to set the ID of the resource.

docs/resources/users_permission.md
```diff
@@ -0,0 +1,85 @@
+# keycloak_users_permissions
```
mrparkers, line 1: We'll need to add a page title to this doc page before this line.  Take a look at the other files within this folder for an example.

docs/resources/users_permission.md
```diff
@@ -0,0 +1,85 @@
...
+
+Allows you to manage all users Scope Based Permissions.
+
+This is part of a preview keycloak feature. You need to enable this feature to be able to use this resource.
+More information about enabling the preview feature can be found here: https://www.keycloak.org/docs/latest/securing_apps/index.html#_token-exchange
```
mrparkers, line 6: Is this the correct link for enabling this preview?

docs/resources/users_permission.md
```diff
@@ -0,0 +1,85 @@
...
+}
+
+resource keycloak_openid_client_permissions "realm-management_permission" {
+	realm_id   = keycloak_realm.realm.id
+	client_id  = data.keycloak_openid_client.realm_management.id
+	enabled = true
+}
+
+resource keycloak_user test {
+	realm_id = "${keycloak_realm.realm.id}"
```
mrparkers, line 35: Can you update this example to remove the deprecated interpolation syntax?  In this case, we should change `"${keycloak_realm.realm.id}"` to `keycloak_realm.realm.id`

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,272 @@
...
+		Schema: map[string]*schema.Schema{
+			"realm_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"enabled": {
+				Type:     schema.TypeBool,
+				Required: true,
+				ForceNew: true,
```
mrparkers, line 28: Did you mean to mark this as `ForceNew`?  This means that an update to the `enabled` attribute will actually cause the resource to be deleted and re-created.

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,272 @@
...
+		err := keycloakClient.EnableUsersPermissions(realmId)
+		if err != nil {
+			return err
+		}
+	} else {
+		err := keycloakClient.DisableUsersPermissions(realmId)
+		if err != nil {
+			return err
+		}
+	}
```
mrparkers, line 128: If you intended to mark the `enabled` attribute as `ForceNew`, then I don't think this block of code would ever actually run.  Update won't be called if this attribute is changed.

Also, we can use `data.HasChange` to detect if this was actually changed, so we don't have to enable or disable here on every single update if the user didn't actually change this attribute.

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,272 @@
...
+
+	viewScopePolicyId, ok := data.GetOkExists("view_scope_policy_id")
```
mrparkers, line 130: Let's replace all of the instances of `data.GetOkExists` with `data.GetOk`, since `data.GetOkExists` is deprecated.

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,272 @@
...
+	if ok && viewScopePolicyId != nil {
```
mrparkers, line 131: `viewScopePolicyId` is probably never going to be nil here.  I think what you want to do is use type assertion to cast this to a string, then check if the string is empty.

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,272 @@
...
+		err := setUsersScopePermissionPolicy(keycloakClient, realmId, "view", viewScopePolicyId.(string))
+		if err != nil {
+			return err
+		}
+	}
+	manageScopePolicyId, ok := data.GetOkExists("manage_scope_policy_id")
+	if ok && manageScopePolicyId != "" {
```
mrparkers, line 138: `manageScopePolicyId` should be cast to a string first before comparing it to an empty string.

This same comment applies to lines 145, 152, 159, and 166.

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,272 @@
...
+func resourceKeycloakUsersPermissionsRead(data *schema.ResourceData, meta interface{}) error {
+	keycloakClient := meta.(*keycloak.KeycloakClient)
+	realmId := data.Get("realm_id").(string)
+
+	usersPermissions, err := keycloakClient.GetUsersPermissions(realmId)
+	if err != nil {
+		return handleNotFoundError(err, data)
+	}
+
+	data.SetId(usersPermissions.RealmId)
```
mrparkers, line 185: The fact that the realm ID is used as this resource's ID implies that this resource can only be set once per realm, is that accurate?

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,272 @@
...
+	data.Set("realm_id", usersPermissions.RealmId)
+
+	data.Set("enabled", usersPermissions.Enabled)
+
+	if !usersPermissions.Enabled {
+		log.Printf("[WARN] Removing resource with id %s from state as it no longer enabled", data.Id())
```
mrparkers, line 191: I'm not sure I understand this, why do we want to remove this resource from state if its disabled?  Also, we aren't actually removing from state here.  You'll want to use `data.SetId("")` to do that.

provider/resource_keycloak_users_permissions.go
```diff
@@ -0,0 +1,272 @@
...
+		return nil
+	}
+
+	data.Set("view_scope_policy_id", nil)
+	data.Set("manage_scope_policy_id", nil)
+	data.Set("map_roles_scope_policy_id", nil)
+	data.Set("manage_group_membership_scope_policy_id", nil)
+	data.Set("impersonate_scope_policy_id", nil)
+	data.Set("user_impersonated_scope_policy_id", nil)
```
mrparkers, line 200: I don't think any of these `data.Set` calls are needed.

mrparkers (changes requested):
I realize that a lot of this code was borrowed from #364, so anything that has to change here will probably have to change there as well.  That's my fault - I should have been more on top of reviewing that PR, so it's okay that these potential changes weren't caught until now.

docs/resources/users_permission.md
jermarchand: nope ;)

provider/resource_keycloak_users_permissions.go
jermarchand: Yes, the User permissions endpoint is "/realms/%s/users-management-permissions"

provider/resource_keycloak_users_permissions.go
jermarchand: if I understand how it should be:
I update from "ForceNew + Required" to "Computed" then user permissions are enabled when "keycloak_users_permissions" is declared and deleted if "keycloak_users_permissions" is not present.

jermarchand:
@mrparkers When this PR is OK for you and merged, I will correct PR #409 and make a new PR to correct PR #364

PS: if you think that my PR (#409 and #364) are compatible with the Hacktoberfest conditions, I will be happy that they are tagged "hacktoberfest-accepted" ;)

provider/resource_keycloak_users_permissions.go
mrparkers: Looking at this again - you actually can use `ImportStatePassthough` here if you want.  Since the only information we need about this resource is the realm ID, which we use as the unique ID for the resource.

This will only work if you handle that edge case within the `resourceKeycloakUsersPermissionsRead` function though.  Importing this resource when using `ImportStatePassthrough` will cause `data.Get("realm_id")` to return an empty string, so you'll have to handle that within the read function to set that attribute using the resources ID.

Or, you can leave this how it is.  Your call!  :smile:

docs/resources/users_permission.md
```diff
@@ -0,0 +1,89 @@
+---
+page_title: "keycloak_users_permissions Resource"
+---
+
+# keycloak_users_permissions
+
+Allows you to manage all users Scope Based Permissions.
+
+This is part of a preview keycloak feature. You need to enable this feature to be able to use this resource.
+More information about enabling the preview feature can be found here: https://www.keycloak.org/docs/latest/server_installation/#profiles
```
mrparkers, line 10: Which of the features listed here is actually needed?  I assume `admin_fine_grained_authz`?

docs/resources/users_permission.md
```diff
@@ -0,0 +1,89 @@
...
+
+When enabling Users Permissions, Keycloak does several things automatically:
+1. Enable Authorization on build-in realm-management client
+1. Create scopes "view", "manage", "map-roles", "manage-group-membership", "impersonate", "user-impersonated"
+1. Create a resource representing the users permissions
+1. Create all scope based permission for the scopes and openid client resource
```
mrparkers, line 16: Is this list accurate as well?  I should have asked this earlier.

docs/resources/users_permission.md
```diff
@@ -0,0 +1,89 @@
...
+
+If the realm-management Authorization is not enable, you have to ceate a dependency (`depends_on`) with the policy and the openid client.
```
mrparkers, line 18: Is this sentence also correct?  I noticed it was borrowed from the other docs as well.

Sorry, I am not very familiar with this feature of Keycloak, so you'll have to help me out here 😅

mrparkers:
Hey @jermarchand, once those last comments are cleared up, I can get this merged, and you can work on the changes for #409 and #364 if you'd like.  If not, I can try to get to those when I have time.

docs/resources/users_permission.md
jermarchand: On a new realm, the "Autorization" tab is not available on the `realm-management` client. Unlike confidential clients (`account`, `broker`), it's not possible to set "Authorization Enabled" to ON.

Go to the tab `Permissions` of Users page and set it to "Enabled". You should see users permissions Scopes.
Return to the `realm-management` client, the "Autorization" tab is available.
In the sub-tabs of "Autorization" : 
- `Authorization Scopes`: the 6 scopes of user permissions are created
- `Resources` : the resource "Users" is created
- `Permissions` : A permission are created by resource and scope combination

docs/resources/users_permission.md
jermarchand: replaced by "1. Create all scope based permission for the scopes and users resource"

docs/resources/users_permission.md
mrparkers: This is really helpful, thanks for the explanation!  I'm starting to understand the scope of this PR much better now.

So it looks like this resource, in general, allows you to add a single policy to each of these permissions.

image

That red box is the only thing we can modify for each of these permissions, correct?  If so, perhaps we'd want to slightly extend this resource to allow the flexibility to also define descriptions and decision strategies some time in the future.  Can I get your thoughts on this?

provider/resource_keycloak_users_permissions.go
mrparkers: Yeah, I think it's okay for the existence of this resource to imply that this feature is enabled.  That makes sense to me.

Although, the `enabled` attribute is kinda useless now.  It's always going to be `true`, right?  The only case where it would be `false` is if the resource was deleted, in which case, you wouldn't have access to the attribute anyways.

It might be best to just remove this attribute entirely.

docs/resources/users_permission.md
jermarchand: Initially my need was just to add a single Policy per Scope. I started from what existed for the `keycloak_identity_provider_token_exchange_scope_permission` resource and with my level of understanding of Go, and Terraform providers, I proposed the client_permission PR and now this one.

Now with your remarks I understand much better what I code and how to do it better :) Thanks 

I have to look at how to use the resource `keycloak_openid_client_authorization_permission` which already exists and is exactly what I should have used from the start :)

mrparkers: Sure thing, happy to help!

I will say that I didn't have much involvement with the `keycloak_openid_client_authorization_permission` resource either, so don't be too scared to propose a new / better way of doing things just because that resource doesn't do it.

I think what you have so far is pretty good, but it could be extended a bit further by using sub schemas to allow users to define some of those other properties for a particular policy.  Perhaps something like this:

```hcl
resource "keycloak_users_permissions" {
  realm_id = "master"
  
  view_scope {
    policy_id   = "some policy id" // this is the one attribute you currently support
    description = "foo"            // this is another attribute that could potentially be supported in the future
  }
}
```

So instead of `view_scope_policy_id`, we can use an attribute like `view_scope`, which is a `schema.TypeSet`, to define the policy ID and other attributes.  Or maybe instead of `policy_id = "a string"`, we could do `policies = ["multiple", "policies"]` so that more than one could be supported.

These are just suggestions - perhaps there are better ways to define these attributes.  What do you think about this?

provider/resource_keycloak_users_permissions_test.go
```diff
@@ -0,0 +1,253 @@
...
+				realmManagementId = client.Id
+				break
+			}
+		}
+
+		if authorizationResourceServerId != realmManagementId {
+			return fmt.Errorf("computed authorizationResourceServerId %s was not equal to %s (the id of the realm-management client)", authorizationResourceServerId, realmManagementId)
+		}
+
+		viewScopePolicyId := rs.Primary.Attributes["view_scope.0.policies.0"]
```
jermarchand, line 63: @mrparkers Is there a better way to get the values ?

mrparkers: Not that I know of 😞

mrparkers:
Hey @jermarchand - I went ahead and pushed some changes to this PR.  If you're good with these changes, I can get this merged for you.  This LGTM as is.

Thanks again for your work on this!

jermarchand:
@mrparkers go merge :) 

I will try to repot the changes in the other PRs :)

mrparkers (approved):

PR #436, opened 2020-11-09
feat: add groups path to LDAP group mapper

### LINKED ISSUE #435: [ldap] - missing groups path in group-ldap-mapper
Hi there,

it seems that the `group path` property is missing in the ldap group mapper.

Will make a PR to fix that.

Screen Shot 2020-11-09 at 12 46 46

Cheers,

AdrienFromToulouse:
Closes #435

### DIFF
--- a/docs-old/resources/keycloak_ldap_group_mapper.md
+++ b/docs-old/resources/keycloak_ldap_group_mapper.md
@@ -70,6 +70,7 @@ The following arguments are supported:
 - `memberof_ldap_attribute` - (Optional) Specifies the name of the LDAP attribute on the LDAP user that contains the groups the user is a member of. Defaults to `memberOf`.
 - `mapped_group_attributes` - (Optional) Array of strings representing attributes on the LDAP group which will be mapped to attributes on the Keycloak group.
 - `drop_non_existing_groups_during_sync` - (Optional) When `true`, groups that no longer exist within LDAP will be dropped in Keycloak during sync. Defaults to `false`.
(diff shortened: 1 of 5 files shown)


AdrienFromToulouse:
Hi, @mrparkers sorry to ask you this but I think I don't quite understand how tests are supposed to validate the terraform state. I cannot see what needs to be modified in order to include the `groups_path` property.
If you have some hints to give me I would be happy to proceed.

Cheers!

AdrienFromToulouse:
👋 @mrparkers your guidance would be appreciated. Please let me know what I need to change when you have a bit of spare time.

Thanks again for your time maintaining this provider.

Cheers,

mrparkers:
Hey @AdrienFromToulouse, sorry for the really late review here.

I took a look at this, and the tests for version 9 and 10 appear to be failing because the ability to set the group path for LDAP group mappers was added in version 11.  So I opened #444 to add some functionality to the Keycloak client so that we can check the version we're using during plan / apply time, which will allow us to decide to optionally not set this attribute when using an older version.

Once I merge #444, I can take a look at why the tests are failing for version 11.

mrparkers:
I went ahead and made a change to mark this attribute as computed instead of specifying a default.  This seemed to fix the test failures we were running into.  I also didn't realize that this attribute expects a group with that path to already exist, so I added a test for this.

Let me know what you think and we can get this merged.

AdrienFromToulouse:
👋 you are awesome @mrparkers I did not pay attention to the fact that the group had to be pre-existing... my bad.

Many thanks for your help on this 🚀 👍

mrparkers (approved):
no problem, thanks for the PR!

PR #438, opened 2020-11-18
Validation of RegistrationEmailAsUsername attribute

spirius:
Background:
I want to have a realm, where `username` and `email` attributes have same value, registration is disabled and user is allowed to manage own email address. To achieve this configuration keycloak can be configured as follows

```
RegistrationAllowed=false
RegistrationEmailAsUsername=true
EditUsernameAllowed=false
```

As a result, keycloak's account management page will allow user to manage own email address (but not username), and username will be updated automatically by keycloak upon change.

This configuration is of course not very intuitive, because `RegistrationEmailAsUsername` attribute _should_ make sense only when registration is enabled, but in fact keycloak behaves just as described above. In order to achieve this from GUI, `RegistrationAllowed` can be temporary enabled / disabled, which will allow setting `RegistrationEmailAsUsername` to true. 

Current implementation of the provider does not allow setting `RegistrationEmailAsUsername` to true when `RegistrationAllowed` is false, resulting in `validation error: RegistrationEmailAsUsername cannot be true if RegistrationAllowed is false`. This change removes that condition.

### DIFF
--- a/keycloak/realm.go
+++ b/keycloak/realm.go
@@ -199,10 +199,6 @@ func (keycloakClient *KeycloakClient) DeleteRealm(name string) error {
 }
 
 func (keycloakClient *KeycloakClient) ValidateRealm(realm *Realm) error {
-	if realm.RegistrationAllowed == false && realm.RegistrationEmailAsUsername == true {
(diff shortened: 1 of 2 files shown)


mrparkers (approved):
makes sense to me, thanks for the PR!

PR #455, opened 2021-01-09
Allow commas in keycloak_custom_user_federation config

### LINKED ISSUE #454: [keycloak_custom_user_federation] config not allowing commas
If the `config` from `keycloak_custom_user_federation` contains a comma, the plugin will ignore everything after the comma from that config.

Example:
```
resource "keycloak_custom_user_federation" "custom_user_federation" {
  name        = "User adapter provider"
  realm_id    = keycloak_realm.realm.id
  provider_id = "user-adapter-provider"

  enabled     = true

  config = {
    configWithComma = "this works,this will be ignored"
  }
}
```

Debug (note `"configWithComma":["this works"]`):
```
2021-01-08T21:29:35.904+0200 [DEBUG] plugin.terraform-provider-keycloak_v2.0.0: 
{
   "name":"User adapter provider",
   "providerId":"user-adapter-provider",
   "providerType":"org.keycloak.storage.UserStorageProvider",
   "parentId":"passwordless",
   "config":{
      "cachePolicy":[
         "DEFAULT"
      ],
      "configWithComma":[
         "this works"
      ],
      "enabled":[
         "true"
      ],
      "priority":[
         "0"
      ]
   }
}
```

This is most likely caused by splitting the config from comma [here](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/resource_keycloak_custom_user_federation.go#L85).

@tomrutsaert what was the reason for splitting by `,` ?

Thanks for this awesome provider!

PaulGgithub:
Allow commas in keycloak_custom_user_federation config. Fixes #454 

User federation config is a MultivaluedMap. A multivalued map has a key and a list of values.

The correct terraform config would be:
```
config = {
key = ["value1", "value2"]
}
```

Changing the config structure to array of strings would break existing configurations therefore only the first value is set and taken into account.

### DIFF
--- a/provider/resource_keycloak_custom_user_federation.go
+++ b/provider/resource_keycloak_custom_user_federation.go
@@ -82,7 +82,7 @@ func getCustomUserFederationFromData(data *schema.ResourceData) *keycloak.Custom
 	config := map[string][]string{}
 	if v, ok := data.GetOk("config"); ok {
 		for key, value := range v.(map[string]interface{}) {
-			config[key] = strings.Split(value.(string), ",")
+			config[key] = []string{value.(string)}
 		}
 	}
 	parentId := ""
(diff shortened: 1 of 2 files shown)


mrparkers (approved):
LGTM, thanks for the PR!

PR #464, opened 2021-01-13
Add support for configuration of client scopes on realm level

### LINKED ISSUE #463: Add support for configuration of client scopes on realm level
Keycloak allows to configure default client-scope mappings on realm level that are used as default scope configuration for clients.

Keycloak supports two variants:
- default default client scopes: client-scopes that are assigned to a client by default 
- default optional client scopes: client-scopes that can be optionally requested by a client

The respective attributes in the [RealmRepresentation](https://github.com/keycloak/keycloak/blob/32319c727b78ef820117798b856643d6baab8c90/core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java#L148) are:
- `List<String> defaultDefaultClientScopes`
- `List<String> defaultOptionalClientScopes`

Both attributes are supported since Keycloak 6.x
The default client-scopes are applied during realm creation via [RealmManager.realmImport](https://github.com/keycloak/keycloak/blob/32319c727b78ef820117798b856643d6baab8c90/services/src/main/java/org/keycloak/services/managers/RealmManager.java#L561):

thomasdarimont:
Fixes #463

### DIFF
--- a/docs/resources/realm.md
+++ b/docs/resources/realm.md
@@ -222,6 +222,11 @@ Each of these attributes are blocks with the following attributes:
 
 - `internal_id` - (Computed) When importing realms created outside of this terraform provider, they could use generated arbitrary IDs for the internal realm id. Realms created by this provider always use the realm's name for its internal id.
 
+## Default Client Scopes
+
+- `default_default_client_scopes` - (Optional) A list of default default client scopes to be used for client definitions. Defaults to `[]` or keycloak's built-in default default client-scopes.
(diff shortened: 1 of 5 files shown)


mrparkers (approved):
Looks great, thanks for the PR!

thomasdarimont:
@mrparkers it turns out that this feature is incomplete... /o\ 
We also need to make the [`clientScopes`](https://github.com/keycloak/keycloak/blob/master/core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java#L147) to have full control over the mapping.

I'll send another PR ASAP.

PR #475, opened 2021-01-28
Add attributes management to keycloak_role

dbolack:
This adds the attributes field to keycloak roles.

A couple of small tweaks popped in due to my linter and I slightly reworked the after the PUT update  in the role update function to return a call to read instead of processing  and updating directly.

This is a pretty simple addition, most of it was cribbed from how realm attributes were handled with minor changes to reflect realm.attributes and role.attributes not being the same type in the JSON.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+## v2.2.0 (January 23, 2021)
+
+FEATURES:
+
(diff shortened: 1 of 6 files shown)


mrparkers (approved):
thanks for the PR!  I made some changes because these attributes appear to be similar to attributes for groups.

PR #481, opened 2021-02-05
Allow identity_provider_mapper to be overriden for keycloak_user_template_importer_identity_provider_mapper

Photonios:
Screenshot 2021-02-05 at 13 26 59

Currently, `identity_provider_mapper` is computed as `[provider id]-username-idp-mapper`. This isn't always the case. Take for example the Facebook identity provider. The correct `identity_provider_mapper` is `oidc-username-idp-mapper`.

This PR adds an optional property to override the `identity_provider_mapper` attribute.

For completeness, this is the request the Keycloak admin console makes:

image

### DIFF
--- a/example/main.tf
+++ b/example/main.tf
@@ -628,6 +628,19 @@ resource keycloak_user_template_importer_identity_provider_mapper oidc {
   }
 }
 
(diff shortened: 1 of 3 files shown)


philipp1992 (commented):
looks good, we also need this, please merge

PR #482, opened 2021-02-05
Fix attribute importer and user template mapper for Facebook/Google

Photonios:
## `keycloak_attribute_importer_identity_provider_mapper`
This does not currently work for the Facebook and Google identity providers because of two reasons:

* The computed `IdentityProviderMapper` is `<provider id>-user-attribute-idp-mapper`, while it should be `<provider id>-user-attribute-mapper`.
* The config fields used are `userAttribute` and `jsonField`.

## `keycloak_user_template_importer_identity_provider_mapper`
This does not currently work for the Facebook and Google identity providers because of:

* The computed `IdentityProviderMapper` is `<provider id>-username-idp-mapper`, while it should be `oidc-username-idp-mapper`.

---

The more correct fix for all of these would be to hit `/{realm}/identity-provider/instances/{alias}/mapper-types` to get a list of available mapper types and config fields and to select the right one based on the category.

This short-cut makes it work till that work can be done. This if-else magic was already happening in the attribute importer resource.

### DIFF
--- a/keycloak/identity_provider_mapper.go
+++ b/keycloak/identity_provider_mapper.go
@@ -11,6 +11,7 @@ import (
 
 type IdentityProviderMapperConfig struct {
 	UserAttribute         string                 `json:"user.attribute,omitempty"`
+	UserAttributeName     string                 `json:"userAttribute,omitempty"`
 	Claim                 string                 `json:"claim,omitempty"`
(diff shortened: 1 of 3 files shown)


mrparkers (approved):
seems to work locally for me.  thanks for the PR!

PR #484, opened 2021-02-12
Update README.md

lathspell:
Added Upgrade instructions for Terraform 0.13


lathspell:
I've read https://github.com/mrparkers/terraform-provider-keycloak/issues/313. Twice. Thrice but it still took me a while to get it running, especially figuring out that the "source" line was important.

mrparkers (changes requested):
Thanks for the PR!  My only request is, if this is meant to be an ordered list of instructions, you can replace each asterisk `*` with `1.` so it's more clear that each step is meant to be ran in order.

mrparkers (approved):
thanks for the PR!

PR #495, opened 2021-03-11
Fix marshalling of `false` value in KeycloakBoolQuoted

### LINKED ISSUE #492: DisplayOnConsentScreen broken for resource keycloak_openid_client_scope
### Details

It seems like recent versions of the provider have broken the ability to turn off the "Display On Consent Screen" flag in Client Scopes.

Here's an example of a resource:
```tf
resource "keycloak_openid_client_scope" "test_client_scope" {
  realm_id = keycloak_realm.realm.id
  name = "resource-server/test-client-scope"
  description = "A test client scope"
  include_in_token_scope = false
}
```
And the resulting Client Scope in Keycloak:
image

The "Display On Consent Screen" flag is "On" when it should be "Off" as `consent_screen_text` is `null`.

### Hypothesis

Digging around a bit, I believe this issue is a side effect of commit https://github.com/mrparkers/terraform-provider-keycloak/commit/4206fa1daf3ad33a5a96bf73c3483cc10e619c47

As part of this commit, the attribute `DisplayOnConsentScreen` was changed from a `string` to a `KeycloakBoolQuoted` ([Source](https://github.com/mrparkers/terraform-provider-keycloak/commit/4206fa1daf3ad33a5a96bf73c3483cc10e619c47#diff-c3cec78922b6955904c2cffb6e486d0749818a6d73ff44f643e8fe3a36080badR14)).

However, when marshalled to JSON, `KeycloakBoolQuoted` is serialising `false` values as an empty string `""`. ([Source](https://github.com/mrparkers/terraform-provider-keycloak/blob/8107493195d5364f010e320faff2cc708a5dcb9d/keycloak/util.go#L12-L20))

As a result, the `DisplayOnConsentScreen` attribute which was previously sent to Keycloak as `"false"` is now being sent as `""` ([Source](https://github.com/mrparkers/terraform-provider-keycloak/commit/4206fa1daf3ad33a5a96bf73c3483cc10e619c47#diff-709400b0f0a02f940ef2ebe1f5977a7e1a48f9276c06e6e30b8dcb76c15bd72fL54)). And it seems Keycloak is interpreting that empty string as a positive instead of a negative.

I don't know if this is a wider issue of `KeycloakBoolQuoted` which should be patched globally; or a specificity of `DisplayOnConsentScreen` which means that the marshalling behaviour of `KeycloakBoolQuoted` is not suitable for it.

vlaurin:
Resolves #492

`false` is currently marshalled to `""` (empty string). This is incorrect: `""` represents the default value for an attribute which hasn't been explicitly set.

`""` could be interpreted by Keycloak as either `true` or `false` depending on what the default value is in Keycloak for the attribute considered. For example, in client scopes, both attributes `include.in.token.scope` and `display.on.consent.screen` are evaluated to `true` (ON) when their value is `""`.

As a result, `""` and `"false"` are not equivalent and `"false"` must be used explicitly to accurately turn off Keycloak attributes and not rely on default Keycloak behaviours.

For the same reason, the assumption made when unmarshalling is incorrect as well, though this has a lesser negative effect: https://github.com/mrparkers/terraform-provider-keycloak/blob/master/keycloak/util.go#L24-L25

### DIFF
--- a/keycloak/util.go
+++ b/keycloak/util.go
@@ -11,11 +11,7 @@ type KeycloakBoolQuoted bool
 
 func (c KeycloakBoolQuoted) MarshalJSON() ([]byte, error) {
 	var buf bytes.Buffer
-	if c == false {
-		buf.WriteString(`""`)
-	} else {
-		buf.WriteString(strconv.Quote(strconv.FormatBool(bool(c))))
-	}
+	buf.WriteString(strconv.Quote(strconv.FormatBool(bool(c))))
 	return buf.Bytes(), nil
 }
 


mrparkers:
The code LGTM, thanks for looking into this.

> For the same reason, the assumption made when unmarshalling is incorrect as well, though this has a lesser negative effect: https://github.com/mrparkers/terraform-provider-keycloak/blob/master/keycloak/util.go#L24-L25

Yeah, this makes sense, but I'm not sure if there's anything we can really do about that.  I think the best solution would be for the Keycloak API to stop using quoted boolean values 😄 .  Although if you have any suggestions I'm definitely open to them.

mrparkers (approved):

vlaurin:
@mrparkers Thank for merging. Yes, the unmarshalling one is tricky. I agree that the proper fix should be on the Keycloak side, if Keycloak is assuming default values for attributes then its API should reflect these defaults rather than returning empty strings.

At the level of this provider, I think the ideal course of action would be to not map `""` to either `true` or `false` but rather to something like `nil` though this breaks the `bool` type.

The risk at present is that if an attribute is set as `false` on a resource, but pulled as `""` from Keycloak then no update will take place; if the default value assumed by Keycloak is `true` then that's a mismatch. To recover from it, one would have to first set the value to `true` to force an update to an explicit value, then switch back to `false`.

Instead, I think what should be expected is:

| Keycloak | Terraform | Current behaviour | Expected behaviour |
|---|---|---|---|
| "" | "false" | Noop | Update to "false" |
| "" | "true" | Update to "true" | Update to "true" |
| "false" | "false" | Noop | Noop |
| "true" | "true" | Noop | Noop |

This would require `KeycloakBoolQuoted` to have 3 states which I believe is not permitted with Go `bool` type?

PR #496, opened 2021-03-11
Upgrade to Keycloak v12.0.4

vlaurin:
New patch versions of Keycloak have been released, the latest is now `12.0.4`.


mrparkers (approved):
Thanks for the PR!

PR #499, opened 2021-03-18
support for multivalue attributes for users, groups and roles

### LINKED ISSUE #451: Multivalued user attributes mapping
While using Multivalued user attributes mapping with ## delimiter as described in https://lists.jboss.org/pipermail/keycloak-user/2015-October/003493.html, if this value is set in the console: value1##value2, using the same value in terraform will show a difference, and none if the ## delimiter is removed (value1value2)
resource keycloak_user user {
  attributes = {
    somevalues = "value1##value2"
  }
}
shows a diff in terraform plan

resource keycloak_user user {
  attributes = {
    somevalues = "value1value2"
  }
}
shows no diff in terraform plan

StatueFungus:
Fixes #451

### DIFF
--- a/docs/resources/group.md
+++ b/docs/resources/group.md
@@ -38,8 +38,8 @@ resource "keycloak_group" "child_group_with_optional_attributes" {
   parent_id  = keycloak_group.parent_group.id
   name       = "child-group-with-optional-attributes"
   attributes = {
-    "key1" = "value1"
-    "key2" = "value2"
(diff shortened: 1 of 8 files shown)


mrparkers (approved):
LGTM.  we might want to revisit splitting on 255 characters still, or throwing a validation error if an attribute is longer than 255 characters, but this should work for now.

thanks for the PR!

PR #507, opened 2021-03-30
Introducing additional headers for requests configuration

### LINKED ISSUE #506: [Feature request] Add custom additional headers in each requests
Hello,

I started using this provider recently and it works fine, amazing job!

Now the hard part: our production keycloak is hidden behind a proxy which filters requests not having a special custom header (key and value).

I'm currently working on implementing a new parameter for the setup of the provider itself to include this feature. 
This will mirror what is already available in a python library we were using: https://github.com/marcospereirampj/python-keycloak/blob/ff4669979973c3744dfcfb415840b44dcaf88a5c/keycloak/keycloak_openid.py#L55

sl-benoitoyez:
Closes #506 

This is my first time developping in golang, let me know if things are missing, I'll gladly fix them right away.

Thanks in advance!


mrparkers (approved):
LGTM, thanks for the PR!

PR #512, opened 2021-04-06
keycloak_saml_identity_provider: support entity_id as required field

PSanetra:
This PR adds support for the required field `entity_id` on the resource `keycloak_saml_identity_provider`.

Notice: This is a breaking change. I think it is weird that this field is required in the Admin Console but the REST API seems to be ok if the field is missing. I would still make it a required field as using the Admin Console on an SAML identity provider may otherwise result in unexpected behavior.

Fixes https://github.com/mrparkers/terraform-provider-keycloak/issues/510

### DIFF
--- a/docs-old/resources/keycloak_saml_identity_provider.md
+++ b/docs-old/resources/keycloak_saml_identity_provider.md
@@ -10,6 +10,7 @@ SAML (Security Assertion Markup Language) identity providers allows to authentic
 resource "keycloak_saml_identity_provider" "realm_identity_provider" {
   realm = "my-realm"
   alias = "my-idp"
+  entity_id = "https://domain.com/entity_id"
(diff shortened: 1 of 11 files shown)


mrparkers:
Hey @PSanetra, thanks for the PR.  this change looks fine overall, but any of the existing tests that use this resource will need to be updated to use this new attribute.  would you be able to do that in order to get the tests passing?

thanks!

PSanetra:
@mrparkers I have fixed the tests and adjusted the documentation

mrparkers (approved):
looks great, thanks!  🚀

PSanetra:
@mrparkers Is there some ETA for a new release with this change or is it possible to get a pre-release?

mrparkers:
@PSanetra v3.0.0 has been released today!

PR #518, opened 2021-04-20
Add datasource keycloak_client_description_converter

Kyos:
This PR implements the client_description_converter endpoint as a datasource (documentation of this endpoint can be found here : https://www.keycloak.org/docs-api/5.0/rest-api/index.html#_paths).

I had to add a method in keycloak_client to be able to send xml data without escaping.

Please let me know if this suit you 😃 

Thanks by advance !

### DIFF
--- a/keycloak/generic_client_description_converter.go
+++ b/keycloak/generic_client_description_converter.go
@@ -1,0 +1,63 @@
+package keycloak
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type GenericClientRepresentation struct {
+	Access                             map[string]string              `json:"access"`
+	AdminUrl                           string                         `json:"adminUrl"`
+	Attributes                         map[string]string              `json:"attributes"`
+	AuthenticationFlowBindingOverrides map[string]string              `json:"authenticationFlowBindingOverrides"`
+	AuthorizationServicesEnabled       bool                           `json:"authorizationServicesEnabled"`
+	AuthorizationSettings              map[string]string              `json:"authorizationSettings"`
+	BaseUrl                            string                         `json:"baseUrl"`
+	BearerOnly                         bool                           `json:"bearerOnly"`
+	ClientAuthenticatorType            string                         `json:"clientAuthenticatorType"`
+	ClientId                           string                         `json:"clientId"`
+	ConsentRequired                    string                         `json:"consentRequired"`
+	DefaultClientScopes                []string                       `json:"defaultClientScopes"`
+	DefaultRoles                       []string                       `json:"defaultRoles"`
+	Description                        string                         `json:"description"`
+	DirectAccessGrantsEnabled          bool                           `json:"directAccessGrantsEnabled"`
+	Enabled                            bool                           `json:"enabled"`
+	FrontchannelLogout                 bool                           `json:"frontchannelLogout"`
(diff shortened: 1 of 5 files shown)


Kyos:
ping @mrparkers for review please 😃

mrparkers:
Could you help me understand what the `client-description-converter` does?  If you could provide me with a use case and some sample HCL code, that would help greatly.

Thanks!

Kyos:
@mrparkers Yes of course ! This endpoint is used to transform any saml metadata or oidc json into the internal representation for keycloak (it is used for example in the administration panel of keycloak when you import a client from a url). I use the result of this endpoint to create different resource on keycloak (saml client, idp etc)

I have (personally) 2 use cases to use this endpoint : 

- First one : Massive import from a list of metadata url (or files)
- Second one (not yet done) : Auto refresh of saml certificate

Here is an example code (using provider http also but it works with the file provider)

`
##
# CONFIG
##

terraform {
  required_version = ">= 0.13.0"

  required_providers {
    keycloak = {
      source = "terraform.example.com/mrparkers/keycloak"
    }
    http = {
      source = "hashicorp/http"
    }
  }
}

provider "http" {}

provider "keycloak" {}

##
# VARIABLES
##

variable "urls" {
  type = list(string)
  default = [
    // put any saml federation uri here
    //"http://localhost:9000/adfs/test/saml/federation.xml"
  ]
}

##
# DATA
##

data "http" "federation" {
  for_each = toset(var.urls)

  url = each.key
  request_headers = {
    Accept = "application/json"
  }
}

data "keycloak_client_description_converter" "client" {
  for_each = toset(var.urls)

  realm_id = "fake"
  body     = data.http.federation[each.key].body
}

resource "keycloak_saml_client" "saml_client" {
  for_each = toset(var.urls)
  realm_id = "fake"

  client_id = data.keycloak_client_description_converter.client[each.key].client_id
  name      = each.key

  sign_documents      = lookup(data.keycloak_client_description_converter.client[each.key].attributes, "saml.client.signature")
  signing_certificate = lookup(data.keycloak_client_description_converter.client[each.key].attributes, "saml.signing.certificate")
  signature_algorithm = lookup(data.keycloak_client_description_converter.client[each.key].attributes, "saml.signature.algorithm")

  assertion_consumer_post_url = lookup(data.keycloak_client_description_converter.client[each.key].attributes, "saml_assertion_consumer_url_post")

  valid_redirect_uris = [
    lookup(data.keycloak_client_description_converter.client[each.key].attributes, "saml_assertion_consumer_url_post")
  ]

}
`

I found this endpoint very useful to "synchronise" multiple federation endpoint with the config I have on my keycloak. (before that I had  a script which parse urls and extract certificates and others options).

Let me know if you need more information 😃 

Thanks by advance

mrparkers:
I see, so the `client-description-converter` endpoint doesn't actually create anything on the Keycloak server, but rather, it uses the Keycloak server to create a representation of what this client would look like.  Is this correct?

I don't have something like `adfs` on hand to run locally, but I just want to make sure that this data source isn't actually making any changes on the Keycloak side.

Kyos:
@mrparkers yes you got it. It doesn't change anything on the keycloak side, this is just an endpoint used for data transformation.

mrparkers (approved):
thanks for helping me understand this, everything looks good!  I'll do a follow-up here to get some docs added so this can be included in the next release.

PR #523, opened 2021-04-30
fix: use consistent config / extra_config across identity providers

### LINKED ISSUE #520: Since version 3.0.0 extra_config parameter on the keycloak_oidc_identity_provider resource for fields syncMode and guiOrder is ignored
Since version 3.0.0 extra_config parameter on the keycloak_oidc_identity_provider resource for fields syncMode and guiOrder is ignored

Terraform still sees those fields in the tf files and still states (during planning) that it needs to add those fields to keycloak,
But it actually does not send it to keycloak api. Even worse, it removes previous set with version 2.3.0 values.
:-( I have now a whole qa environment without syncMode ="FORCE" and without guiOrder on all my idps :-(
This breaks old configs, I will need to revert to 2.3.0, I will try to look at a later time what could be the culprit

example config:

```
resource "keycloak_oidc_identity_provider" "my_idp" {
  realm                 = keycloak_realm.myrealm.id
  //other fields 

  extra_config = {
    syncMode         = "FORCE"
    guiOrder          =  "1"
  }
}
```

mrparkers:
- `gui_order` and `sync_mode` are now supported by all identity providers as top-level attributes
- `extra_config` attribute now fails validation if a key is used that conflicts with the existing identity provider config JSON model 

This doesn't really "fix" the root problem, but the extra validation here will prevent potentially bad configuration for an identity provider from reaching a Keycloak instance.

Fixes #520 

cc @tomrutsaert @alex-hempel

### DIFF
--- a/docs/resources/oidc_google_identity_provider.md
+++ b/docs/resources/oidc_google_identity_provider.md
@@ -22,10 +22,7 @@ resource "keycloak_oidc_google_identity_provider" "google" {
   client_secret = var.google_identity_provider_client_secret
   trust_email   = true
   hosted_domain = "example.com"
-
-  extra_config = {
-    "syncMode" = "IMPORT"
-  }
+  sync_mode     = "IMPORT"
 }
 ```
 
@@ -49,6 +46,8 @@ resource "keycloak_oidc_google_identity_provider" "google" {
 - `accepts_prompt_none_forward_from_client` - (Optional) When `true`, unauthenticated requests with `prompt=none` will be forwarded to Google instead of returning an error. Defaults to `false`.
 - `disable_user_info` - (Optional) When `true`, disables the usage of the user info service to obtain additional user information. Defaults to `false`.
 - `hide_on_login_page` - (Optional) When `true`, this identity provider will be hidden on the login page. Defaults to `false`.
+- `sync_mode` - (Optional) The default sync mode to use for all mappers attached to this identity provider. Can be once of `IMPORT`, `FORCE`, or `LEGACY`.
+- `gui_order` - (Optional) A number defining the order of this identity provider in the GUI.
 - `extra_config` - (Optional) A map of key/value pairs to add extra configuration to this identity provider. This can be used for custom oidc provider implementations, or to add configuration that is not yet supported by this Terraform provider.
 
 ## Attribute Reference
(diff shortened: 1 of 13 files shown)


provider/generic_keycloak_identity_provider.go
```diff
@@ -89,12 +99,72 @@ func resourceKeycloakIdentityProvider() *schema.Resource {
 				Default:     "",
 				Description: "Alias of authentication flow, which is triggered after each login with this identity provider. Useful if you want additional verification of each user authenticated with this identity provider (for example OTP). Leave this empty if you don't want any additional authenticators to be triggered after login with this identity provider. Also note, that authenticator implementations must assume that user is already set in ClientSession as identity provider already set it.",
 			},
+			// all schema values below this point will be configuration values that are shared among all identity providers
+			"extra_config": {
```
tomrutsaert, line 103: So basically extra config(and gui_order and sync_mode) was moved to the generic identity provider, :+1:

provider/generic_keycloak_identity_provider.go
```diff
@@ -89,12 +99,72 @@ func resourceKeycloakIdentityProvider() *schema.Resource {
...
+
+						if jsonKey != "-" && field.CanSet() {
+							if _, ok := extraConfig[jsonKey]; ok {
+								diags = append(diags, diag.Diagnostic{
+									Severity: diag.Error,
+									Summary:  "Invalid extra_config key",
+									Detail:   fmt.Sprintf(`extra_config key "%s" is not allowed, as it conflicts with a top-level schema attribute`, jsonKey),
+									AttributePath: append(path, cty.IndexStep{
+										Key: cty.StringVal(jsonKey),
+									}),
```
tomrutsaert, line 125: This is great, This will help people to understand what is going on, even if they did not read the release notes

docs/resources/oidc_google_identity_provider.md
```diff
@@ -22,10 +22,7 @@ resource "keycloak_oidc_google_identity_provider" "google" {
   client_secret = var.google_identity_provider_client_secret
   trust_email   = true
   hosted_domain = "example.com"
-
-  extra_config = {
-    "syncMode" = "IMPORT"
-  }
```
tomrutsaert, line 28: I would still give an example of extra config:

```
  extra_config = {
    "myCustomConfigKey" = "myValue"
  }
```

tomrutsaert (approved):
LGTM, This a very good improvement for my personal use-cases

PR #527, opened 2021-05-07
Members pagination

### LINKED ISSUE #526: Implement pagination for group memberships
The current keycloak_group_memberships implementation doesn't paginate on refresh. 

This might break in environment where you can find Keycloak instances with groups with lots of members and not enough resources to respond under the timeout breaking the execution.

sl-benoitoyez:
Closes #526 

I set arbitrarily the pagination value to 50, I can change it if you think another value would be better.

The changes are only on the backstage side, no changes to the resource signature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## v3.0.1 (May 5, 2021)
+
+BUG FIXES:
+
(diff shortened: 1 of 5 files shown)


mrparkers (approved):
Apologies for the late review.  This looks good, and the existing test for this resource with 100 members seems to pass.

Thanks for the PR!

PR #529, opened 2021-05-11
Allow configuring SAML Signuture Key Name

torstenwalter:
image


provider/resource_keycloak_saml_client.go
```diff
@@ -93,6 +94,12 @@ func resourceKeycloakSamlClient() *schema.Resource {
 				Optional:     true,
 				ValidateFunc: validation.StringInSlice(keycloakSamlClientSignatureAlgorithms, false),
 			},
+			"signature_key_name": {
```
mrparkers, line 97: since we never call a `data.Get` or `data.Set` for this schema attribute, it is essentially ignored.  take a look at the `mapToSamlClientFromData` and `mapToDataFromSamlClient` functions to see where these should be added.

mrparkers (changes requested):
Thanks for the PR!  it looks like you just missed one thing here

PR #536, opened 2021-05-25
Add import attribute for keycloak_openid_client

spirius:
Keycloak creates few openid clients by default (like account, account-console) and the only way to manage those clients with terraform is by manual import. This PR adds `import` attribute to `keycloak_openid_client` resources, which controls whether or not the client should be created or "imported". Note, that if `import` attribute is true, the resource will not be removed during destruction.

With this change, for example, one can disable the `account` client with a configuration like this:

```hcl
  realm = "test-realm"
}

resource "keycloak_openid_client" "account" {
  realm_id    = keycloak_realm.realm.id
  client_id   = "account"
  access_type = "PUBLIC"
  enabled     = false
  import      = true
}
```

Acceptance tests:

```
=== RUN   TestAccKeycloakOpenidClient_import
=== PAUSE TestAccKeycloakOpenidClient_import
=== CONT  TestAccKeycloakOpenidClient_import
--- PASS: TestAccKeycloakOpenidClient_import (19.38s)
```

### DIFF
--- a/provider/resource_keycloak_openid_client.go
+++ b/provider/resource_keycloak_openid_client.go
@@ -206,6 +206,11 @@ func resourceKeycloakOpenidClient() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"import": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 		CustomizeDiff: customdiff.ComputedIf("service_account_user_id", func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) bool {
 			return d.HasChange("service_accounts_enabled")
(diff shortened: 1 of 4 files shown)


provider/resource_keycloak_openid_client.go
```diff
@@ -401,9 +406,23 @@ func resourceKeycloakOpenidClientCreate(data *schema.ResourceData, meta interfac
...
-	if err != nil {
-		return err
+	if data.Get("import").(bool) {
+		existingClient, err := keycloakClient.GetOpenidClientByClientId(client.RealmId, client.ClientId)
+		if err != nil {
+			return err
+		}
+		rootUrlString := data.Get("root_url").(string)
+		client.Id = existingClient.Id
+		client.RootUrl = &rootUrlString
```
mrparkers, line 416: why are we only setting `RootUrl` for the client as opposed to other fields.  should we just call `getOpenidClientFromData` here?

mrparkers (commented):
Overall this looks good, I just had one question

mrparkers:
I made some changes to this and opened a new PR #747.

Thanks for your work on this!

PR #376, opened 2020-09-03
add cache policy parameter to ldap user federation

cgroschupp:

PR #537, opened 2021-05-25
add cache settings for `custom_user_federation`

### LINKED ISSUE #521: Allow specifying cache settings for custom_user_federation
The `CustomUserFederation` allows specifying the `CachePolicy` using one of the pre-defined defaults, but does not have the struct fields needed to define custom cache settings. 

For example, here are the cache-related variables in the `CustomUserFederation` struct: 
```
	CachePolicy string
```

and here are the cache-related variables from `LdapUserFederation`:  

```
	CachePolicy    string
	MaxLifespan    string // duration string (ex: 1h30m)
	EvictionDay    *int
	EvictionHour   *int
	EvictionMinute *int
```

bl00mber:
close #521
see #376


mrparkers:
Hey @bl00mber, thanks for the PR.  Could you modify one of the [existing test cases](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/resource_keycloak_custom_user_federation_test.go) to include some of these new settings?  I'll be happy to merge this PR once this is done.

Thanks!

PR #542, opened 2021-06-07
Fix documentation typo

pstanton237:


mrparkers (approved):
Thanks for the fix, LGTM

PR #549, opened 2021-07-02
Add a simple retry mechanism to keycloak get method

Useurmind:
Hi,

we have encountered a problem with the provider when configuring our keycloak instance.

From what we can determine from the terraform logs, a get request fails when reading the state from keycloak. It seems as if the creation of the object is not yet fully finished (probably done asynchonously in keycloak) when the read starts.

Here is some extract from the logs:

```
2021-07-01T09:53:56.897+0200 [DEBUG] plugin.terraform-provider-keycloak_v3.1.1.exe: 2021/07/01 09:53:56 [DEBUG] Sending POST to /auth/admin/realms/master/components
2021-07-01T09:53:56.897+0200 [DEBUG] plugin.terraform-provider-keycloak_v3.1.1.exe: 2021/07/01 09:53:56 [DEBUG] Request body: {"name":"firstName","providerId":"user-attribute-ldap-mapper","providerType":"org.keycloak.storage.ldap.mappers.LDAPStorageMapper","parentId":"04e6cd95-2d57-4c12-a8bf-b66b040b69ee","config":{"always.read.value.from.ldap":["false"],"is.mandatory.in.ldap":["false"],"ldap.attribute":["givenName"],"read.only":["false"],"user.model.attribute":["firstName"]}}
2021-07-01T09:53:56.897+0200 [DEBUG] plugin.terraform-provider-keycloak_v3.1.1.exe: 2021/07/01 09:53:56 [DEBUG] Sending POST to /auth/admin/realms/master/components
2021-07-01T09:53:56.897+0200 [DEBUG] plugin.terraform-provider-keycloak_v3.1.1.exe: 2021/07/01 09:53:56 [DEBUG] Request body: {"name":"telephoneNumber","providerId":"user-attribute-ldap-mapper","providerType":"org.keycloak.storage.ldap.mappers.LDAPStorageMapper","parentId":"04e6cd95-2d57-4c12-a8bf-b66b040b69ee","config":{"always.read.value.from.ldap":["false"],"is.mandatory.in.ldap":["false"],"ldap.attribute":["telephoneNumber"],"read.only":["false"],"user.model.attribute":["telephoneNumber"]}}
2021-07-01T09:53:56.952+0200 [DEBUG] plugin.terraform-provider-keycloak_v3.1.1.exe: 2021/07/01 09:53:56 [DEBUG] Response: 201 Created
2021-07-01T09:53:56.952+0200 [DEBUG] plugin.terraform-provider-keycloak_v3.1.1.exe: 2021/07/01 09:53:56 [DEBUG] Sending GET to /auth/admin/realms/master/components/f361392e-9145-4177-92fc-a11e44f76266
2021-07-01T09:53:57.099+0200 [DEBUG] plugin.terraform-provider-keycloak_v3.1.1.exe: 2021/07/01 09:53:57 [DEBUG] Response: 404 Not Found
2021-07-01T09:53:57.099+0200 [DEBUG] plugin.terraform-provider-keycloak_v3.1.1.exe: 2021/07/01 09:53:57 [DEBUG] Response body: {"error":"Could not find component"}
2021-07-01T09:53:57.099+0200 [DEBUG] plugin.terraform-provider-keycloak_v3.1.1.exe: 2021/07/01 09:53:57 [WARN] Removing resource with id f361392e-9145-4177-92fc-a11e44f76266 from state as it no longer exists
```

A solution to this problem would be a retry mechanism in the read method for keycloak. As I do not fully know in which situations this problem could arise, I have put it into the getRaw method of the keycloak client.

Feel free to critizize as this is only a rough draft of what the solution could be. I am also not fully sure how well the solution fits into the terraform provider lifecycle.

Kind Regards


mrparkers:
While I don't necessarily have an issue with this, I'm more curious about what's causing the underlying problem here.  Have you ran into this issue with your Keycloak installation before?  Or is it just while using this provider?  And is this the only resource that's affected, or are there other resources / keycloak features where this seems to happen?

Also, would you be able to re-submit this PR using a different branch name?  I think CircleCI isn't running the tests because the origin branch is called `master`.

Useurmind:
Regarding the occurence of the problem: It's hard to say. We recently noticed that when configuring a fresh keycloak with a terraform script, we encounter this problem. It really is configuring a lot of elements and the reason could be a performance thing. When applying only small portions of the configuration via terraform we never encountered this problem yet. We only saw it for components thus far.
My current guess is that keycloak configures them asynchronously.

As I already said this is a pretty crude implementation, do you have an idea where the configuration of these retry settings would fit? Like in the provider block or as environment variables?

As I need to do some more work on this before we can accept it, it is no problem to create a proper feature branch for it.

mrparkers:
I think this would have to be a togglable setting within the provider before it could be accepted.  As I mentioned earlier, I'm concerned that this is just covering up a potential problem with your Keycloak installation that would manifest itself in other ways besides configuring it with this provider.

With regards to how these settings are stored within Keycloak (asynchronously or otherwise), it really depends on your jboss/wildfly configuration under the hood.  I used to know a lot more about it than I do now, but here's a StackOverflow question I made about this topic a few years ago that might help demonstrate the number of possibilities here: https://stackoverflow.com/questions/57467224/keycloak-remote-store-configuration-for-dedicated-infinispan-cluster.  The configuration that is used for this provider's acceptance tests is much simpler than this (using an embedded h2 database instead of something like postgres, for example), and these problems you're describing don't seem to manifest themselves during acceptance tests.

Anyways, I think I'm okay with this change, as long as this is an opt-in setting that can be specified as a part of the provider's configuration.

mrparkers:
I've since updated the `keycloak` package to use [go-retryablehttp](https://github.com/hashicorp/go-retryablehttp) for the http client, so errors should automatically be retried.  So I think this can be considered resolved.

PR #560, opened 2021-07-16
Update resource_keycloak_ldap_role_mapper.go

DOboznyi:
Seems as this field was accidentally removed


DOboznyi:
@mrparkers could you check this change?

mrparkers:
Hey @DOboznyi, maybe I'm missing something here, but it looks like this field was removed from Keycloak.  I tested Keycloak 11, 12, and 13, and none of them have this setting anymore.  That's why I removed it.

Are you using a version of Keycloak that still uses this field?  If so, I can accept this PR and cut a release for you, but I would also want to see the docs updated to specify this as a supported field again.

DOboznyi:
Hey. They have such setting and as i can see from keycloak code it's still present ([Link](https://github.com/keycloak/keycloak/pull/6945))

mrparkers (approved):
I see, my mistake then.  Thanks for the PR!  I'll get the docs fixed as well.

DOboznyi:
When will be new release with this fix?

PR #576, opened 2021-08-09
Add github action running acceptance tests

arkadiusjonczek:
Hi @mrparkers ,

this PR adds a new github action running the integration tests. The workflow is based on the existing circle ci workflow so it should be quite the same.

I think it is easier because it can be used in forks and PR's so it could replace circle ci workflow. (if you like)

Best regards,
Arkadius

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,63 @@
+name: test
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        keycloak-version:
+          - '13.0.1'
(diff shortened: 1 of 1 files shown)


mrparkers:
Hi @arkadiusjonczek, I'm sorry for neglecting this PR.  I actually do plan on switching to GitHub actions soon, using some custom runners within a Kubernetes cluster that I maintain.  It will most likely look very similar to this, with the exception of the `services`, as I plan to have long-running Keycloak instances within the Kubernetes cluster that these jobs will run in.

So I probably won't accept this PR as-is, but I'll draw from this as inspiration when I'm ready to work on this.

Thanks!

PR #578, opened 2021-08-19
feat: add oauth2 device authorization grant

shellrausch:
Suppport for the relatively recent OAuth2 Device Authorization Grant. The corresponding settings are present in a realm and a client.

In Keycloak the settings can be found as following:

Realm (Tokens tab):
- OAuth 2.0 Device Code Lifespan
- OAuth 2.0 Device Polling Interval 

Client (Settings tab):
- OAuth 2.0 Device Authorization Grant Enabled 
- OAuth 2.0 Device Code Lifespan
- OAuth 2.0 Device Polling Interval 

Please note the [OAuth2 Device Authorization Grant](https://www.keycloak.org/docs/latest/release_notes/#oauth-2-0-device-authorization-grant-rfc-8628) was firstly introduced in Keycloak 13. This is also the reason why the tests with Keycloak 11 and 12 fail.

### DIFF
--- a/docs/resources/realm.md
+++ b/docs/resources/realm.md
@@ -132,6 +132,11 @@ The arguments below should be specified as [Go duration strings](https://golang.
 - `access_code_lifespan_user_action` - (Optional) The maximum amount of time a user has to complete login related actions, such as updating a password.
 - `action_token_generated_by_user_lifespan` - (Optional) The maximum time a user has to use a user-generated permit before it expires.
 - `action_token_generated_by_admin_lifespan` - (Optional) The maximum time a user has to use an admin-generated permit before it expires.
+- `oauth2_device_code_lifespan` - (Optional) The maximum amount of time a client has to finish the device code flow before it expires.
+
+The attributes below should be specified in seconds.
+
+- `oauth2_device_polling_interval` - (Optional) The minimum amount of time in seconds that the client should wait between polling requests to the token endpoint.
 
 ### SMTP
 
--- a/provider/resource_keycloak_openid_client_test.go
+++ b/provider/resource_keycloak_openid_client_test.go
@@ -291,6 +291,42 @@ func TestAccKeycloakOpenidClient_ClientTimeouts_basic(t *testing.T) {
 	})
 }
 
+func TestAccKeycloakOpenidClient_Device_basic(t *testing.T) {
+	if ok, _ := keycloakClient.VersionIsGreaterThanOrEqualTo(keycloak.Version_13); !ok {
+		t.Skip()
+	}
+
+	t.Parallel()
+	clientId := acctest.RandomWithPrefix("tf-acc")
+
+	oauth2DeviceCodeLifespan := "300"
+	oauth2DevicePollingInterval := "60"
+	oauth2DeviceAuthorizationGrantEnabled := true
+
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: testAccProviderFactories,
+		PreCheck:          func() { testAccPreCheck(t) },
+		CheckDestroy:      testAccCheckKeycloakOpenidClientDestroy(),
+		Steps: []resource.TestStep{
+			{
+				Config: testKeycloakOpenidClient_oauth2DeviceTimes(clientId,
+					oauth2DeviceCodeLifespan, oauth2DevicePollingInterval, oauth2DeviceAuthorizationGrantEnabled,
+				),
+				Check: testAccCheckKeycloakOpenidClientOauth2Device("keycloak_openid_client.client",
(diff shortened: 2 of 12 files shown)


shellrausch:
Please note the [OAuth2 Device Authorization Grant](https://www.keycloak.org/docs/latest/release_notes/#oauth-2-0-device-authorization-grant-rfc-8628) was firstly introduced in Keycloak 13. This also the reason why the tests with Keycloak 11 and 12 fail.

shellrausch:
Please note the [OAuth2 Device Authorization Grant](https://www.keycloak.org/docs/latest/release_notes/#oauth-2-0-device-authorization-grant-rfc-8628) was firstly introduced in Keycloak 13. This also the reason why the tests with Keycloak 11 and 12 fail.

provider/resource_keycloak_realm.go
```diff
@@ -352,6 +352,17 @@ func resourceKeycloakRealm() *schema.Resource {
 				Computed:         true,
 				DiffSuppressFunc: suppressDurationStringDiff,
 			},
+			"oauth2_device_code_lifespan": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Computed:         true,
+				DiffSuppressFunc: suppressDurationStringDiff,
+			},
+			"oauth2_device_polling_interval": {
```
mrparkers, line 361: should this attribute also be a duration string?

provider/resource_keycloak_openid_client_test.go
```diff
@@ -256,6 +256,38 @@ func TestAccKeycloakOpenidClient_ClientTimeouts_basic(t *testing.T) {
 	})
 }
 
+func TestAccKeycloakOpenidClient_Device_basic(t *testing.T) {
+	t.Parallel()
+	clientId := acctest.RandomWithPrefix("tf-acc")
+
+	oauth2DeviceCodeLifespan := "300"
```
mrparkers, line 263: since we're handling this as a duration string, I expected this to have a different format
```suggestion
	oauth2DeviceCodeLifespan := "300s"
```

mrparkers (commented):
Thanks for the PR, overall this looks good, just had a couple of comments.

For the failing Keycloak 11/12 tests, you can use this to skip them:

```go
if ok, _ := keycloakClient.VersionIsGreaterThanOrEqualTo(keycloak.Version_13); !ok {
    t.Skip()
}
```

provider/resource_keycloak_realm.go
shellrausch: The polling interval should be set as seconds. There is also no possibility to choose other time units in the UI, see attached screenshot.

provider/resource_keycloak_openid_client_test.go
shellrausch: The client settings are using for every time based option just a string without a unit.

Examples in other test cases: 

`TestAccKeycloakOpenidClient_AccessToken_basic`:
- `accessTokenLifespan := "1800"`

`TestAccKeycloakOpenidClient_ClientTimeouts_basic`:
- `offlineSessionMaxLifespan := "1900"`
- `sessionMaxLifespan := "2100"`

mrparkers: My bad, I got this confused with the `keycloak_realm` attributes, which are duration strings.

docs/resources/realm.md
```diff
@@ -132,6 +132,8 @@ The arguments below should be specified as [Go duration strings](https://golang.
 - `access_code_lifespan_user_action` - (Optional) The maximum amount of time a user has to complete login related actions, such as updating a password.
 - `action_token_generated_by_user_lifespan` - (Optional) The maximum time a user has to use a user-generated permit before it expires.
 - `action_token_generated_by_admin_lifespan` - (Optional) The maximum time a user has to use an admin-generated permit before it expires.
+- `oauth2_device_code_lifespan` - (Optional) The maximum amount of time a client has to finish the device code flow before it expires.
+- `oauth2_device_polling_interval` - (Optional) The minimum amount of time in seconds that the client should wait between polling requests to the token endpoint.
```
mrparkers, line 136: So I understand why this attribute is an int, because in Keycloak it has to be specified as seconds, but this part of the docs is misleading because this is within a section that states that these attributes are duration strings.  I think we'll probably want to move this somewhere else, or allow the `oauth2_device_polling_interval` attribute to be specified as a duration string, and converted to seconds for the sake of the Keycloak API.  So, for example, you could do `oauth2_device_polling_interval = "1m"`, which would be converted to 60 seconds before it's sent to the Keycloak API.

I'm fine with either approach, but I don't think we can leave this as-is.

mrparkers (approved):
Sorry for taking so long to get back to this.  I'm happy with the docs change so I think this can be merged now.  Thanks again for submitting this PR!

Issue #568, opened 2021-08-03
Clients Backchannel* settings support


myardyas:
This plugin is really cool. Thanks! Implementing it in our environment. Is there any chance to support client's backchannel* settings like:
- Backchannel Logout URL
- Backchannel Logout Session Required
- Backchannel Logout Revoke Offline Sessions
?

myardyas:
@mrparkers Thanks for implementing, but settings look like missed up -)
https://github.com/mrparkers/terraform-provider-keycloak/blob/v3.4.0/provider/resource_keycloak_openid_client.go#L299-300

mrparkers:
Oops, I fixed this in #600

Issue #387, opened 2020-09-17
[Clients] - Add attributes


slavko-vega:
Hi  @mrparkers,

we noticed that Clients are missing attributes, which are needed by our project. Currently the Attributes, as they are implemented, are limited to a subset of Client's properties which are used by Keycloak. What we actually need is possibility to add our custom attributes as it's possible for other resources.

Haven't seen that the issue is reported yet.  I'd like to contribute on its implementation, if you consider this as an issue.

Do you agree with that?

mrparkers:
Hey @slavko-vega, that sounds great! I'd definitely accept a PR that adds an `attributes` argument to the `keycloak_openid_client` and/or `keycloak_saml_client` resources.

slavko-vega:
Hey @mrparkers I'm glad to contribute on this issue and PR is prepared.

mrparkers:
Closed via #589

PR #579, opened 2021-08-19
add extra_config for openid clients to handle custom attributes (#387)

olivierboudet:
Hello @mrparkers ,

As the PR #389 seems blocked since several months and because I need to manage custom attributes on openid clients, I tried to make a new PR. This new PR handles only a new `extra_config` field on `keycloak_openid_client`.

However I have a question : as Keycloak recently added 2 new attributes with default values (`backchannel.logout.revoke.offline.tokens=false` and `backchannel.logout.session.required=true`, as described in #568), how should we handle these attributes ? I modified all the tests to add these new attributes in data, is this correct waiting for an official support for these fields ?
The advantage on this solution is that it allows to set all new attributes added by Keycloak before an official support in the provider.
The drawback is it become mandatory to set this extra_config when using KC >= 12.x with a provider which do not support this fields natively.
```
extra_config = {
  "backchannel.logout.revoke.offline.tokens" = "false"
  "backchannel.logout.session.required"      = "true"
}
```

### DIFF
--- a/docs/resources/openid_client.md
+++ b/docs/resources/openid_client.md
@@ -31,6 +31,11 @@ resource "keycloak_openid_client" "openid_client" {
   ]
 
   login_theme = "keycloak"
+      
+  extra_config = {
+		"key1" = "value1"
+		"key2" = "value2"
+    }
 }
 ```
 
@@ -81,6 +86,7 @@ is set to `true`.
 - `backchannel_logout_url` - (Optional) The URL that will cause the client to log itself out when a logout request is sent to this realm. If omitted, no logout request will be sent to the client is this case.
 - `backchannel_logout_session_required` - (Optional) When `true`, a sid (session ID) claim will be included in the logout token when the backchannel logout URL is used. Defaults to `true`.
 - `backchannel_logout_revoke_offline_sessions` - (Optional) Specifying whether a "revoke_offline_access" event is included in the Logout Token when the Backchannel Logout URL is used. Keycloak will revoke offline sessions when receiving a Logout Token with this event.
+- `extra_config` - (Optional) A map of key/value pairs to add extra configuration attributes to this client. This can be used for custom attributes, or to add configuration attributes that is not yet supported by this Terraform provider. Use this attribute at your own risk, as s may conflict with top-level configuration attributes in future provider updates.
 
 ## Attributes Reference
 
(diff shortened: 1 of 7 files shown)


docs/resources/openid_client.md
```diff
@@ -76,8 +81,10 @@ is set to `true`.
 - `authorization` - (Optional) When this block is present, fine-grained authorization will be enabled for this client. The client's `access_type` must be `CONFIDENTIAL`, and `service_accounts_enabled` must be `true`. This block has the following arguments:
     - `policy_enforcement_mode` - (Required) Dictates how policies are enforced when evaluating authorization requests. Can be one of `ENFORCING`, `PERMISSIVE`, or `DISABLED`.
     - `decision_strategy` - (Optional) Dictates how the policies associated with a given permission are evaluated and how a final decision is obtained. Could be one of `AFFIRMATIVE`, `CONSENSUS`, or `UNANIMOUS`. Applies to permissions.
-    - `allow_remote_resource_management` - (Optional) When `true`, resources can be managed remotely by the resource server. Defaults to `false`.
+    - `allow_remote_resource_management` - (Optional) When `true`, resources can be managed remotely by the resource server. Defaults to 
+    `false`.
```
mrparkers, line 85: looks like an extra newline was added here

provider/resource_keycloak_openid_client.go
```diff
@@ -211,6 +214,37 @@ func resourceKeycloakOpenidClient() *schema.Resource {
...
 			},
+			"extra_config": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				// you aren't allowed to specify any keys in extra_config that could be defined as top level attributes
+				ValidateDiagFunc: func(v interface{}, path cty.Path) diag.Diagnostics {
+					var diags diag.Diagnostics
+
+					extraConfig := v.(map[string]interface{})
+					value := reflect.ValueOf(&keycloak.IdentityProviderConfig{}).Elem()
```
mrparkers, line 225: We need to validate the `extra_config` fields against the `keycloak.OpenidClientAttributes` type here, since that's what we use when talking to the Keycloak API for openid clients.

```suggestion
					value := reflect.ValueOf(&keycloak.OpenidClientAttributes{}).Elem()
```

mrparkers (changes requested):
Hey @olivierboudet, thanks for the PR.  I think supporting `extra_config` is fine for this resource.

As far as the backchannel attributes goes, I went ahead and opened #583 to support these as top level attributes.  So you can remove the changes that adds these attributes through `extra_config`.

mrparkers (approved):
Thanks for the PR, LGTM 🚀

Issue #569, opened 2021-08-03
Support for setting realm keys


ringods:
There is support for retrieving the realm keys (#126), but this is a request to add support for configuring the keys of a realm.

![imgpsh_mobile_save](https://user-images.githubusercontent.com/77923/128034897-a4da80a6-e8a2-47f7-a2bb-7f94a3849d39.jpg)

![659e9f82-dd6e-4685-b26d-b90ad0abf89f](https://user-images.githubusercontent.com/77923/128034918-58858854-7d76-4357-ba4a-455c13026bef.jpg)

![14786ea6-440a-4be9-98bd-458d228238c0](https://user-images.githubusercontent.com/77923/128034937-2aec452f-ef30-41b0-aff5-c4a86a69ba05.jpg)

Admin Guide section: https://www.keycloak.org/docs/latest/server_admin/index.html#realm_keys

After some debugging together with my colleague @vlad-kirichenko, we found out that this part of the admin UI calls into this set of REST endpoints:

* `/{realm}/components`
* `/{realm}/components/{id}`

These endpoints not only support managing the `Keystores`, but also the `Client Registration` section of a realm.

Retrieving all the components (keystores & client registrations) is done with a `GET`:

https://www.keycloak.org/docs-api/15.0/rest-api/index.html#_getcomponents

Response:

```
[
    {
        "id": "8b3b314c-bde3-44f6-ae4f-269b0aefac54",
        "name": "Full Scope Disabled",
        "providerId": "scope",
        "providerType": "org.keycloak.services.clientregistration.policy.ClientRegistrationPolicy",
        "parentId": "agent",
        "subType": "anonymous",
        "config": {}
    },
    {
        "id": "405cbfa7-c739-45dd-a986-fbc081638dfc",
        "name": "rsa-generated",
        "providerId": "rsa-generated",
        "providerType": "org.keycloak.keys.KeyProvider",
        "parentId": "agent",
        "config": {
            "priority": [
                "100"
            ]
        }
    },
    {
        "id": "ed5c36eb-10c3-44f1-a71b-69b2c0c45e5f",
        "name": "Trusted Hosts",
        "providerId": "trusted-hosts",
        "providerType": "org.keycloak.services.clientregistration.policy.ClientRegistrationPolicy",
        "parentId": "agent",
        "subType": "anonymous",
        "config": {
            "host-sending-registration-request-must-match": [
                "true"
            ],
            "client-uris-must-match": [
                "true"
            ]
[...]

PR #582, opened 2021-08-30
new resources: realm key aes, ecdsa, hmac, rsa, java keystore

Vlad-Kirichenko:
Hi @mrparkers,

This PR adds new resources to manage keystores. #569 
Tried creating everything in one resource but decided to separate them.
That's why:

- Easier data validation
- Fewer fields need to be filled when creating a resource
- The scheme for the config for the component is more complicated (also I faced the fact that if you pass empty fields to the keyloak, it will not update the resource, but it also does not throw an error)

Work is still in progress, here's what needs to be done:

- Add tests for new resources
- Fix resource import functions

Best regards,
Vlad

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -30,6 +30,12 @@ func KeycloakProvider(client *keycloak.KeycloakClient) *schema.Provider {
 		ResourcesMap: map[string]*schema.Resource{
 			"keycloak_realm":                                             resourceKeycloakRealm(),
 			"keycloak_realm_events":                                      resourceKeycloakRealmEvents(),
+			"keycloak_realm_key_aes_generated":                           resourceKeycloakRealmKeyAesGenerated(),
+			"keycloak_realm_key_ecdsa_generated":                         resourceKeycloakRealmKeyEcdsaGenerated(),
+			"keycloak_realm_key_hmac_generated":                          resourceKeycloakRealmKeyHmacGenerated(),
+			"keycloak_realm_key_java_keystore":                           resourceKeycloakRealmKeyJavaKeystore(),
+			"keycloak_realm_key_rsa":                                     resourceKeycloakRealmKeyRsa(),
+			"keycloak_realm_key_rsa_generated":                           resourceKeycloakRealmKeyRsaGenerated(),
 			"keycloak_required_action":                                   resourceKeycloakRequiredAction(),
 			"keycloak_group":                                             resourceKeycloakGroup(),
 			"keycloak_group_memberships":                                 resourceKeycloakGroupMemberships(),
--- a/keycloak/realm_key_aes_generated.go
+++ b/keycloak/realm_key_aes_generated.go
@@ -1,0 +1,116 @@
+package keycloak
+
+import (
+	"fmt"
+	"strconv"
+)
+
+type RealmKeyAesGenerated struct {
+	Id       string
+	Name     string
+	RealmId  string
+	ParentId string
+
+	Active     bool
+	Enabled    bool
+	Priority   int
+	SecretSize int
+}
+
+func convertFromRealmKeyAesGeneratedToComponent(realmKey *RealmKeyAesGenerated) *component {
+	componentConfig := map[string][]string{
+		"active": {
+			strconv.FormatBool(realmKey.Active),
+		},
+		"enabled": {
+			strconv.FormatBool(realmKey.Enabled),
+		},
+		"priority": {
+			strconv.Itoa(realmKey.Priority),
+		},
+		"secretSize": {
+			strconv.Itoa(realmKey.SecretSize),
+		},
+	}
+
+	return &component{
+		Id:           realmKey.Id,
+		Name:         realmKey.Name,
+		ParentId:     realmKey.ParentId,
+		ProviderId:   "aes-generated",
+		ProviderType: "org.keycloak.keys.KeyProvider",
+		Config:       componentConfig,
+	}
+}
+
+func convertFromComponentToRealmKeyAesGenerated(component *component, realmId string) (*RealmKeyAesGenerated, error) {
+	active, err := parseBoolAndTreatEmptyStringAsFalse(component.getConfig("active"))
+	if err != nil {
+		return nil, err
+	}
+
+	enabled, err := parseBoolAndTreatEmptyStringAsFalse(component.getConfig("enabled"))
+	if err != nil {
+		return nil, err
+	}
+
+	priority := 0 // Default priority
+	if component.getConfig("priority") != "" {
+		priority, err = strconv.Atoi(component.getConfig("priority"))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	secretSize := 16 // Default key size for aes key
+	if component.getConfig("secretSize") != "" {
+		secretSize, err = strconv.Atoi(component.getConfig("secretSize"))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	realmKey := &RealmKeyAesGenerated{
+		Id:       component.Id,
+		Name:     component.Name,
+		ParentId: component.ParentId,
+		RealmId:  realmId,
+
+		Active:     active,
+		Enabled:    enabled,
+		Priority:   priority,
+		SecretSize: secretSize,
+	}
+
+	return realmKey, nil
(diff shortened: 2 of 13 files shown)


mrparkers:
Thanks for the PR @Vlad-Kirichenko, this is looking good so far.  Let me know if you need help with anything.

Vlad-Kirichenko:
Hello @mrparkers 
I want to write a test for Java keystore. And for this I need to add an alias in the keystore inside instance with the keycloak.
Can you please advise the way to do this?

Vlad-Kirichenko:
That's all right. Then @mrparkers, what do you think if I will add step in `.circleci/config.yml` which will create java keystore and add alias into it(by using [keytool](https://docs.oracle.com/javase/7/docs/technotes/tools/solaris/keytool.html)) before `Run Tests` script?

mrparkers:
Yeah I have no problem with that at all.  If all this is doing is generating a file to mount to Keycloak's file system, you could also just generate it once and store it within `provider/misc`, which is where I put a cert / key for tests relating to SAML clients.  Or we could make a dedicated folder for test artifacts.

ringods:
Hello @mrparkers, joining the conversation here for a quick question. @Vlad-Kirichenko is one of my team members, and the work done here is contributed by the company we work for. As a technical lead, I was able to have an approved list of OpenSource projects we use internally, so we could contribute back to these projects. Executive management approved but they would like these contributions linked to their brand.

We would very much like to see a notion of this work logged somewhere, something like name of the company, hyperlinked to our website, and mentioning this PR and any later ones (e.g. #594).

Can you live with this? If so, where would be the best place to put this? The README.md file?

mrparkers:
I'm not opposed to something like that, but I'm not sure that the README is the appropriate place for this.

How about this - I'll make some changes to the [changelog](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/CHANGELOG.md) file, so that each release includes a "thank you" section that tags the individuals that contributed code towards that release.  Once I've done that, you can submit a PR to update one of these entries to include something like "@Vlad-Kirichenko on behalf of [Example Company](https://example-company.com)".  Does that sound like a good compromise?

mrparkers:
Okay, I've gone ahead and updated the [changelog](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/CHANGELOG.md) to include a list of each contributor under every release.  Once I've cut a release that includes these changes, feel free to make a PR to this file to include a reference to the company that these changes were contributed on behalf of.

Vlad-Kirichenko:
Hello @mrparkers.
This PR and #594 are ready to review.
Thank you!

keycloak/realm_keystore_rsa.go
```diff
@@ -0,0 +1,132 @@
...
+
+	priority := 0 // Default priority
+	if component.getConfig("priority") != "" {
+		priority, err = strconv.Atoi(component.getConfig("priority"))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	keySize := 2048 // Default key size for rsa key
```
mrparkers, line 79: we shouldn't have to specify a key size for a non generated key, right?

provider/provider.go
```diff
@@ -31,6 +31,12 @@ func KeycloakProvider(client *keycloak.KeycloakClient) *schema.Provider {
 		ResourcesMap: map[string]*schema.Resource{
 			"keycloak_realm":                                             resourceKeycloakRealm(),
 			"keycloak_realm_events":                                      resourceKeycloakRealmEvents(),
+			"keycloak_realm_key_aes_generated":                           resourceKeycloakRealmKeystoreAesGenerated(),
+			"keycloak_realm_key_ecdsa_generated":                         resourceKeycloakRealmKeystoreEcdsaGenerated(),
+			"keycloak_realm_key_hmac_generated":                          resourceKeycloakRealmKeystoreHmacGenerated(),
+			"keycloak_realm_key_java_keystore":                           resourceKeycloakRealmKeystoreJavaKeystore(),
+			"keycloak_realm_key_rsa":                                     resourceKeycloakRealmKeystoreRsa(),
+			"keycloak_realm_key_rsa_generated":                           resourceKeycloakRealmKeystoreRsaGenerated(),
```
mrparkers, line 39: what did you intend for these resources to be called?

for example, you use `keycloak_realm_key_rsa_generated` here and in the headers for the documentation, but the file names that were added are `realm_keystore_rsa_generated.md` and `resource_keycloak_realm_keystore_rsa_generated.go`.

to me, I think `keystore` makes more sense than `key`, since the Keycloak GUI refers to these as Keystores.  regardless, we'll want to make some changes so the resource names are consistent.

provider/resource_keycloak_realm_keystore_aes_generated.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = setRealmKeystoreAesGeneratedData(data, realmKey)
+	if err != nil {
+		return err
+	}
+
+	return keycloakClient.UpdateRealmKeystoreAesGenerated(realmKey)
```
mrparkers, line 151: is there a reason we're invoking `keycloakClient.UpdateRealmKeystoreAesGenerated` twice here.  all of these new resources appear to do this as well.

provider/resource_keycloak_realm_keystore_java_kyestore_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/mrparkers/terraform-provider-keycloak/keycloak"
+	"regexp"
+	"strconv"
+	"testing"
+)
+
+func TestAccKeycloakRealmKeystoreJava_basic(t *testing.T) {
+	t.Parallel()
+
+	skipIfEnvSet(t, "CI") // temporary while I figure out how to put java keystore file to keycloak container in CI
```
mrparkers, line 17: we could probably just mount the `misc` directory as a volume on the running container and reference it that way.  but I'm fine with leaving this as-is for now.

provider/resource_keycloak_realm_keystore_rsa.go
```diff
@@ -0,0 +1,221 @@
...
+				Description: "Priority for the provider",
+			},
+			"algorithm": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice(keycloakRealmKeystoreRsaAlgorithm, false),
+				Default:      "RS256",
+				Description:  "Intended algorithm for the key",
+			},
+			"key_size": {
```
mrparkers, line 62: left this comment earlier - but is `key_size` needed for non generated keys?

provider/resource_keycloak_realm_keystore_rsa.go
```diff
@@ -0,0 +1,221 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Private RSA Key encoded in PEM format",
+			},
+			"certificate": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "X509 Certificate encoded in PEM format",
+			},
+			"disable_read": {
```
mrparkers, line 79: we probably don't want to add attributes specifically for discarding configuration drift, there are other ways we can deal with that.

it looks like `private_key` comes back from keycloak as `**********` - so we'll probably want to catch this and not set this value in state when reading.  the state value should contain the private key submitted by the user.  this will allow us to detect changes to this attribute appropriately.

mrparkers (changes requested):
overall this looks really good, there's just a couple of issues I found

provider/resource_keycloak_realm_keystore_java_kyestore_test.go
Vlad-Kirichenko: Yes, I tried to to this(using the `misc` dir) but the problem that I need keystore file inside keycloak instance, not in runner instance. 

In the terraform manifest inside `keystore` field I paste path for keycloak itself.

StuxxNet:
@mrparkers when do you plan to release a new version with this resource available? :D

ringods:
@StuxxNet it seems @mrparkers merged our latest PR and created that release: v3.5.0 😄

meckhardt:
Thanks @Vlad-Kirichenko

PR #585, opened 2021-09-05
feat: Support for configuring authentication OTP policy (#394)

### LINKED ISSUE #394: Support for configuring authentication OTP policy
Hi, 

i have not come across means to change the "OTP Policy" in the authentication section of keycloak. 
Default provided is SHA1 but would like to configure SHA256 for my terraform provided realms.

nolte:
Adding support for configuring authentication OTP policy of some Realm, will be fixed #394

### DIFF
--- a/provider/resource_keycloak_realm.go
+++ b/provider/resource_keycloak_realm.go
@@ -7,6 +7,56 @@ import (
 )
 
 func resourceKeycloakRealm() *schema.Resource {
+
+	otpPolicySchema := map[string]*schema.Schema{
+		"type": {
+			Type:         schema.TypeString,
+			Description:  "OTP Type, totp for Time-Based One Time Password or hotp for counter base one time password",
+			Optional:     true,
+			Default:      "totp",
+			ValidateFunc: validation.StringInSlice([]string{"totp", "hotp"}, false),
+		},
+		"algorithm": {
+			Type:         schema.TypeString,
+			Description:  "What hashing algorithm should be used to generate the OTP.",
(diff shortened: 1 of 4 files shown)


provider/resource_keycloak_realm.go
```diff
@@ -7,6 +7,56 @@ import (
...
+		},
+		"digits": {
+			Type: schema.TypeInt,
+			Elem: &schema.Schema{
+				Type: schema.TypeInt,
+			},
+			Default:  6,
+			Optional: true,
+		},
+		"initial_counter": {
```
mrparkers, line 34: I don't actually see the setting for this in the latest three Keycloak versions.  Am I missing something?

mrparkers (commented):
Thanks for the PR, overall this LGTM, just had one question

provider/resource_keycloak_realm.go
nolte: The Option is only visible if you use a  `OTP Type:` _Counter Based_  , for the Default Type, _Time Based_, you will got the field `OTP Token Period` ( [period](https://github.com/mrparkers/terraform-provider-keycloak/blob/efa42c17c5b5e549861dfc5a6983b2f4672ed868/provider/resource_keycloak_realm.go#L50)).

mrparkers: Perfect, I just missed that then.  Thanks for your help

mrparkers (approved):
LGTM 🚀

PR #587, opened 2021-09-07
feat: add data source keycloak_user_roles

f-stibane:


f-stibane:
I need help with the failing build please; seems unrelated to my changes.

```
Failed
    testing_new.go:70: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: error sending GET request to /auth/admin/realms/tf-acc-4991586546742877040/client-scopes: 500 Internal Server Error. Response body: {"error":"unknown_error"}
```

mrparkers:
Yeah, sometimes the tests can get flaky with Keycloak throwing random 500s here and there.  I restarted the failed test, you should be good to go after that.

Everything looks good, but what do you think about renaming this data source to `keycloak_user_roles`, and having two computed attributes, one for `realm_role_names` and the other for `client_role_names`?

f-stibane:
Thanks for the fast feedback!

I'd be happy to add that, but I'm unfamiliar with `client_roles` - can you provide an example of how to create one?

f-stibane:
As requested, I renamed it to `keycloak_user_roles` with field `realm_role_names`.

Could we merge this and leave the `client_role_names` open for another PR later on? At least it's easily extensible now :wink:

f-stibane:
Also, what's up with the builds? They seem stuck (or not reporting back).

mrparkers:
Apologies @f-stibane, but after looking at it again, I actually think it would be best to revert back to your original proposal.

My original thinking was that we could include a `client_roles` attribute that was something like a `map[string][]string`, where each key would be the client's ID, and each value would be a slice of role IDs.  but after looking at it again, it looks like maps in terraform have to have primitive values, so they can't be slices.

So if you could, would you be able to undo that last change?  Also, I think the reason the builds aren't kicking off is because you're working off of the `master` branch in your fork.  If you use a non `master` branch, the builds should work.

f-stibane:
Closed in favor of https://github.com/mrparkers/terraform-provider-keycloak/pull/596 - changed branch name of fork.

f-stibane:
No problem! Reset to original suggestion.

f-stibane:
Also, I found out why the builds were not running (or at least reporting back) - I revoked circleci's access to my repositories, because I don't use it for my projects :grimacing:

PR #590, opened 2021-09-08
feat: Adding support for login theme on saml_client resource

cw1o:
I've been using the provider for a little while and noticed the SAML Client doesn't support setting the login theme via TF. This should enable that. If there's anything that needs tweaking let me know.

### DIFF
--- a/docs/resources/saml_client.md
+++ b/docs/resources/saml_client.md
@@ -38,6 +38,7 @@ resource "keycloak_saml_client" "saml_client" {
 - `name` - (Optional) The display name of this client in the GUI.
 - `enabled` - (Optional) When false, this client will not be able to initiate a login or obtain access tokens. Defaults to `true`.
 - `description` - (Optional) The description of this client in the GUI.
+- `login_theme` - (Optional) The login theme of this client.
 - `include_authn_statement` - (Optional) When `true`, an `AuthnStatement` will be included in the SAML response.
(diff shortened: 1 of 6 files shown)


mrparkers (changes requested):
Hey @cw1o, thanks for the PR.  This looks good, but we're missing calls to `data.Get` and `data.Set`, which is responsible for retrieving the value from your HCL configuration and persisting it back to state, respectively.

You'll want to modify these two functions to add these lines of code:
- [`mapToSamlClientFromData`](https://github.com/mrparkers/terraform-provider-keycloak/blob/cb74346df6c3e4bb4e2696824d60b65932a1bea3/provider/resource_keycloak_saml_client.go#L218) - we should call `data.Get` within this function to help build the `keycloak.SamlClient` to send off to Keycloak
- [`mapToDataFromSamlClient`](https://github.com/mrparkers/terraform-provider-keycloak/blob/cb74346df6c3e4bb4e2696824d60b65932a1bea3/provider/resource_keycloak_saml_client.go#L317) - we should call `data.Set` within this function to persist the value from Keycloak back to Terraform state.

cw1o:
I've updated both of them functions now, tests look okay. That 11.0.3 one might need rerunning don't know what happened there.

mrparkers (approved):
LGTM, thanks for the PR!

Issue #324, opened 2020-06-16
Default (Client) Roles


jcharlytown:
Hi,

first of all thanks everyone for making this provider happen. This makes managing keycloak realms so much easier!

However, I was wondering if there is a functionality to set default roles on a realm, specifically client roles? Using the security console this is possible via "Roles" -> "Default Roles". In case of client roles, on the API level this seems to be a field within the client representation, similar to default client scopes.

Is there such a functionality already included in this project and i'm just missing it? If not, are there any plans to implement such a feature?

Thanks!

paulvollmer:
hello,
someone working on that feature? if not, then i can work on that.

/cc @mrparkers

mrparkers:
@paulvollmer feel free to pick it up if you want to work on it, I appreciate the help!

duc12597:
+1 for default realm roles

paulvollmer:
@mrparkers  i started working on that.

ringods:
Hello @paulvollmer, were you able to make any progress on this? I'm interested in this feature as well.

ringods:
Hello @arkadiusjonczek, were you able to make some progress on this? Even if you have something incomplete, could you already submit this as a draft PR?

Vlad-Kirichenko:
Hi all.
I started working on this(realm default roles). PR will be created soon.

PR #599, opened 2021-09-21
Feature `default_roles` in `keycloak_realm` resource & data source

Vlad-Kirichenko:
Hello @mrparkers.

Changes in this PR gonna implement feature mentioned in #324.(Realm level roles assigned to new users)

Best regards,
Vlad

### DIFF
--- a/docs/resources/realm.md
+++ b/docs/resources/realm.md
@@ -73,6 +73,8 @@ resource "keycloak_realm" "realm" {
     relying_party_id          = "keycloak.example.com"
     signature_algorithms      = ["ES256", "RS256"]
   }
+
+  default_roles = [ "uma_authorization" ]
 }
 ```
 
@@ -199,7 +201,7 @@ The arguments below can be used to configure authentication flow bindings:
 
 ### OTP Policy
 
-The `otp_policy` block with following arguments can be found in the "OTP Policy" tab within the realm settings. 
+The `otp_policy` block with following arguments can be found in the "OTP Policy" tab within the realm settings.
 
 - `type` - (Optional) One Time Password Type, supported Values are `totp` for Time-Based One Time Password and `hotp` for Counter Based. Defaults to `totp`.
 - `algorithm` - (Optional) What hashing algorithm should be used to generate the OTP, Valid options are `HmacSHA1`,`HmacSHA256` and `HmacSHA512`. Defaults to `HmacSHA1`.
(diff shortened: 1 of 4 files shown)


keycloak/realm.go
```diff
@@ -131,6 +131,9 @@ type Realm struct {
 	WebAuthnPolicyPasswordlessRpId                            string   `json:"webAuthnPolicyPasswordlessRpId"`
 	WebAuthnPolicyPasswordlessSignatureAlgorithms             []string `json:"webAuthnPolicyPasswordlessSignatureAlgorithms"`
 	WebAuthnPolicyPasswordlessUserVerificationRequirement     string   `json:"webAuthnPolicyPasswordlessUserVerificationRequirement"`
+
+	// Roles
+	DefaultRoles []string `json:"defaultRoles"`
```
mrparkers, line 136: This probably isn't the API we'd want to use to accomplish this.  It looks like `defaultRoles` is only honored on realm creation / update, but it isn't returned when reading.  This is going to result in a perpetual diff for this attribute.

It looks like default realm roles are handled as a composite role, so adding a new default realm role is just modifying the realm's default role to add a new composite role.

When you `GET` a realm, there's a property in the JSON that looks like this:

```json
"defaultRole": {
  "id": "bb6b80f2-0f34-4135-93ca-b3b47f8b7fd4",
  "name": "default-roles-master",
  "description": "${role_default-roles}",
  "composite": true,
  "clientRole": false,
  "containerId": "test"
}
```

So we're going to want to add / remove composites to this role in order to correctly influence the realm's default roles.  You can refer to the `keycloak_role` resource, which also uses composites, for some guidance on how to do this.

provider/resource_keycloak_realm_test.go
```diff
@@ -1624,3 +1716,16 @@ resource "keycloak_realm" "realm" {
...
+
+func testKeycloakRealm_default_roles(realm string, defaultRoles []string) string {
+
+	defaultRolesString := fmt.Sprintf("%s", arrayOfStringsForTerraformResource(defaultRoles))
+
+	return fmt.Sprintf(`
+resource "keycloak_realm" "realm" {
+	realm                  = "%s"
+	enabled                = true
+	default_default_roles  = %s
```
mrparkers, line 1728:
```suggestion
	default_roles  = %s
```

docs/resources/realm.md
```diff
@@ -238,6 +240,9 @@ Each of these attributes are blocks with the following attributes:
 - `default_default_client_scopes` - (Optional) A list of default default client scopes to be used for client definitions. Defaults to `[]` or keycloak's built-in default default client-scopes.
 - `default_optional_client_scopes` - (Optional) A list of default optional client scopes to be used for client definitions. Defaults to `[]` or keycloak's built-in default optional client-scopes.
 
+## Roles
+- `default_roles` - (Optional) A realm level roles assigned to new users. Defaults to `[]` or keycloak's built-in default roles.
```
mrparkers, line 244: maybe this attribute should be clear that it's for realm roles

```suggestion
- `default_realm_roles` - (Optional) Realm level roles assigned to new users by default. Defaults to `[]` or keycloak's built-in default roles.
```

mrparkers (changes requested):

StianHaugland1:
Would be nice to have default client roles as well 😄

keycloak/realm.go
Vlad-Kirichenko: Yes, my bad. I made it for version 12 and tested it on it... and it worked). But since version 13 api has changed for default_roles.

Now i made changes for 13 and above.

Issue #423, opened 2020-11-02
[ldap - ad] missing mappers ? what's the roadmap ?


semangard:
Hello,

It seems there are some missing mappers on the LDAP-AD user storage federation (cf. screenshot below).

Especially the **ldap-role-mapper** is really missing.
Do you plan to implement this quite soon  ?

![image](https://user-images.githubusercontent.com/57065741/97917394-b3724180-1d54-11eb-8e09-d710fe26ffda.png)

angeloxx:
The opened branch implements the hardcoded-attribute-mapper.

PR #602, opened 2021-09-29
Add LDAP hardcoded attribute mapper

francois-travais:
Relates to #423


Useurmind:
Will this be integrated?
This would be especially useful to auto verify email addresses in combination with oauth2-proxy.

PR #603, opened 2021-09-29
Changing config for identity_provider_mapper

### LINKED ISSUE #571: Unable to configure oidc-role-idp-mapper using keycloak_custom_identity_provider_mapper
Hi,

When I try to setup "Claim to Role" mapper for oidc identity provider, terraform ignores Claim, ClaimValue and Role parameters specified in "extra_config"

I suspect this is related to the same issue mentioned in https://github.com/mrparkers/terraform-provider-keycloak/issues/546 and https://github.com/mrparkers/terraform-provider-keycloak/issues/520
The only parameter that is actually applied is "syncMode"

This is the actual terraform config I was trying:
```
resource "keycloak_custom_identity_provider_mapper" "master_azure_ad_test" {
  realm                    = keycloak_realm.master.id
  name                     = "test"
  identity_provider_alias  = keycloak_oidc_identity_provider.master_azure_ad.alias
  identity_provider_mapper = "oidc-role-idp-mapper"

  // none of these creates claim to role mapping

  extra_config = {
    Claim      = "groups"
    ClaimValue = "60da9df6-5bd9-461f-a637-945be8c596c6"
    Role       = "admin"
    syncMode   = "INHERIT"
  }

  //  extra_config = {
  //    claim = "groups"
  //    "claim.value" = "60da9df6-5bd9-461f-a637-945be8c596c6"
  //    role = "admin"
  //    syncMode = "INHERIT"
  //  }

}
```

This is how mapper looks in admin console
image

Am I doing something wrong (besides using "master" realm) or is this an issue with keycloak provider?
Provider version: v3.2.1

aceresia:
I was having the same issue as in the issue below.
Fixes #571 

I don't really know go, but this seems to work for me.  To work with this branch you will need to move all of the config items into the "config" block, and leave the extra config items in the "extra_config" block.

Config items are the ones listed [here](https://github.com/mrparkers/terraform-provider-keycloak/blob/f29d1f3b932de0479354bd19ead640507095dfe6/keycloak/identity_provider_mapper.go#L9)

### DIFF
--- a/provider/generic_keycloak_identity_provider_mapper.go
+++ b/provider/generic_keycloak_identity_provider_mapper.go
@@ -36,6 +36,10 @@ func resourceKeycloakIdentityProviderMapper() *schema.Resource {
 				ForceNew:    true,
 				Description: "IDP Alias",
 			},
+			"config": {
+				Type:     schema.TypeMap,
+				Optional: true,
+			},
 			"extra_config": {
 				Type:     schema.TypeMap,
 				Optional: true,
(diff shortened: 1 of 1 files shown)


mrparkers:
Hey @aceresia, I appreciate the work you put into this, but I ended up going with a different approach in #607 so we wouldn't have to add a new attribute to the resource.  So I'm going to close this in favor of #607.

PR #627, opened 2021-11-11
Enable custom client_authenticator_type

### LINKED ISSUE #396: JWT credentials
I believe this is a feature request.  I've got an openid client configured, but I don't see a way to set up JWT credentials.

```
resource "keycloak_openid_client" "openid_client" {
  realm_id            = data.keycloak_realm.realm.id
  client_id           = module.cluster_elb.load_balancer_dns
  name                = "${local.cluster_name}-${local.environment}"
  enabled             = true

  standard_flow_enabled = true
  direct_access_grants_enabled = true

  access_type         = "CONFIDENTIAL"
  valid_redirect_uris = [
    "https://${module.cluster_elb.load_balancer_dns}:8443/*",
  ]
  web_origins = [
    module.cluster_elb.load_balancer_dns
  ]
}
```

This screenshot shows the setting I'm looking for on the UI.

Screenshot from 2020-10-09 13-33-25

I looked at the `keycloak_oidc_identity_provider`, but that has additional required fields.

oysteinhauan:
Fixes #396 
Kept the hardcoded clientAuthenticatorType for backwards compatibility, while enabling it to be specified with the possible values ["client-secret", "client-secret-jwt", "client-x509", "client-jwt"].


provider/resource_keycloak_openid_client.go
```diff
@@ -64,6 +64,11 @@ func resourceKeycloakOpenidClient() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+			"client_authenticator_type": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				Computed:  true,
```
mrparkers, line 70: We should set the default value here instead of checking for an empty string and setting it within the `keycloak` package.

mrparkers (changes requested):
Hey @oysteinhauan, sorry for the delay on this.  I just had one comment before this can get merged.  Thanks!

provider/resource_keycloak_openid_client.go
oysteinhauan: Hey, sorry, didn't get notified of this reply. 
Fixed it now!

oysteinhauan:
@mrparkers is this ready to be merged now? :)

mrparkers (approved):
Sorry again for the delay. I went ahead and pushed some fixes to your branch to speed this up for you (format changes and a quick fix to the test you added).  Thanks again for the PR!  I'll cut a release that includes these changes within a few days.

mrparkers (approved):
Sorry again for the delay. I went ahead and pushed some fixes to your branch to speed this up for you (format changes and a quick fix to the test you added).  Thanks again for the PR!  I'll cut a release that includes these changes within a few days.

mrparkers:
Not sure why that approval posted twice 😄

oysteinhauan:
Wonderful! Thanks! Looking forward to the release - we've been using a local version of the provider and it seems to work great :)

PR #644, opened 2022-01-20
feat: Add support for frontchannel logout url. 

oysteinhauan:
Frontchannel logout was added in Keycloak v15. 

Also bumped github.com/hashicorp/terraform-plugin-sdk/v2  v2.7.0 -> v2.10.1


oysteinhauan:
Not sure why the 13.0.1 acceptance test fails. 
Could be because of the new terraform-plugin-sdk version.
But since Keycloak version 16 was released some time ago, version 13 may not need official support anymore?

mrparkers (approved):
Thanks for the PR!  The failing test was just flaky, it was fine after I ran it again and merged `master` with your branch.

PR #649, opened 2022-01-31
feat: add default client scope management on realm

Korsarro69:
My project need to apply created client-scopes as default to automatically assign the client-scope to each newly created client


mrparkers:
Hey @Korsarro69, thanks for submitting this PR.  I'm just curious if you have tried the [`keycloak_openid_client_default_scopes`](https://registry.terraform.io/providers/mrparkers/keycloak/latest/docs/resources/openid_client_default_scopes) resource for this use case, or if there was a reason why that didn't work for you.

I don't mind adding a resource for a single client scope, but we would probably want to make it clear in the docs why you might want to use one over the other.

Korsarro69:
Hello, I'm sorry for the very late response. I needed to assign a client scope in default client scopes, but without link it to a client.
The resource [keycloak_openid_client_default_scopes](https://registry.terraform.io/providers/mrparkers/keycloak/latest/docs/resources/openid_client_default_scopes) required a client

lebronnecf:
Hi, we are also interested in this feature. This is not linked to a specific client, but is a realm level configuration. It has its own endpoint in the Keycloak api (that is used in this PR).

daviddelannoy:
Hi @mrparkers 
I agree, these options are not currently supported by the provider and is fixed by this PR, isn't it ? 

image

dglozano:
>I'm just curious if you have tried the [keycloak_openid_client_default_scopes](https://registry.terraform.io/providers/mrparkers/keycloak/latest/docs/resources/openid_client_default_scopes) resource for this use case, or if there was a reason why that didn't work for you.

Hi @mrparkers 👋  I ran into the same problem as @Korsarro69 
The existing `keycloak_openid_client_default_scopes` is for setting up the default scope for a specific client.
What this PR adds support for is for setting the _"Relam's Default Default Client Scopes"_, which is a client scope that will be added as default scope to all newly created clients in the Realm.

There is the possibility to set the "Default Default Client Scopes" already in the realm level as far as I understand, but that doesn't work for custom created scopes, as there is a chicken-and-egg problem since the client scope requires the realm to exist from before (ref.: https://github.com/mrparkers/terraform-provider-keycloak/issues/470#issuecomment-1983134078 ). Therefore I think it makes sense to introduce a new resource so set this up.

aslafy-z:
This PR was superseded by https://github.com/keycloak/terraform-provider-keycloak/pull/1079

sschu:
@aslafy-z Thanks for noting, I will go ahead and close this PR.

PR #653, opened 2022-02-10
feat: add support for realm client_session_idle_timeout and client_session_max_lifespan

### LINKED ISSUE #642: client_session_idle_timeout & client_session_max_lifespan on realm level
Hey, 

is there a way to configure these two properties for a realm? 

See here https://www.keycloak.org/docs/latest/server_admin/#_timeouts as well (Client Session Idle, Client Session Max). 

It looks like both properties are missing? Any intention on supporting these properties?

fapian:
closes #642


mrparkers (approved):
Looks great, sorry for the delay in getting this merged.  I'll be cutting a release today that will include this change.

Thanks!  🚀

PR #656, opened 2022-02-16
Support custom saml provider idp

tomrutsaert:
This allows to set a custom provider_id on the saml idp resource. This is similar as on the oidc idp resource.


mrparkers (approved):
thanks for the PR @tomrutsaert, this looks fine to me.  would you like to update the docs for this new attribute too?  if not I can get to it later.

tomrutsaert:
I will add the docs

PR #657, opened 2022-02-17
Update openid_client.md with additional info on web_origins

marwol-fdir:
Although not well documented there is a small but important caveat with the use of '+' for web_origins.
'+' will permit all origins in `valid_redirect_uris` except '*'.
So without this additional info it is easy to make the mistake in believing that, when `valid_redirect_uris` contains '*', including '+' in `web_origins` will cause '*" to be included, which it is not.
The only reference in official Keycloak sources I could find comes from the source code of the admin-ui found here: https://github.com/keycloak/keycloak-admin-ui/blob/5fd344a7bd76bd7f2d3e91794c6fbf738e917bd8/src/clients/help.ts#L44


marwol-fdir:
Sorry for spamming with the re-request. Had to update with latest changes from `master`.

mrparkers:
No problem at all, thanks again!

PR #658, opened 2022-02-21
feat: add `keycloak_realm_user_profile` resource

maximepiton:
Hello there,

This PR adds a User Profile resource to the provider (as my team need it).

As the User Profile is still a preview feature, and is managed through a specific endpoint, I thought that it was better to set it as a separate resource and not set user profile settings in the keycloak_realm resource directly.

Some tests unrelated to this feature (keystores, federation) fail when I run them locally, but I think it's a flakiness issue..?

Let me know what you think!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,7 +94,7 @@ jobs:
           KEYCLOAK_LOGLEVEL: INFO
           KEYCLOAK_USER: keycloak
           KEYCLOAK_PASSWORD: password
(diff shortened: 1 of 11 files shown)


maximepiton:
To have the tests running successfully in CI, the `-Dkeycloak.profile.feature.declarative_user_profile=enabled` flag should be added in `.github/workflows/test.yml:97`, but I apparently don't have the rights to do so (which makes sense)

PR #661, opened 2022-02-24
Support for NameIdPolicyFormat Transient on the saml idp

tomrutsaert:
Added support for NameIdPolicyFormat Transient on the saml idp provider resource


mrparkers (approved):
LGTM 🚀

PR #668, opened 2022-03-12
Adding resource_authentication_bindings 

### LINKED ISSUE #298: Assign realm default flow not possible
Within the current design it's not possible to modify the realm flows caused by a __cyclic dependency__ between __keycloak_realm__ and __keycloak_authentication_flow__:
* The creation of a flow requires a realm to be present.
* The default flow of the realm must be provided during realm creation.

```
  realm                     = "my realm"
  browser_flow              = "my flow"
}

resource "keycloak_authentication_flow" "flow" {
  realm_id          = keycloak_realm.realm.id
  alias             = "my flow"
}
```

For a newly created realm, the presence of the __browser_flow__ value will lead to a 500 error. Patching an existing realm (by modifying the terraform and reapply) will work.

I would assume it will require an additional resource type that will take an existing realm_id and flow_id to link them.

~~The current workaround is to apply the flow to all applications of a realm separat.~~

camjjack:
Enable setting custom authentication flows to a realm created in the same run.

Closes #298

Credit to [bruno-delfino1995](https://github.com/bruno-delfino1995) code referenced https://github.com/mrparkers/terraform-provider-keycloak/issues/298#issuecomment-885753951


mrparkers:
Hey @camjjack, thanks for this PR.  I ended up pushing a fix that's related to some changes I made earlier this week, so this should be in a mergable state now.

I was testing this locally and it looks like this resource is unfortunately fighting with some of the attributes within the `keycloak_realm` resource.  For example, if you don't specify the `browser_flow` for the `keycloak_realm` resource and you use this resource to specify it instead, these two resources will fight with each other on each run of `terraform apply`.

I can continue to mess with this when I have some free time to find a solution.  I think we'll have to update the `Realm` struct to turn those attributes into string pointers, so they aren't sent to Keycloak unless they're specified by the user.

msvechla:
Any update on this? It looks like the fix with the string pointers has been applied. I would also be open to jump in if something still needs some work

mrparkers:
Unfortunately, it isn't quite done yet.  I had a bit of time last month to push a few commits but I've been swamped lately and haven't had time since to revisit this.

When I left it, there was still some issues with the expected behavior for removing authentication flow assignment via this resource or the `keycloak_realm` resource.  I might be able to put another few hours into this over the weekend so I could hopefully close this out and cut a release.

mrparkers:
@camjjack @msvechla I've gone ahead and made the required changes to get this in a mergable state.  Thanks again @camjjack for all of your hard work on this!

mrparkers (approved):

Issue #686, opened 2022-05-11
keycloak_authentication_bindings: Wrong API-URLs are generated if technical realm id is different to realm name


dmeyerholt:
Hi,
while evaluating the new keycloak_authentication_bindings resource and thus updating the provider from 3.7.0 to 3.8.1 we encountered an error which is related to the old topic of handling realm internal IDs and the realms name (also see my old PR #270).

We try to create a keycloak_authentication_bindings resource  on a realm which has an "internal ID" like this: "My Realm ID" (this is the ID field in the JSON response of keycloak) but an realm's name is "myrealm" (this is the "realm" field in the response). 
Now an API-call to keycloak is built with URLs like /realm/My Realm ID instead of /realm/myrealm . This yields to 404s.

dmeyerholt:
Ok, while digging into it, the referenced PR #675 changed the references from realm.realm to realm.id in the examples and/or documentation so while this is not good because of the mentioned possible clash between "internal id" and "realm" the observed behaviour should be solved be just changing this line https://github.com/mrparkers/terraform-provider-keycloak/blob/87447fef93084fc203d508a8d0360b351656069e/provider/resource_keycloak_authentication_bindings.go#L101 and replacing realm.id with realm.realm . 
I have noticed, that the authentication binding was correctly set on the realm but only the keycloakClient.GetRealm needs to be changed so the read after update of the realm should work again. Will try to setup a local built of the provider to verify.

dmeyerholt:
Changing that single line fixes the issue in PR #687 . Furthermore I checked compatibility with our very extensive state and found no other issues regarding the realm.Id and realm.Realm handling. 
Nonetheless I think the documentation should be reverted to referencing the realm-field of a realm-resource and not the id-field because of the difference of the id-field when the technical id differs from the realm's realm name.

mrparkers:
Thanks for opening this issue.  The way you're describing this makes sense, and I appreciate that you opened a PR for this.  My apologies for taking so long to get around to this.

PR #687, opened 2022-05-11
Fix authentication bindings when internalId differs from realm name

dmeyerholt:
Fix for #686


mrparkers (approved):
LGTM, thanks for this PR!

PR #697, opened 2022-06-21
Changed validators to schema to support options

surfdudester:
The current implementation does not support option validator, since Keycloak requires it to be an array of strings. We couldn't find an easy way to implement it with the current schema, we had to add a specific schema for every validator that Keycloak supports. We added tests to cover this new schema.

This is a breaking change in the schema and should probably be in a major version. Let us know what you think.

@maximepiton is the original contributor for this feature, please review if possible

### DIFF
--- a/docs/resources/realm_user_profile.md
+++ b/docs/resources/realm_user_profile.md
@@ -45,14 +45,15 @@ resource "keycloak_realm_user_profile" "userprofile" {
     }
 
     validator {
-      name = "person-name-prohibited-characters"
-    }
-
-    validator {
-      name   = "pattern"
-      config = {
+      person_name_prohibited_characters {}
+      pattern {
         pattern       = "^[a-z]+$"
         error_message = "Nope"
+      }
+      length {
+        min           = 1
+        max           = 10
+        trim_disabled = false
       }
     }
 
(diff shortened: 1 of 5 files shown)


maximepiton:
Hi @surfdudester,

Good catch, I obviously missed this one!
Unfortunately, adding specific config schemas for each validator prevent using custom ones, that's the major drawback I see about your solution...

#705 also fixes the issue about having an array in the config, but allows having custom validators, so I would go for this one instead. (My company uses custom validators, using your solution would force us to maintain a fork of the provider only for this)

Cheers!

mrparkers:
Hi @surfdudester, thanks for opening this PR.  I'm not really familiar with the user profile feature of Keycloak, so I spent some time playing around with it to see what I think about this API.

Normally, I think I'd agree with you that I prefer well-defined APIs for this provider to expose to the user for configuring Keycloak like this.  However, when the underlying Keycloak API that we're interacting with isn't well-defined or structured, it ends up making the code pretty tough to maintain going forward.

I also think that the point that @maximepiton raised is valid here - folks who implement the User Profile SPI to add custom validators will need to be able to use raw JSON with this provider in order to configure those.  So the current functionality that's already in the provider would need to remain in place anyways.

While I appreciate the time and effort you went through to put this together, I'm going to close it in favor of #705.  I'll work on testing that out and ensuring that each validator that's bundled with Keycloak works as it should.

Thanks again to everyone in this thread for chiming in with your use case, it really helps me better understand these Keycloak features that I don't use myself.

PR #705, opened 2022-07-07
Declarative User Profile : Annotations & validator config as Json

jermarchand:
Hello,

In "Declarative User Profile", Annotations and Validators are not just `map[string]string`, it should be `map[string]interface{}`

In Terraform, an attribut (in a map) for a resource cannot be String or Object in the same time (Generics missing ;))

With this PR, the workaround I propose is to serialize the object as a Json string.

Hope this will help

### DIFF
--- a/keycloak/realm_user_profile.go
+++ b/keycloak/realm_user_profile.go
@@ -23,7 +23,7 @@ type RealmUserProfileSelector struct {
 type RealmUserProfileValidationConfig map[string]interface{}
 
 type RealmUserProfileAttribute struct {
-	Annotations map[string]string                           `json:"annotations,omitempty"`
+	Annotations map[string]interface{}                      `json:"annotations,omitempty"`
 	DisplayName string                                      `json:"displayName,omitempty"`
 	Group       string                                      `json:"group,omitempty"`
 	Name        string                                      `json:"name"`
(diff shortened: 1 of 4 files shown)


keycloak/realm_user_profile.go
```diff
@@ -65,5 +65,50 @@ func (keycloakClient *KeycloakClient) GetRealmUserProfile(ctx context.Context, r
 		return nil, err
 	}
 
+	for _, attr := range realmUserProfile.Attributes {
+		if attr.Validations != nil {
+			for name, config := range attr.Validations {
+
+				c := make(map[string]interface{})
+				for k, v := range config {
+					if _, ok := v.([]interface{}); ok {
```
mrparkers, line 74: I know that the validators that come bundled with Keycloak either have string or array keys (which is what we're checking for here), but do you think it would also be useful to handle nested objects?  You mention that you have custom validators, do any of them use nested objects?

mrparkers (commented):
Also, when I use an object as a key for an annotation, like this:

```hcl
    annotations = {
      bar = jsonencode({
        "hello" = "world"
      })
    }
```

it ends up looking like this on the keycloak side:

image

That doesn't seem right to me, is this something you'd expect?

daviddelannoy:
> Also, when I use an object as a key for an annotation, like this:
> 
> ```terraform
>     annotations = {
>       bar = jsonencode({
>         "hello" = "world"
>       })
>     }
> ```
> 
> it ends up looking like this on the keycloak side:
> 
> image
> 
> That doesn't seem right to me, is this something you'd expect?

Hi @mrparkers 
It seems OK to me, it is just a display limitation on Keycloak side, but we see it's an object and then can go to JSON editor to get more details ?

nickej:
Sorry for nagging but is there any progress on this one? We need to be able to create a profile attribute which have a list as validator configuration and an object as annotation. It is for a select input with given options.

Anything we can do to help out?

Thanks

mrparkers (approved):
Thanks again for this PR, sorry for taking so long to get it merged.

PR #749, opened 2022-10-10
Update docs regarding OpenID Connect Logout

jjojo:
Keycloak v18+ implements OpenID Connect Logout (https://www.keycloak.org/2022/04/keycloak-1800-released) and it requires post_logout_redirect_uri to be set to properly sign out. This terraform provider can handle this by utilizing the "extra_config" option.


earnubs:
Using extra_config for this appears to fail during an upgrade:

```
ERROR: Migration failed for change set META-INF/jpa-changelog-19.0.0.xml::19.0.0-10135::keycloak:\n     Reason: liquibase.exception.DatabaseException: ERROR: duplicate key value violates unique constraint \"constraint_3c\"\n  Detail: Key (client_id, name)=(18e9c7c6-2224-4a05-9a90-417ff499daca, post.logout.redirect.uris) already exists. [Failed SQL: (0) INSERT INTO public.client_attributes (CLIENT_ID, NAME, VALUE) VALUES ('18e9c7c6-2224-4a05-9a90-417ff499daca', 'post.logout.redirect.uris', '+')]
```

mrparkers:
Thanks for opening this PR, but I decided to add top-level support for this attribute instead in #777.

PR #750, opened 2022-10-11
Added totp option for keyclock login

HK78-HAHA:
If grant_type is password, this is a Pull Request that reads the `export KEYCLOAK_TOTP` environment variable and changed it so that Terraform can handle it as optional.

The background of this Pull Request started with the acquisition of "User Audit Data".

Strict requirements exist in Korea's security regulations (K-ISMS). In particular, most of the records of user behavior and 2-factor authentication are required.

If grant_type is "client_credentials", it is difficult to know which user did which action. We recommend "client_credentials" in this provider, but we can't use it for the reasons mentioned above.

Also, for "password" type authentication, totp authentication is not supported in the current provider version.

Therefore, during "password" authentication, the totp environment variable is read from Terraform and changed so that it can be selectively processed.


HK78-HAHA:
@mrparkers any news on this pull request?

lens0021:
ping @mrparkers

lens0021:
Rebased.

PR #760, opened 2022-11-04
FEAT: Allow multivalue for Config attibutes

ChrisHubinger:


ChrisHubinger:
Accidentially created

Issue #771, opened 2022-11-14
Deleting one keycloak_generic_role_mapper for a client deletes all of them


rd-robert-avram:
### Version
`mrparkers/keycloak 4.0.0` and `mrparkers/keycloak 4.0.1`

### Issue
When a client has more than 1 `keycloak_generic_role_mapper` defined for it, deleting one will lead to the provider deleting **all** of the client's `keycloak_generic_role_mapper`s.
This basically means that the provider will delete a clients entire scope when trying to delete just one entry within it.

### Root cause
We've debugged a little and found out that upon deletion, the provider sends a **bodyless** `DELETE` request to the endpoint
`/admin/realms/<realm>/clients/<client-uuid>/scope-mappings/realm`.

The Keycloak API expects the deleted resource (in the case of a role, the role itself) to be provided in the body.
Without it, Keycloak basically deletes the entire set of scopes for that client.

### Steps to reproduce

1. Define some realm roles and a client, then add those realm roles to its scope:
```terraform
resource "keycloak_role" "realm-role-1" {
realm_id = "my-realm"
name = "realm-role-1"
}

resource "keycloak_role" "realm-role-2" {
realm_id = "my-realm"
name = "realm-role-2"
}

resource "keycloak_role" "realm-role-3" {
realm_id = "my-realm"
name = "realm-role-3"
}

resource "keycloak_openid_client" "my-client" {
  access_type                         = "CONFIDENTIAL"
  realm_id                            = "my-realm"
  client_id                           = "my-client"
  enabled                             = true
  full_scope_allowed                  = false
  direct_access_grants_enabled        = false
  service_accounts_enabled            = false
  standard_flow_enabled               = false
  implicit_flow_enabled               = false
  use_refresh_tokens                  = true
  client_authenticator_type           = "client-secret"
  backchannel_logout_session_required = false
  name                                = "my-client"
  valid_redirect_uris = []
  web_origins         = []
}

resource "keycloak_generic_role_mapper" "my-client_realm-role-1" {
  realm_id  = "my-realm"
  client_id = keycloak_openid_client.my-client.id
  role_id   = keycloak_role.realm-role-1.id
}

resource "keycloak_generic_role_mapper" "my-client_realm-role-2" {
  realm_id  = "my-realm"
  client_id = keycloak_openid_client.my-client.id
  role_id   = keycloak_role.realm-role-2.id
}

resource "keycloak_generic_role_mapper" "my-client_realm-role-3" {
[...]

PR #772, opened 2022-11-15
fix: use payload when deleting keycloak_generic_role_mapper for realm roles

rd-robert-avram:
Fixes: #771

Everything I've added here is based on what I've seen Keycloak do when performing this action in the UI.

The test itself indirectly catches the misbehaviour described in the issue by failing due to the fact that there's a non-empty intermediate plan occurring during the test case itself. That plan is the creation of the resources that had never actually been explicitly destroyed in the first place.

I hope this is sufficient.


mrparkers (approved):
Thanks for opening this PR!  The bug report, reproduction steps, fix, and test case all look great!

Issue #782, opened 2022-12-13
keycloak_ldap_user_federation import with for_each fails


MatrixCrawler:
When trying to impoort an existing LDAP User federation we get the following error

`
terraform import module.customer.keycloak_ldap_user_federation.ldap_user_federation[\"xxx\"] XXX/UUID
module.customer.keycloak_ldap_user_federation.ldap_user_federation["xxx"]: Importing from ID "XXX/UUID"...
module.customer.keycloak_ldap_user_federation.ldap_user_federation["xxx"]: Import prepared!
  Prepared keycloak_ldap_user_federation for import
  
 
 Error: strconv.Atoi: parsing "": invalid syntax
`

We are using terraform 1.3.6 and provider version 4.1.0

MatrixCrawler:
I believ that this is the same problem as in #684

MatrixCrawler:
@mrparkers Any chance of fixing this?

PR #784, opened 2022-12-14
Fix for #782. Treat empty strings in import as zero

MatrixCrawler:
This should fix the probem described in #782


mrparkers (approved):
This looks good, thanks for submitting this PR!

MatrixCrawler:
Hey @mrparkers,
when could we have a release for that? This is actually a blocking issue for us :-)

mrparkers:
Hi, sorry about that.  I meant to do this week and it slipped my mind.  I published v4.2.0 today.

Issue #333, opened 2020-07-01
data `keycloak_group` fails with nested group


chanhht:
When trying `data keycloak_group` with the nested group, it returns an error
Step to reproduce:
```
// First, run a TF apply for groups creation
data "keycloak_realm" "realm" {
    realm   = "master"
}

resource "keycloak_group" "parent_group" {
    realm_id = data.keycloak_realm.realm.id
    name     = "parent-group"
}

resource "keycloak_group" "child_group" {
    realm_id  = data.keycloak_realm.realm.id
    parent_id = keycloak_group.parent_group.id
    name      = "child-group"
}

// Then run another TF plan/apply for query groups by datasource
data "keycloak_group" "parent_group" {
    realm_id = data.keycloak_realm.realm.id
    name     = "parent-group"
}

data "keycloak_group" "child_group" {
    realm_id = data.keycloak_realm.realm.id
    name     = "child-group"
}
```
Then received this error:
```
Error: no group with name child-group found

  on keycloak_test.tf line 30, in data "keycloak_group" "child_group":
  30: data "keycloak_group" "child_group" {
```
Both groups are present in Keycloak after the first creation run.

_Originally posted by @chanhht in https://github.com/mrparkers/terraform-provider-keycloak/issues/330#issuecomment-651817895_

PR #786, opened 2022-12-27
Handle nested group paths as names

Arkaniad:
We're trying to manage nested groups via Terraform. Our nested group hierarchy mirrored by this YAML:

``` yaml
teams:
 team-foo:
   read-only:
   developer:
   admin:
 team-bar:
   read-only:
   developer:
   admin:
```

Unfortunately, when attempting to query a group by path with provider 3.10.0 and 4.1.0 fails with the following message:

```
│ Error: no group with name /teams/team-foo/admin found
│ 
│   with data.keycloak_group.admin_groups["/teams/team-foo/admin"],
│   on .terraform/modules/dummy/main.tf line 1, in data "keycloak_group" "admin_groups":
│    1: data "keycloak_group" "admin_groups" {
│ 
```

Digging deeper it seems that the `search` parameter used [here](https://github.com/mrparkers/terraform-provider-keycloak/blob/aec21154d7ae77fcfc14db303bef5e1cc2e07036/keycloak/group.go#L135) is not returning any results when searching for the group by path.

I followed along with [#333](https://github.com/mrparkers/terraform-provider-keycloak/issues/333) and [#334](https://github.com/mrparkers/terraform-provider-keycloak/pull/334) but in our case we can't search by 'group name' as there will be multiple indistinguishable yet unique groups returned.

This is a possibly inefficient implementation of handling group paths - if `name` starts with a backslash it assumes that `name` is actually a `path` and to search subgroups downward from the first group specified in the path until it either finds the group specified at the end of the path or comes up empty handed.

### DIFF
--- a/keycloak/group.go
+++ b/keycloak/group.go
@@ -127,8 +127,54 @@ func (keycloakClient *KeycloakClient) GetGroup(ctx context.Context, realmId, id 
 	return &group, nil
 }
 
+func (keycloakClient *KeycloakClient) GetGroupByPath(ctx context.Context, realmId, path string) (*Group, error) {
+	// truncate empty first element
+	var searchGroups []string = strings.Split(path, "/")[1:]
+
+	idx := 0
+	// Get the root group at the start of this search path
+	searchRoot, err := keycloakClient.GetGroupByName(ctx, realmId, searchGroups[idx])
+
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


Arkaniad:
I'm thinking this was an older version of Keycloak we were using. After some re-working of our groups things are working sensibly now. I'm going to close this as I'm not sure if it'll be useful for anyone anymore. Sorry :D

Issue #634, opened 2022-01-03
Set manually realm ID


jitbasemartin:
Hi !

I would like to set manually the realm ID, like:

```
resource "keycloak_realm" "realm_1" {
  id                = "b2c07820-16de-4350-94d6-54d9f0796d2b"
  realm             = "realm-1"
  enabled           = true
  ...
}
```

But I got 

```
| Error: Invalid or unknown key
│ 
│   with keycloak_realm.realm_1,
│   on main.tf line 37, in resource "keycloak_realm" "realm_1":
│   37:   id                = "b2c07820-16de-4350-94d6-54d9f0796d2b"
```

Is it possible to do that ?

Thanks

cpesch:
Actually, in Keycloak 15 the Keycloak API accepts an ID that is passed. Just the terraform plugin insists on

` Error: Invalid or unknown key`

cpesch:
@mrparkers Hi Michael, I've tested that the Keycloak API accepts and ID this is passed. About three years ago, [Keycloaks ID of the realm was introduced as internal_id](https://github.com/mrparkers/terraform-provider-keycloak/commit/bdf9111bb6ccb59deeb9f02795f842431dc5a46f) and I wanted to ask if this could be changed so that the ID may be optionally like suggested by @jitbasemartin ?

cpesch:
Hi @jitbasemartin,
I'm in to process of setting up a bunch of Keycloak clusters for a 4 digit number of realms – all setup and provisioned by terraform from a central configuration database. And I want to allow the usage of all realm names that are _currently_ not used on any of the clusters. Due to Keycloaks "feature" to use the realm names for the ID, I'd additionally need to deny the usage of all realm names that were used at creation time and I'd like to avoid that.

Since I'm distributing the creation of new realms via events to downstream systems, I cannot create the realms with a UUID as their name and then switch to the desired name since that would produce two events. So I'd like to create a UUID in the configuration database and use it in the keycloak_realm resource and in the create/update/delete events for the downstream.
 
Does this sound reasonable?

cpesch:
Hi @dmeyerholt,
did you find time to take a look at the "set external ID for realm creation" issue?

PR #800, opened 2023-01-26
Manually set realm ID

cpesch:
A step forward to address issue #634 

I want to allow the usage of all realm names that are currently not used. Due to Keycloaks "feature" to use the realm names for the ID, I'd additionally need to deny the usage of all realm names that were used at creation time and I'd like to avoid that.

* Changes the internal_id from 'Computed' to 'Optional' to allow setting it externally to a UUID
* Avoids setting ResourceData's internal_id to null when it is not passed in setRealmData() to avoid endless updates from the ID to null
* Retrieves the realm name when deleting realms since the Keycloak API relies on it and with the change the realm ID might differ from the name

Successfully tested with 

```go
locals {
  realms = {
    "abc" = {},
    "def" = {}
  }
}

resource "keycloak_realm" "id_realm" {
  for_each     = local.realms
  internal_id  = "my-${each.key}-id"
  realm        = "my-id-${each.key}-realm-name"
}

resource "keycloak_realm" "name_realm" {
  for_each     = local.realms
  realm        = "my-named-${each.key}-realm"
}
```

@mrparkers @dmeyerholt please have a look

### DIFF
--- a/docs/resources/realm.md
+++ b/docs/resources/realm.md
@@ -82,7 +82,7 @@ resource "keycloak_realm" "realm" {
 - `enabled` - (Optional) When `false`, users and clients will not be able to access this realm. Defaults to `true`.
 - `display_name` - (Optional) The display name for the realm that is shown when logging in to the admin console.
 - `display_name_html` - (Optional) The display name for the realm that is rendered as HTML on the screen when logging in to the admin console.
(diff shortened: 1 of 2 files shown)


dmeyerholt:
Thanks for taking over this issue @cpesch .as long as the older tests from the initial PR are still working it should be fine. will try your branch in a local build with our extensive config this week.

Issue #798, opened 2023-01-23
strconv.ParseBool error on custom user federation resource import


jhelsethbushel:
Receiving an error when trying to import a custom user federation from keycloak. Here is the input:
```shell
terraform import module.setup_user_federation.keycloak_custom_user_federation.bushel_user_storage bushel/284937d8-8b15-48bc-b103-b9f6364e4979
```

Here is the output:
```shell
module.setup_user_federation.keycloak_custom_user_federation.bushel_user_storage: Importing from ID "bushel/284937d8-8b15-48bc-b103-b9f6364e4979"...

Error: strconv.ParseBool: parsing "": invalid syntax
```

When I was looking through the code in github I found that line 58 of `keycloak/custom_user_federation.go` is looking to parse the `enabled` property of the custom user federation resource. When I look at the server output from our Keycloak `18.0.2` wildfly server, `enabled` is not a returned property in the json. Is there a work around for this, or is it a bug? Either a bug in the terraform provider, or the keycloak api?

jhelsethbushel:
https://github.com/mrparkers/terraform-provider-keycloak/blob/aec21154d7ae77fcfc14db303bef5e1cc2e07036/keycloak/custom_user_federation.go#L58 is the place I was referencing for my description for this issue.

jhelsethbushel:
I had two keycloak servers I was working with and only encountered this error on one of them. On the server that was not returning the `enabled` property I changed the priority value in the admin panel and clicked save. When I reloaded the page, the api was now returning the `enabled` property and the terraform import for the customer user federation worked like it was supposed to.

I will keep this issue open because a missing optional `enabled` property should not prevent this import operation from working.

PR #809, opened 2023-02-18
#798 fix for parsing empty bool that leads to error.

MatrixCrawler:
Implemented fix for #798 where parsing an empty bool leads to an error. 
Similar error as in #782 and #684


mrparkers (approved):
Thanks again for your helpful PRs 😄

PR #812, opened 2023-02-22
Fix key_size argument name in the documentation

imykolenko:
Align argument name in documentation with code.

https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/resource_keycloak_realm_keystore_rsa_generated.go#L61


mrparkers (approved):
Thanks for the fix!

PR #817, opened 2023-02-27
add function to get all mappers from idp

chifu1234:
This function is needed for a none terraform project.


mrparkers (approved):
I'm happy to add more functionality to the `keycloak` package, just keep in mind that this package isn't covered by the versioning strategy for this project, and breaking changes may be published without being mentioned in the changelog.

Thanks for the PR!

PR #825, opened 2023-03-17
WIP: Support multivalue in component config

ChrisHubinger:
Make the Component.getConfig(key) method return the naive array value, not just the first element or "".
Provide Component.getConfigFirstOrDefault(key) that behaves like the original.


olivierboudet:
I ran into the same issue, perhaps this PR should be updated ?

PR #830, opened 2023-03-22
Allow passing a login hint to saml idps

toddkazakov:
Keycloak >19 (possibly lower versions too) allows passing a login hint to SAML IDPs. It's handled as a string, not as a quoted bool, to avoid backward compatibility issues with existing OIDC providers.


toddkazakov:
Hey @mrparkers could you please review this PR?

mrparkers (approved):
Sorry for the delay on this.  Thanks for submitting this PR!

PR #833, opened 2023-04-06
Add missing docs for resources

malfter:
First of all, many thanks for your work and the provision of this provider.

When using the provider, I noticed that there is still no documentation for some resources. I have added some in this pull request, I hope it fits so far.


mrparkers (approved):
These look good, thanks for the contribution!

PR #841, opened 2023-05-09
fix: avoid validation errors during plan for missing realm and openid client resources 

### LINKED ISSUE #738: Avoiding validation error: client with id <id> does not exist in keycloak_openid_client_default_scopes
Hi, I have a sandbox environment that allows a mix of clients managed by terraform and some manually provisioned resources. Occasionally, a client managed by Terraform gets deleted accidentally. Typically, Terraform will figure out that a resource was deleted while refreshing the state and offer to recreate it. However, for the `keycloak_openid_client_default_scopes` and `keycloak_openid_client_optional_scopes` resources, the missing client causes validation to fail during plan:

```
╷
│ Error: validation error: client with id <internal id> does not exist
│ 
│   with keycloak_openid_client_default_scopes.my_client,
│   on main.tf line 27, in resource "keycloak_openid_client_default_scopes" "my_client":
│   27: resource "keycloak_openid_client_default_scopes" "my_client" {
│ 
```

Currently, the workaround is to use `terraform state rm` on the client scopes before running `terraform apply` again. Could we update the provider to accept missing state as valid and allow the scope associations to be recreated?

kherock:
Fixes #738

This brings the 404 handling of missing OpenID clients in line with the implementation for missing SAML clients when resolving client scopes. Basically, the validation of the 404 response was occurring too early, and the read context was throwing instead of return a blank ID via the `handleNotFound` handler.

This also includes a similar fix for `generic_role_mapper` resources.

The way that I tested these fixes was to configure a realm using these resource types, and run `terraform refresh` after manually deleting the realm.

### DIFF
--- a/keycloak/openid_client.go
+++ b/keycloak/openid_client.go
@@ -241,9 +241,7 @@ func (keycloakClient *KeycloakClient) getOpenidClientScopes(ctx context.Context,
 	var scopes []*OpenidClientScope
 
 	err := keycloakClient.get(ctx, fmt.Sprintf("/realms/%s/clients/%s/%s-client-scopes", realmId, clientId, t), &scopes, nil)
-	if err != nil && ErrorIs404(err) {
-		return nil, fmt.Errorf("validation error: client with id %s does not exist", clientId)
-	} else if err != nil {
+	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


kherock:
I don't think I can fix the test error showing in CI without splitting the `*Reconcile` methods into separate create and update functions. Only the create function is supposed to show the expected validation error.

kherock:
I was able to restore the original message! Would really appreciate a review on this one

kherock:
@mrparkers sorry to be a bother, but could this be looked at soon?

mrparkers:
Sorry, I meant to get to this before the latest release, but it slipped my mind.

This LGTM, I'll merge after tests pass and cut a release today.

mrparkers (approved):

Issue #849, opened 2023-06-05
keycloak_user_template_importer_identity_provider_mapper creates invalid mapper for keycloak_oidc


stale-vegait:
When using keycloak\_user\_template\_importer\_identity\_provider\_mapper to create a username mapper for a keycloak\_oidc identity provider the mapper that ends up being created is invalid. It cannot be opened on the UI nor executed when logging in with the associated provider. 

This is due to the invalid value being passed for "identityProviderMapper". The expected value is  "oidc-username-idp-mapper" and the value that gets passed is "keycloak-oidc-username-idp-mapper". Updating the value in the DB manually fixes the issue.

stale-vegait:
It's the same issue as #803

PR #850, opened 2023-06-05
fix: set the correct value for the identityProviderMapper 

stale-vegait:
when using keycloak-oidc identity provider

fixes: #849


mrparkers (approved):

mrparkers:
Thanks for the PR!

Issue #862, opened 2023-07-18
Add support for custom LDAP mappers


NataliaKhodiakova:
For the custom LDAP mapper which is deployed into Keycloak as a custom LDAP mapper provider there is no configuration possibility at the moment.
The existing mappers do not allow setting the provider id and provider type for an LDAP mapper, the values are hardcoded.

PR #863, opened 2023-07-18
Feat: add LDAP custom mapper resource

NataliaKhodiakova:
Added resource to configure self-deployed custom LDAP mapper (#862)

fixes: #862


punycode:
Today, while working on terraforming a Keycloak installation for a customer, I came across the exactly same need and already wanted to fire up my Go IDE. Then I found this PR and can totally agree with the need for such a terraform resource.

I only have one additional thing to add: such a resource should support a map backed `config` attribute in the same way as [keycloak_generic_protocol_mapper](https://registry.terraform.io/providers/mrparkers/keycloak/latest/docs/resources/generic_protocol_mapper#config) ([code](https://github.com/mrparkers/terraform-provider-keycloak/blob/master/provider/resource_keycloak_generic_protocol_mapper.go)) does. This would be needed for some custom LDAP mappers, that need additional configuration.

NataliaKhodiakova:
Hi @punycode thanks for the feedback! I added the optional `config` attribute to the resource

mrparkers:
Hi @NataliaKhodiakova @punycode, sorry for the delay on this.  I was able to pull this down and experiment with this locally.  Thanks for adding the example and tests.  Everything here looks great

mrparkers (approved):

PR #885, opened 2023-10-11
Compute `extra_config` for data sources

### LINKED ISSUE #884: `extra_config` is empty on `keycloak_openid_client` data source
When resolving a Keycloak client's details using `data.keycloak_openid_client`, the extra_config object computes to an empty object. It seems like there's some code in place that prevents this value from being computed properly:

https://github.com/mrparkers/terraform-provider-keycloak/blob/bfee4d7cd1591b015f4db84cce12bea36b6e871f/provider/extra_config.go#L47-L50

I presume this is to avoid interfering with existing attributes that are managed outside of Terraform. I think we need to disable this condition when the provider is computing the value for a data source.

kherock:
Fixes #884

The [build of the first commit](https://github.com/mrparkers/terraform-provider-keycloak/actions/runs/6489290999) shows the expected bug that was originally hidden due to a bad test definition:

```
    data_source_keycloak_openid_client_test.go:52: Step 1/1 error: Check failed: 1 error occurred:
        	* Check 1/1 error: data.keycloak_openid_client.test_extra_config: Attribute "extra_config.key1" not set, but "extra_config.key1" is set in keycloak_openid_client.test_extra_config as "value1"
```

To fix, I modified the data source routines to set `extra_config` to the value directly obtained from the client attributes.

### DIFF
--- a/provider/data_source_keycloak_openid_client.go
+++ b/provider/data_source_keycloak_openid_client.go
@@ -250,6 +250,10 @@ func dataSourceKeycloakOpenidClientRead(ctx context.Context, data *schema.Resour
 	}
 
 	err = setOpenidClientData(ctx, keycloakClient, data, client)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	data.Set("extra_config", client.Attributes.ExtraConfig)
 
-	return diag.FromErr(err)
+	return nil
 }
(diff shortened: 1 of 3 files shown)


mrparkers (approved):
LGTM, thanks for submitting this PR!

PR #910, opened 2023-12-28
Add support for client policy and client profile on realms

### LINKED ISSUE #888: Support for client profiles and policies
It looks like the provider does not support 'client profiles' and 'client policies', as can be found under realm settings.

* Is this correct? 
* If so, what would it take to support this?

Relevant parts of REST API:
* https://www.keycloak.org/docs-api/22.0.4/rest-api/index.html#_get_adminrealmsrealmclient_policiespolicies
* https://www.keycloak.org/docs-api/22.0.4/rest-api/index.html#_get_adminrealmsrealmclient_policiesprofiles

JSON fragment from export realm:

```
  "clientPolicies": {
    "policies": [
      {
        "name": "Client secret rotation",
        "enabled": true,
        "conditions": [
          {
            "condition": "client-roles",
            "configuration": {
              "is-negative-logic": "true",
              "roles": [
                "no-rotation"
              ]
            }
          }
        ],
        "profiles": [
          "Client secret rotation"
        ]
      }
    ]
  }
```

image

yaron:
This PR fixes #888 and adds support for client_policy and client_profile settings on a realm resource.

It uses json for the configuration because that field can contain booleans, integers and strings and terraform does not like the unpredictability of that.

Example code:
```
  realm             = "my-realm"
  enabled           = true
  display_name      = "my realm"
  display_name_html = "<b>my realm</b>"

  client_profile {
    name        = "test profile"
    description = "testing"

    executor {
      name = "secure-ciba-signed-authn-req"
      configuration = jsonencode({
        available-period = "3600"
      })
    }
    executor {
      name = "secure-ciba-signed-authn-req"
      configuration = jsonencode({
        available-period = "3600"
      })
    }
  }

  client_policy {
    name        = "test policy"
    description = "description"
    profiles    = ["test profile"]
    enabled     = false

    condition {
      name = "any-client"
      configuration = jsonencode({
        is-negative-logic = false
      })
    }
  }
}

```

### DIFF
--- a/provider/data_source_keycloak_realm.go
+++ b/provider/data_source_keycloak_realm.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/keycloak/terraform-provider-keycloak/keycloak"
@@ -540,7 +541,10 @@ func dataSourceKeycloakRealmRead(ctx context.Context, data *schema.ResourceData,
 		return diag.FromErr(err)
 	}
 
-	setRealmData(data, realm)
+	diagErr := setRealmData(data, realm)
+	if err != nil {
+		return diagErr
+	}
 
 	return nil
 }
--- a/provider/data_source_keycloak_realm.go
+++ b/provider/data_source_keycloak_realm.go
@@ -527,6 +527,94 @@ func dataSourceKeycloakRealm() *schema.Resource {
 					Schema: webAuthnSchema,
 				},
 			},
+			// Client policies
+			"client_policy": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+							Optional: true,
+						},
+						"description": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
+						"enabled": {
+							Type:     schema.TypeBool,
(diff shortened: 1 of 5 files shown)


gim-:
We're looking forward for this to get merged. Is the failing test the only blocker right now?

yaron:
The failing test is not related to the change. This PR is waiting on the maintainer to have a look.

robson90:
Hey @yaron , awesome work !!! Maybe it is a better idea, to have ClientPolicy and ClientProfile as a seperate resource ?

For example:
keycloak_realm_client_policy
and
keycloak_realm_client_profile
and accorodingly for conditions and executors

yaron:
Hey @robson90 , the issue with that is that on the api call to create or update a realm, you need to provide the policies and profiles or they will be reset (see the issue that this solves). So if you have seperate resources it might break stuff if you update a realm, but nothing changes on the profile or policy.

kherock:
I have been able to run this branch in my stack and it's been working great so far. However, the one problem I have noticed @yaron is that the provider does not support destroying policies that it creates.

provider/data_source_keycloak_realm.go
```diff
@@ -526,6 +527,94 @@ func dataSourceKeycloakRealm() *schema.Resource {
...
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"name": {
+										Type:     schema.TypeString,
+										Optional: true,
+										Computed: true,
+									},
+									"configuration": {
+										Type:     schema.TypeString,
+										Optional: true,
```
kherock, line 571: for both client profiles and client policies, I get an error message like `java.lang.IllegalArgumentException: configNode must be not null for provider xxx`. Even if I pass an empty object `"{}"` as a value, the provider seems to send a null value to Keycloak.

This seems to be required by the Keycloak API, at least as of Keycloak 26.0!

PR #912, opened 2024-01-05
Update keycloak_group documentation for importing

PabloHiro:
I have been using the provider and it is possible to import existing groups, updating the documentation so it is aligned with what is implemented.


PabloHiro:
Sorry, confused `keycloak_user_groups` and `keycloak_group`, this does not apply

PR #917, opened 2024-01-11
add admin url override when different from Keycloak base url

frpicard:
When Keycloak is deployed with an administration hostname different from the base one (for example to whitelist admin accesses), the provider should use the base URL for client credentials authentication and the admin URL for anything else when calling Keycloak API.
Adds an optional parameter to override the base url.


tommasobrunowork:
Curious about this one as well, the latest keycloak allows to have multiple urls for admin / auth.

Issue #919, opened 2024-01-12
Provider crash when adding keycloak_default_roles on new realm


TBeijen:
Having `keycloak_default_roles` on a newly created realm causes a provider crash. If the realm already exists all works fine.

Sample code:

```hcl
resource "keycloak_realm" "sample" {
  realm = "tf_foo"
}

resource "keycloak_default_roles" "sample" {
  # Commenting this out on first apply and all works fine on second apply after uncommenting.
  realm_id      = keycloak_realm.sample.id
  default_roles = ["uma_authorization"]
}
```

Crash:

```
keycloak_default_roles.sample: Creating...
╷
│ Error: Plugin did not respond
│
│   with keycloak_default_roles.sample,
│   on main.tf line 47, in resource "keycloak_default_roles" "sample":
│   47: resource "keycloak_default_roles" "sample" {
│
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may contain more details.
╵

Stack trace from the terraform-provider-keycloak_v4.3.1 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x102744754]

goroutine 11 [running]:
github.com/mrparkers/terraform-provider-keycloak/provider.resourceKeycloakDefaultRolesReconcile({0x102ac1808, 0x140003b3560}, 0x0?, {0x102ab3140?, 0x14000380a20?})
  github.com/mrparkers/terraform-provider-keycloak/provider/resource_keycloak_default_roles.go:106 +0x174
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).create(0x140003a7260, {0x102ac1840, 0x1400032d5f0}, 0xd?, {0x102ab3140, 0x14000380a20})
  github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:707 +0xe8
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).Apply(0x140003a7260, {0x102ac1840, 0x1400032d5f0}, 0x1400019d2b0, 0x1400017f400, {0x102ab3140, 0x14000380a20})
  github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:837 +0x86c
[...]

PR #922, opened 2024-01-19
Fix 919 provider crash

TBeijen:
Fix for #919 

Always refresh token after create realm.

Rationale is this should not be expensive since:

* Creating a realm is relatively rare
* In a lot of scenarios the token refresh would happen anyway if a new realm is created, when it's referenced elsewhere

Existing acceptance test `testKeycloakDefaultRoles_basic` should cover this, provided the acceptance tests include Keycloak v22 and newer. For that, the following PRs exist:

* #921 Test Keycloak images for v22 & v23
* #920 Include v22 in acc. tests (draft, pending #921)

### DIFF
--- a/keycloak/realm.go
+++ b/keycloak/realm.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/hashicorp/terraform-plugin-log/tflog"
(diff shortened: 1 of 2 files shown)


vincent-petit (approved):
Seems good :)

PR #937, opened 2024-03-03
feat: add keycloak_realm_keystore_custom resource

rruxandra:
This feature introduces the capability to create and manage custom Realm keystores within Keycloak.

Currently, this provider only supports configuring keystores defined in the Keycloak core, leaving out those defined in extensions. This limitation prevents users from fully leveraging the provider's capabilities to manage their Keycloak infrastructure, as they are unable to configure all keystores uniformly through Terraform.

### DIFF
--- a/docs/resources/realm_keystore_custom.md
+++ b/docs/resources/realm_keystore_custom.md
@@ -1,0 +1,59 @@
+---
+page_title: "keycloak_realm_keystore_custom Resources"
+---
+
+# keycloak\_realm\_keystore\_custom Resources
+
+Allows for creating and managing custom Realm keystores within Keycloak.
+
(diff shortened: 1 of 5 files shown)


tboerger:
Any chance to get this merged and published? I'm currently really in the need to get a rsa-enc-generated key but am unable to provide it.

rruxandra:
> Any chance to get this merged and published? I'm currently really in the need to get a rsa-enc-generated key but am unable to provide it.

@mrparkers, what do you think?

tboerger:
Since this repo doesn't receive lot of merged as mrparkers doesn't use keycloak anymore I'm currently discussing an active fork with my boss where I would include this contribution.

PR #974, opened 2024-06-12
Add mTLS support

lucdew:
Hi,
The keycloak server can be protected with [mutual TLS authentication](https://www.keycloak.org/server/enabletls#_enabling_mutual_tls) done directly by Keycloak or with the help of  a reverse proxy.

The current pull request adds support for mTLS client authentication.

I have added 2 parameters instead of 1 single like some tools do (HAproxy for instance) where the client certificate and private key are provided as a file concatenation of PEM encoded assets. The main reason is that it allows them to be protected differently by the caller, the private key can be passed as a sensitive variable of a terraform module. It avoids parsing a file to extract them.
Also I think it is better to pass strings instead of file paths it gives more flexibility for the caller to store these assets.

But the drawback is that it adds 2 parameters and functions start to have too many parameters. I did not want to refactor that using option callbacks for instance.

No unit tests are provided.

I also go formatted the changed source files.

BR


lucdew:
Closing the pull reques.
I will re-create a new one following the module renaming (from mparkers) and other changes in the client. I had created the original from a repository that had already diverged from the source one.

PR #976, opened 2024-07-01
add support for unmanagedAttributePolicy in user profile

### LINKED ISSUE #946: keycloak_realm_user_profile: should support 'unmanagedAttributePolicy'
The `unmanagedAttributePolicy` attribute has been added to the [User Profile Config definition](https://www.keycloak.org/docs-api/24.0.2/rest-api/index.html#UPConfig) as of Keycloak version 24. This is exposed in the Admin UI as the _Unmanaged Attributes_ realm setting.

Allowed values are:
- _omitted_ - Disabled
- `ENABLED` - Enabled
- `ADMIN_VIEW` - Only administrators can view
- `ADMIN_EDIT` - Only administrators can write

The provider should allow this new attribute to be set in the `keycloak_realm_user_profile` resource.

ili16:
This PR extends the resource `keycloak_realm_user_profile` with the attribute `unmanagedAttributePolicy` which can be ommited or used with the values "DISABLED,ENABLED,ADMIN_VIEW,ADMIN_EDIT". I kept the changes as simple as possible in order to not break any legacy stuff even more than it currently is. It closes #946 

I did have to add the mandatory fields "username" and "email" to the user profile if one chooses to delete the user profile. This does have the adverse effect that creating a new realm has a different user profile config than creating a new realm and then deleting the user profile.


ili16:
obsolete, done by #1042

PR #1024, opened 2024-11-29
Added migration info to README.md

sschu:


thomasdarimont (approved):
LGTM

aslafy-z:
It looks like the new provider is not yet published to the terraform registry. Am I missing something?

sschu:
@aslafy-z You are right. I wanted to do the release today but I am running into a permission problem because the permissions required by the Terraform registry are quite excessive. I fear we have to wait until Monday to sort this out.

PR #1033, opened 2024-12-16
feat: allow configuration of display_name for oidc identity providers

### LINKED ISSUE #639: should be able to set a displayName on the `keycloak_oidc_google_identity_provider` as well
Hi,

There's an issue in newer keycloak's, that if an IDP Provider does not have a `displayName`, then the IDP Linking flow will fail with some `StringIndexOutOfBoundsException`, see: https://github.com/keycloak/keycloak/issues/9217

I cannot set the `displayName` with `keycloak_oidc_google_identity_provider`, so that triggers this error. Instead of that currently we use the `keycloak_oidc_identity_provider` resource, like this:

```hcl
resource "keycloak_oidc_identity_provider" "google_identity_provider" {
  realm         = keycloak_realm.public.id
  alias         = "google"
  client_id     = var.google_identity_provider_client_id
  client_secret = var.google_identity_provider_client_secret
  first_broker_login_flow_alias = local.flow_idp_first_broker_login
  post_broker_login_flow_alias = keycloak_authentication_flow.bkk-post-idp-flow.alias

  authorization_url = "https://not-used.realcity.io"
  token_url         = "https://not-used.realcity.io"

  provider_id = "google"
  sync_mode = "IMPORT"
  default_scopes = "openid profile email"
  display_name = "Google"
  backchannel_supported = false
}
```

Coudl you add the `display_name` attribute to the `keycloak_oidc_google_identity_provider` or set it to something by default?

Thanks,

bnjns:
I'm glad I've found something that allows me to contribute to this provider, as it's something I've been using for a really long time.

Came across this issue earlier today - this change allows users to configure the display name of OIDC and Google OIDC identity providers.

Resolves #639

### DIFF
--- a/provider/resource_keycloak_oidc_google_identity_provider.go
+++ b/provider/resource_keycloak_oidc_google_identity_provider.go
@@ -17,8 +17,9 @@ func resourceKeycloakOidcGoogleIdentityProvider() *schema.Resource {
 		},
 		"display_name": {
 			Type:        schema.TypeString,
+			Optional:    true,
 			Computed:    true,
-			Description: "Not used by this provider, Will be implicitly Google",
+			Description: "The human-friendly name of the identity provider, used in the log in form.",
 		},
 		"provider_id": {
 			Type:        schema.TypeString,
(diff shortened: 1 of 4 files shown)


sschu:
@bnjns Thanks for the contribution! Do you consider this ready for review?

bnjns:
> @bnjns Thanks for the contribution! Do you consider this ready for review?

@sschu Apologies, I completely forgot about this :facepalm: I wasn't sure exactly how much testing is necessary, but I think this is probably ready yeah

thomasdarimont (approved):
LGTM

PR #1036, opened 2024-12-20
feat: Recognise the 'apple' provider ID in attribute importer mapper

greed42:
This allows the provider to be configure `keycloak_attribute_importer_identity_provider_mapper` resources with the [Apple Identity Provider for Keycloak plug-in](https://github.com/klausbetz/apple-identity-provider-keycloak).

Fortunately, all the actual differences are buried in the provider itself and the mapper just needs to do the same thing as for `facebook` and `google`.


sschu (approved):
Thanks!

PR #1039, opened 2024-12-21
Fixed examples and user-federation-example

denniskniep:
* Added `user-federation-example` target to `make local`
* Fixed `custom-user-federation-example` by using dependencies in version 25.0.3 (More refactoring is necessary to compile it with 26.0.7)
* Fixed examples when running against KC 26.0.7 (There was an error regarding user-profile and deprecation warnings)
* Eliminated warns regarding usage of ubuntu-latest in CI


sschu (approved):
LGTM, thanks!

PR #1042, opened 2024-12-27
Fix user & userprofile tests

denniskniep:
This PR fixes some `user` & `userprofile` tests. Therefore it was necessary to implement the `UnmanagedAttributePolicy` property in `userprofile`.

I removed the `t.Parallel()` in certain tests, because those tests are failing when all tests are running!
Concrete problem is that `firstName` and `lastName` can not be stored.
Maybe there is a multithreading problem in Keycloak? Once this PR is merged I will prepare a branch to reproduce the problem and raise a issue.


faculbsz-ah:
Would close #946

sschu (approved):
LGTM, thanks a lot!

PR #1043, opened 2024-12-27
feat:(roles): Importable role by rolename

denniskniep:


sschu (approved):
LGTM, thanks a lot!

denniskniep:
@sschu thanks for the review, but the dependency is not new. It is already used by `resource_keycloak_openid_client.go` for the same purpose: Handle `import: true` (see [here](https://github.com/keycloak/terraform-provider-keycloak/blob/a68ebf1efc13fef8365dfa160a872ffcbb7694c2/provider/resource_keycloak_openid_client.go#L536C18-L536C23))

sschu (approved):
I see. Ok, then I guess this is fine.

kherock:
After testing this provider against my stack, my plans are showing hundreds of replacements for the roles caused by this attribute:

```
  # keycloak_role.admin_console_administrator must be replaced
-/+ resource "keycloak_role" "admin_console_administrator" {
      ~ attributes      = {} -> (known after apply)
      + composite_roles = (known after apply)
      ~ id              = "3af3eb01-7c22-4d30-89e3-77194f4dc445" -> (known after apply)
      + import          = false # forces replacement
        name            = "administrator"
        # (3 unchanged attributes hidden)
    }
```

Is this expected?

PR #1054, opened 2025-01-03
Prevent force replacing resources when `import` value in state is nil

### LINKED ISSUE #1053: Existing Role & User are forced for replacement with new introduced import property
### Describe the bug

@kherock reported the following issue with `keycloak_role` resource [here](https://github.com/keycloak/terraform-provider-keycloak/pull/1043#issuecomment-2569541391)

> After testing this provider against my stack, my plans are showing hundreds of replacements every roles caused by this new attribute:

```
  # keycloak_role.admin_console_administrator must be replaced
-/+ resource "keycloak_role" "admin_console_administrator" {
      ~ attributes      = {} -> (known after apply)
      + composite_roles = (known after apply)
      ~ id              = "3af3eb01-7c22-4d30-89e3-77194f4dc445" -> (known after apply)
      + import          = false # forces replacement
        name            = "administrator"
        # (3 unchanged attributes hidden)
    }
```

### Version

nightly

### Expected behavior

An already existing role/user should not introduce changes when missing `import` property

### Actual behavior

There are changes for newly introduced `import` property with "forces replacement"

### How to Reproduce?
* `git checkout main`
* ` make local`
* `git checkout v4.5.0`
* `make build-example`
* `git checkout main`
* `cd example`
* `terraform init`
* Remove priority Property from resource `keycloak_authentication_execution` and `keycloak_authentication_subflow`
* `terraform plan -out tfplan`
* `terraform apply tfplan`
* ` rm -r terraform.d && rm -r .terraform && rm .terraform.lock.hcl`
*  `cd ..`
* `rm terraform-provider-keycloak_v4.5.0`
* `make build-example`
* `cd example`
* `terraform init`
* `terraform plan -out tfplan`

Example
```
  # keycloak_role.pet_api_read_pet_details must be replaced
-/+ resource "keycloak_role" "pet_api_read_pet_details" {
      ~ attributes      = {} -> (known after apply)
      + composite_roles = (known after apply)
      ~ id              = "0cf59187-d7d9-4537-af7c-56ea3c97d793" -> (known after apply)
      + import          = false # forces replacement
        name            = "read-pet-with-details"
        # (3 unchanged attributes hidden)
    }
```

kherock:
Fixes #1053. I believe that Terraform state older than v4.0.0 with `keycloak_openid_client` resources is also impacted by this, so I've tried to address include a fix for those as well.

### DIFF
--- a/provider/resource_keycloak_openid_client.go
+++ b/provider/resource_keycloak_openid_client.go
@@ -571,6 +571,10 @@ func resourceKeycloakOpenidClientRead(ctx context.Context, data *schema.Resource
 		return diag.FromErr(err)
 	}
 
+	if _, ok := data.GetOk("import"); !ok {
+		data.Set("import", false)
+	}
+
 	return nil
 }
 
--- a/provider/resource_keycloak_role.go
+++ b/provider/resource_keycloak_role.go
@@ -213,6 +213,10 @@ func resourceKeycloakRoleRead(ctx context.Context, data *schema.ResourceData, me
 		data.Set("composite_roles", compositeRoleIds)
 	}
 
+	if _, ok := data.GetOk("import"); !ok {
+		data.Set("import", false)
+	}
+
 	return nil
 }
 
--- a/provider/resource_keycloak_user.go
+++ b/provider/resource_keycloak_user.go
@@ -264,6 +264,10 @@ func resourceKeycloakUserRead(ctx context.Context, data *schema.ResourceData, me
 
 	mapFromUserToData(data, user)
 
+	if _, ok := data.GetOk("import"); !ok {
+		data.Set("import", false)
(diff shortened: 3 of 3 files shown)


denniskniep:
@kherock  I solved it [here](https://github.com/keycloak/terraform-provider-keycloak/pull/1055) by removing the default value of the import property. Not sure which is the better approach.
Plz treat it simply as another option

kherock:
This PR's approach is slightly better in that it allows for someone to add an explicit `import = false` argument to these resources without the provider planning to replace the resource.

denniskniep:
Yeah, that makes sense.
Thanks for your fix 👍

sschu (approved):
LGTM, thanks!

PR #1071, opened 2025-01-13
fix: allow definition of multivalued user profile attributes

### LINKED ISSUE #1064: Support multivalued on keycloak_realm_user_profile
### Description

the keycloak_realm_user_profile attributes allow a boolean `multivalued` flag that allows admins to enter multiple values for the profile attribute.

See: https://www.keycloak.org/docs-api/latest/rest-api/index.html#UPAttribute

### Discussion

_No response_

### Motivation

This is supported by both the API and the Admin UI and is needed for cases where a profile attribute needs to be a list of values.

### Details

_No response_

thomasdarimont:
In order to replicate the existing userprofile configuration, we need to support the multivalued attribute.

We now support the `multi_valued = bool` attribute.

Example:
```hcl
  attribute {
    name = "myAttr"

    multi_valued = true

    validator {
      name = "options"
      config = {
        options = jsonencode(["opt1", "opt2", "opt3"])
      }
    }

    annotations = {
      foo = jsonencode({ "key" : "val" })
    }
  }
 ```

 Fixes #1064

### DIFF
--- a/docs/resources/realm_user_profile.md
+++ b/docs/resources/realm_user_profile.md
@@ -28,6 +28,7 @@ resource "keycloak_realm_user_profile" "userprofile" {
     display_name = "Field 1"
     group        = "group1"
 
+    multi_valued = false
     enabled_when_scope = ["offline_access"]
 
     required_for_roles  = ["user"]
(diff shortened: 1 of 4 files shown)


sschu (approved):
LGTM!

Issue #701, opened 2022-06-24
Unable to update `default_default_client_scopes`


Stono:
Hi, 
I'm unable to update an existing realms `default_default_client_scopes` and `default_optional_client_scopes`.

As you can see, they get picked up in the changes:

```
  # module.master-realm.keycloak_realm.default will be updated in-place
  ~ resource "keycloak_realm" "default" {
      ~ default_default_client_scopes            = [
          + "email",
          + "profile",
          + "roles",
        ]
      ~ default_optional_client_scopes           = [
          + "offline_access",
        ]
        id                                       = "master"
        # (43 unchanged attributes hidden)
        # (5 unchanged blocks hidden)
    }

Plan: 0 to add, 2 to change, 0 to destroy.
```

And the apply phase appears to apply them:
```
module.master-realm.keycloak_realm.default: Modifying... [id=master]
module.master-realm.keycloak_realm.default: Modifications complete after 1s [id=master]
```

But nothing has changed.

Running it again yields the same results.

This is on keycloak `18.0.1`, with the latest version of the provider.

gps035:
Looks like this might be a duplicate of #547. If not, #470 is definitely related

Issue #547, opened 2021-06-16
Empty `default_default_client_scopes` when creating a realm does not force an empty list of default client scopes


MiguelAlho:
When creating a realm, the docs do refer to the `default_default_client_scopes` property that allows you to set the default scopes to add to clients. This can be used to set and change the defaults as long as a value is added to the set; and empty set in the TF file on this property will not create an empty set of defaults in Keycloak. It is simply ignores, and Keycloak will continue to use it's own defaults (the doc feels a bit ambiguous in this aspect).

Is this the expected behaviour, and/or would there be a good way to be able to clear the default scopes added to clients in the realm?

meckhardt:
This is also true for other settings that you want to reset. For instance setting a description to "" also will not be applied to the resource.

mrparkers:
Right now, the support for setting scopes on the realm level is incomplete, see https://github.com/mrparkers/terraform-provider-keycloak/issues/470.

I'll leave this issue up for the time being.  I don't personally have the bandwidth to take this issue on, but I think the approach that @meckhardt mentioned is a good workaround.  I'd also recommend using something like https://registry.terraform.io/modules/matti/resource/shell/latest to simplify the process of running arbitrary scripts within your Terraform code.

MiguelAlho:
Thanks! Good to know. I didn’t know about null_resource to execute scripts, and I think I can leverage it in other places, so thanks for that @meckhardt  and @mrparkers!

olol13:
With provider version 3.70 in combination with keycloak 16.1.0 the workaround doesn't seem to for me. I deleted the default and optional client scopes on realm level and set them via default_default_client_scopes and default_optional_client_scopes respectively. Although terraform reports applied changes, the lists on server side stay empty.

Issue #470, opened 2021-01-25
Support for Scope configuration on realm level is incomplete


thomasdarimont:
In PR #464 we introduced support for the configuration of client scopes on realm level.

As it turned out this was not complete and needs to be reworked.
- [`clientScopes`](https://github.com/keycloak/keycloak/blob/eacc95b6998cd16d8a0574660946ee7d8727b527/core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java#L147) need to be provided as a list of [ClientScopeRepresentation](https://github.com/keycloak/keycloak/blob/master/core/src/main/java/org/keycloak/representations/idm/ClientScopeRepresentation.java)

Custom scopes are created during [realmImport via RepresentationToModel#realmImport](https://github.com/keycloak/keycloak/blob/eacc95b6998cd16d8a0574660946ee7d8727b527/server-spi-private/src/main/java/org/keycloak/models/utils/RepresentationToModel.java#L305)

blajoie55:
It's been almost 3 years and this feature still doesn't work (despite the feature being in the documentation for the [keycloak_realm resource](https://registry.terraform.io/providers/mrparkers/keycloak/latest/docs/resources/realm#default-client-scopes)) and this issue is still open. We have hit this same problem now at my company. Is there going to be any movement on this?

nicolas-urbantz:
We also have the same problem and had to make an additional API call. Is it planned to be fixed?

Thanks

TBeijen:
It might be worth considering moving the attribute [realm.default_default_client_scopes](https://registry.terraform.io/providers/mrparkers/keycloak/4.4.0/docs/resources/realm#default_default_client_scopes) to a separate resource. Similar to how it is handled on a per-client basis via [keycloak_openid_client_default_scopes](https://registry.terraform.io/providers/mrparkers/keycloak/latest/docs/resources/openid_client_default_scopes)

It's easy to hit chicken-egg dependency problems in the current setup:
* Adding a new client scope requires realm to exist
* Then how to add a non-built-in client scope to the `default_default_client_scopes` list that is part of the realm definition?

pboehm:
@thomasdarimont @TBeijen I have implemented the management of default and optional client scopes on a realm level via dedicated resources in this PR https://github.com/keycloak/terraform-provider-keycloak/pull/1079

PR #1079, opened 2025-01-20
Configure default and optional client scopes per realm via dedicated resources

pboehm:
Allow configuring the set of "default client scopes" and "optional client scopes" for a realm using two new dedicated resources `keycloak_realm_default_client_scopes` and `keycloak_realm_optional_client_scopes`.

There are currently multiple open issues with the existing solution via attributes on the `keycloak_realm` resource #470 #547 #701 which should all be addressed by this PR.


thomasdarimont (approved):
LGTM

PR #1088, opened 2025-01-24
Update to version 5.1.0

sschu:


thomasdarimont (approved):
LGTM, I agree it is better to do a minor release because of the new features.

PR #1092, opened 2025-01-24
feat: add all supported alogithms for rsa-enc keystore

tboerger:
This PR adds simply all algorithms which are available for keystores based on rsa-enc provider. Currently it got only a single one while there are 3 different available.

### DIFF
--- a/provider/resource_keycloak_realm_keystore_java_keystore.go
+++ b/provider/resource_keycloak_realm_keystore_java_keystore.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+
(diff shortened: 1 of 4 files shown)


tboerger:
Funny that the tests have been working before, should we split the `rsa` and `rsa-enc` algorithms?

tboerger:
I have split the valid alorithms and tried to add some basic testing for both variants.

tboerger:
Getting some review for this PR would be pretty great, so far it's not possible to configure supported encryption keys via Terraform.

sschu:
@tboerger I'll have a look later today.

sschu (approved):
LGTM, thanks!

PR #1105, opened 2025-01-30
doc: fix token exchange doc urls

emouty:
Urls for token exchange documentation where not valid anymore so I updated them.


sschu (approved):
Thanks!

PR #1112, opened 2025-02-03
Correct incorrect descriptions for session timeout arguments

joshhodgson:
As per title, the descriptions were the wrong way round for normal and offline session arguments.


sschu:
@joshhodgson Can you please also sign off your commit (just do `git commit --amend -s` and force-push your branch)?

Issue #699, opened 2022-06-22
Feature Request : Add realm localizations to keycloak_realm Resource


artur-baltabayev:
Hello,
during our evaluation we noticed that the provider doesn't support realm localizations, which would be very important for us.
A localization is defined through a locale (e.g 'en') and a list of key / value pairs, where the key is meant to be a message key (that can be used in custom .ftl templates) and the value is the actual translation.

What do you think about this ? Any feedback would be greatly appreciated !

Greetings,
Artur

donarus:
voting UP .. this is also our case and we'd welcome this feature

PR #1123, opened 2025-02-14
feat: add realm translation resource

jonathandavies-CS:
Closes https://github.com/keycloak/terraform-provider-keycloak/issues/699


JoKrefting:
@sschu can you pls take a look and merge if this PR is fine with you? I could really make use of this instead of manually patching localization information. :)

jonathandavies-CS:
@JoKrefting Thanks for your review. What you understood is correct. I've now renamed as per your recommendations which definitely make more sense :)

example/main.tf
```diff
@@ -115,10 +123,6 @@ resource "keycloak_required_action" "update-password" {
   default_action = true
   enabled        = true
   name           = "Update Password"
-
-  config {
-    max_auth_age = "600"
```
sschu, line 120: @jonathandavies-CS  Any specific reason why this was deleted?

sschu (commented):
In general LGTM, thanks for the contribution. Can you please address the one comment I have?

sschu:
@jonathandavies-CS So this is good to go from your perspective?

jonathandavies-CS:
@sschu Yes I'm happy for it to go in 👍

sschu (approved):
Thanks a lot, LGTM!

PR #1144, opened 2025-03-12
fix: Render json `null` to indicate an empty value of type KeycloakSliceHashDelimited (#1142).

### LINKED ISSUE #1142: Error "Invalid Redirect Uri" for client without valid_post_logout_redirect_uris
### Describe the bug

I try to create an OIDC client while a Client Policy with “secure-redirect-uris-enforcer” is enabled.
For the client the property “valid_post_logout_redirect_uris” is not specified because the application does not support a Logout.
The creation fails with:
```
│ Error: error sending POST request to /auth/admin/realms/test/clients: 400 Bad Request. Response body: {"error":"invalid_request","error_description":"Invalid Redirect Uri: invalid uri"}
│
│   with keycloak_openid_client.cli,
│   on main.tf line 27, in resource "keycloak_openid_client" "cli":
│   27: resource "keycloak_openid_client" "cli" {
```
It turns out that the provider sends an empty string for the attribute "valid_post_logout_redirect_uris".
Keycloak deserializes this to an array of length 1 and an empty string in the first element.
The SecureRedirectUrisEnforcerExecutor throws an exception because it does not allow empty URIs.

When adding valid_post_logout_redirect_uris = ["+"] the creation succeeds.

### Version

25.0.6 

### Expected behavior

Client creation should succeed.

### Actual behavior

Client creation fails.

### How to Reproduce?

1. Keycloak
Have a Keycloak running at  http://localhost:8080/auth and admin/admin credentials.

2. Create realm
Create realm "test" with Client Policy using "secure-redirect-uris-enforcer".
```bash
./kcadm.sh config credentials --server http://localhost:8080/auth --realm master --user admin --password admin

./kcadm.sh create realms -s realm=test -s enabled=true

./kcadm.sh update client-policies/profiles -r test -f - << EOF
{
  "profiles" : [ {
    "name" : "secure-redirect",
    "description" : "",
    "executors" : [ {
      "executor" : "secure-redirect-uris-enforcer",
      "configuration" : {
        "allow-http-scheme": true,
        "allow-ipv4-loopback-address": true,
        "allow-ipv6-loopback-address": false,
        "allow-open-redirect": false,
        "allow-private-use-uri-scheme": false,
        "allow-wildcard-context-path": true,
        "oauth-2-1-compliant": false
      }
    } ]
  } ]
}
EOF

./kcadm.sh update client-policies/policies -r test -f - << EOF
{
  "policies" : [ {
    "name" : "secure-redirect",
    "description" : "",
    "enabled" : true,
    "conditions" : [ {
      "condition" : "any-client",
      "configuration" : {
        "is-negative-logic" : false
      }
    } ],
    "profiles" : [ "secure-redirect" ]
  } ]
}
EOF

```
3. Create new client via terraform

```bash
terraform init
terraform plan
terraform apply -auto-approve
```

main.tf
```
terraform {
  required_providers {
    keycloak = {
      source  = "keycloak/keycloak"
      version = "5.1.1"
    }
  }
}

provider "keycloak" {
  url       = "http://localhost:8080/auth"
  client_id = "admin-cli"
  username  = "admin"
  password  = "admin"
}

import {
  to = keycloak_realm.test
  id = "test"
}

  realm   = "test"
  enabled = true
}

resource "keycloak_openid_client" "cli" {
  realm_id                     = keycloak_realm.test.id
  access_type                  = "PUBLIC"
  client_id                    = "cli"
  enabled                      = true
  standard_flow_enabled        = true
  direct_access_grants_enabled = false
  valid_redirect_uris = [
    "http://localhost:8000",
  ]
  web_origins = [
    "+"
  ]
}
```

thomasdarimont:
Previously, `KeycloakSliceHashDelimited` rendered an empty `""` string for a `nil` slice, which resulted in an attribute value being created in Keycloak. The empty `""`attribute then caused issues during validation of `Post Logout URIs`. We now emit a `null` value  to indicate a missing value, the null value is then ignored by Keycloak when creating / updating the client.

Note to explicitly override an set an empty value, an empty slice can be used, e.g:
```
valid_post_logout_redirect_uris = []
```

Fixes #1142

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,14 +22,13 @@ services:
     environment:
     - KC_BOOTSTRAP_ADMIN_USERNAME=keycloak
     - KC_BOOTSTRAP_ADMIN_PASSWORD=password
-    - KC_LOG_LEVEL=INFO
+    - KC_LOG_LEVEL=INFO,org.keycloak:debug
     - KC_DB=postgres
     - KC_DB_URL_HOST=postgres
     - KC_DB_URL_PORT=5432
     - KC_DB_URL_DATABASE=keycloak
     - KC_DB_USERNAME=keycloak
     - KC_DB_PASSWORD=password
-    - KC_LOG_LEVEL=INFO
     - KC_LOG_CONSOLE_COLOR=true
     - KC_FEATURES=preview
     - QUARKUS_HTTP_ACCESS_LOG_ENABLED=true
--- a/keycloak/types/keycloak_slice_hash_delimited.go
+++ b/keycloak/types/keycloak_slice_hash_delimited.go
@@ -12,7 +12,8 @@ type KeycloakSliceHashDelimited []string
 func (s KeycloakSliceHashDelimited) MarshalJSON() ([]byte, error) {
 	var buf bytes.Buffer
 	if s == nil || len(s) == 0 {
-		buf.WriteString(`""`)
+		// Allows omitempty to work
+		return []byte("null"), nil
 	} else {
 		buf.WriteString(`"`)
 
--- a/keycloak/types/keycloak_slice_hash_delimited_test.go
+++ b/keycloak/types/keycloak_slice_hash_delimited_test.go
@@ -1,0 +1,31 @@
+package types
+
+import (
+	"reflect"
+	"testing"
+)
+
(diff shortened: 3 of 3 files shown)


sschu (approved):
LGTM!

PR #1169, opened 2025-04-01
Increase MaxIdleConnsPerHost in http.Transport

Filirom1:
The default behaviour of http.Transport does not handle the use case of "many concurrent requests against a single host" very well.

By default, it will only keep 2 persistent connections in the pool. Everything above that is burst capacity, and those connections are closed. The result is a lot of connection churn, which can lead to port exhaustion (especially on more constrained resources like a shared NAT).

Increase the idle pool from 2 to 100, allowing for more connection reuse and in turn reducing connection churn.


sschu:
Thanks!

PR #1177, opened 2025-04-08
Allow the alias of Google IdP to be set

### LINKED ISSUE #1164: Set alias and display_name to keycloak_oidc_google_identity_provider
### Description

When creating an IDP Google from Keycloak UI, we can set custom alias and display name.
However, without the terraform provider, alias and display_name are only computed, we can not set them to a custom value.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

irby:
Resolves #1164 

Allows the `alias` property of the Google IdP to be configured.


sschu (approved):
LGTM, thanks!

PR #1189, opened 2025-04-29
Add support for Standard Token Exchange configuration in OpenID Clients

sventorben:
- Introduced `StandardTokenExchangeEnabled` field to `OpenidClientAttributes` struct.
- Updated `ValidateOpenidClient` to enforce that standard token exchange cannot be enabled on public clients.

This enhancement enables managing the Standard Token Exchange setting for OpenID Clients through the Keycloak Terraform provider.


sschu (approved):
LGTM, thanks a lot!

KyriosGN0:
thanks for the PR, just been looking for this

Issue #1058, opened 2025-01-07
Strategy for keeping OIDC client secrets out of Terraform state


ianwestcott:
### Description

I am evaluating this provider for use in my organization. One potential dealbreaker for us is that certain resources have attributes with values that are considered sensitive, and those sensitive values are cached in Terraform's state file, even if the configuration does not manage the attribute.

The example that is most pertinent to use is the `keycloak_openid_client` resource, which has a `client_secret` attribute. The [provider documentation](https://registry.terraform.io/providers/keycloak/keycloak/latest/docs/resources/openid_client#client_secret-2) notes, _"this value is sensitive and should be treated with the same care as a password. If omitted, this will be generated by Keycloak."_ However, even if the `client_secret` is omitted from configuration, the sensitive value is still cached into the `terraform.tfstate` file on apply. 

### Motivation

Credentials and other secrets should only be stored where needed, and only if absolutely necessary. Caching Keycloak OIDC client secrets in Terraform is not necessary if we aren't asking Terraform to manage the value, and doing so creates unnecessary risk that the client secret could be compromised.

### Details

My preference would be that, if the `client_secret` attribute is not specified in the `keycloak_openid_client` resource, its value should not be cached in Terraform state. Note that this means the client secret would have to be extracted from the Keycloak console or API as part of a separate credential management workflow. I believe that's a necessary tradeoff in order to ensure that OIDC client secrets are adequately protected.

raffraffraff:
Right now the only options for showing resource outputs are:
1. Show it
2. Hide it (sensitive = true)
3. Do not add a value to the output at all

As @sschu  said, if the value for client secret isn't added to the state at all, this would affect any future use of it eg: referencing the client secret when creating an OIDC IDP that requires client ID and secret.

What if we could selectively choose whether or not to add the client secret to the resource outputs at all? That would keep it out of the state entirely. This might be achieved by wrapping these lines in an `if` statement that references some boolean parameter in the oidc client resource (eg: "discard_client_secret = true"):
```
	client.ClientSecret = clientSecret.Value
``` 

Obviously, if you need to reference that value then don't set that boolean to true! But the rest of us, who do not use the client secret in any other Terraform resource, the secret is no longer in the state. I quickly skimmed the rest of the keycloak provider, and aside from markdown files and test cases, I only see it referred to in identity provider resources. If you're not creating an OIDC identity provider, you don't need it.

[...]

PR #1190, opened 2025-04-29
Add support for ephemeral/write-only arguments for keycloak_oidc_identity_provider & keycloak_openid_client

refucktor:
## Summary

This PR adds support for write-only arguments, specifically focusing on handling client secrets more securely.

## Key Changes

- Affected resources `keycloak_openid_client` and `keycloak_oidc_identity_provider`
- Added `client_secret_wo` as a write-only field that never persists to Terraform state.
- Added `client_secret_wo_version` as a tracking field to detect changes to the write-only secret.
- Implemented conflict handling between the standard and write-only approaches.
- Added library built-in validation to encourage using the write-only field pattern for better security

### Issues Related:

- #1058


datenzar:
Hi @sschu,
is there anything you need in order to getting this PR merged? We'd love to use this feature as it allows to manage our Keycloak configuration without storing secrets in TF state.

sschu:
@datenzar Currently, I am just missing time. I hope I will have more time next week to look at the open PRs...

sschu:
@datenzar I was drowning in some incident handling for the last three weeks, I am sorry. I hope I can pick this up on Monday and give you at least a reliable statement.

sschu:
@datenzar I am on it, should come out today.

refucktor:
@sschu I'm trying to run the tests without success, did you/team made any changes in the setup that requires HTTPS now for the communication?

refucktor:
@sschu when I start the docker-compose and try to reach Keycloak from the browser, I get `HTTPS required`. I try to run the tests in the exact same way they run when I created the PR and it's failing.

This is very much a personal opinion but this screams a lot like flaky tests, you should consider moving to [testcontainers](https://testcontainers.com/modules/keycloak/?language=go) to have better grasp on the integration services

sschu:
Why do you think docker-compose causes tests to be flaky? It launches containers in a pretty well-defined way and I never had any issues running them locally. The test setup could generally be improved and made more consistent between CI and local tests, maybe also using test containers. But there are no huge problems with test stability currently. You are free to provide a PR to use test containers if you think that would improve the situation and I will have a look.

irby:
@refucktor @sschu I am also running into the `HTTPS required` issue. Why does this occur? I'm launching the Docker conainers with `docker compose up --build`, and I see this when I visit `http://localhost:8080`.

image

This is impeding my ability to test with Keycloak. None of the API calls will work because of the HTTPS requirement.

cwegener:
@refucktor Why does the `resource_keycloak_oidc_identity_provider` (and **only** that resource type) have a custom validation callback that operates on the raw cty values?

The custom  validation callback function is causing bugs in my terraform configs (and everything works fine if I remove the callback)

(from the plugin SDK docs ...):

> 	// Developers should prefer other validation methods first as this validation function
>	// deals with raw cty values.
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema

cwegener:
The test has been added in #1433 

@sschu can you guide on how to properly contribute the PR? I can't find a CONTRIBUTING.md file ...

cwegener:
> I'm curious if you are facing the issues trying to migrate to the new `wo` attributes, or if it is for a new resource definition?

I have only tested new resources definitions.

I never had a use case where I wanted to migrate an existing resource from a regular attribute to a `wo` attribute, so I did not think of testing a case like that. I might look at that as well, but that seems like an edge case.

PR #1201, opened 2025-05-15
Docs: Updated Markdown to format correctly

bubbletroubles:
On the page https://registry.terraform.io/providers/keycloak/keycloak/latest/docs/data-sources/openid_client_service_account_user

the Attributed Reference displays as one block with no page breaks. This PR adds two spaces to the end of each line to ensure the docs format correctly

image

### DIFF
--- a/docs/data-sources/openid_client_service_account_user.md
+++ b/docs/data-sources/openid_client_service_account_user.md
@@ -55,10 +55,10 @@ resource "keycloak_user_roles" "service_account_user_roles" {
 
 ## Attributes Reference
 
-`username` - (Computed) The service account user's username.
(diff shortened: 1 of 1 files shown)


sschu:
@bubbletroubles Thanks for your PR! I think the correct way to do it would by by using a bullet point list by just adding a `-` in front of every line (see the argument reference above in the same file).

bubbletroubles:
> @bubbletroubles Thanks for your PR! I think the correct way to do it would by by using a bullet point list by just adding a `-` in front of every line (see the argument reference above in the same file).

Thanks @sschu - PR has been updated to make it a bullet point list as requested.

sschu (approved):
Thanks!

PR #1233, opened 2025-07-01
Added support for configuring provider with token

christopher-svensson-stegra:
We found that neither of the two available provider configuration and authentication methods (username/password or client credentials) suited our needs, as we require users to complete multi-factor authentication (MFA) before applying changes to our IdP platform.

This pull request introduces a third option: allowing users to supply their own token to authenticate against the Keycloak instance.

To support this, we’ve developed our own CLI—similar to the AWS CLI (e.g. `aws sso login --sso-session stegra`)—which launches a browser for user authentication (including MFA) and then stores the resulting token locally. This token can then be passed to the provider configuration as follows:

```terraform
locals {
  jwt = jsondecode(file(pathexpand("path/to/jwt.json")))
}

provider "keycloak" {
  url       = "<keycloak-url>"
  client_id = "admin-cli"
  token     = local.jwt.access_token
}
```

You can also provide the token via the `KEYCLOAK_TOKEN` environment variable, which will be used if the `token` argument is not specified in the provider configuration.


keycloak/keycloak_client.go
```diff
@@ -60,22 +61,28 @@ var redHatSSO7VersionMap = map[int]string{
...
 	}
-	if password != "" && username != "" {
-		clientCredentials.Username = username
-		clientCredentials.Password = password
-		clientCredentials.GrantType = "password"
-	} else if clientSecret != "" {
-		clientCredentials.GrantType = "client_credentials"
+
+	if token != "" {
+		clientCredentials.AccessToken = token
```
christopher-svensson-stegra, line 71: Perhaps the argument/variable should be called `accessToken` to make it clear all the way up what kind of token this is.

killerwhile:
Cool stuff, thanks! Interestingly I wrote the same patch few days back and when I wanted to open a PR, I found this one. The details are slightly different, especially the env var name, but the feature is exactly the same: https://github.com/weisshorn-cyd/terraform-provider-keycloak/tree/feature/externally-provided-token
I'm definitely +1'ing this PR, thanks!

thomasdarimont:
Interesting use case, do you always manually trigger IdP changes then?

Just today, we added support for client authentication via private key JWT which involved changes to keycloak_client.go. Would you mind rebasing this branch and adjust keycloak_client accordingly?

In an upcoming iteration, we'll probably refactor the keycloak_client type to use a dedicated KeycloakClientConfig struct instead of passing all client values via method parameters.

mpas97:
Any updates here? Would love to see that released soon

thomasdarimont:
I'll take a look at this this week. I created #1319. 

My plan is to adapt this PR to the changes in #1317 add @christopher-svensson-stegra as a co-author and add additional documentation.

Is this fine with you @christopher-svensson-stegra?

PR #1234, opened 2025-07-01
Openid client secret rotation

laureat-natzka:


laureat-natzka:
@sschu Im using Terraform to manage Keycloak openid clients and storing their credentials (client secrets) in a secret management service e.g. Hashicorp Vault. 
According to Keycloak [docs](https://www.keycloak.org/docs/latest/server_admin/index.html#rules-for-client-secret-rotation) the secret regeneration doesn't happen automatically, but rather needs to be triggered "manually". This PR would support client secret rotation directly from Terraform e.g. using the Time provider. This way you could have periodic secret rotation and Terraform would be able to manage that

PR #1287, opened 2025-08-29
feat: transform computed to default arguments  to match documentation

thomas-VIGINUM:
## Current problem
The `keycloak_openid_client` resource documentation states some arguments defaults, especially on the flow definition. However, those attributes are computed instead of defaulted. 

## Changes
Move some arguments from `Computed` to `Default`.

### DIFF
--- a/provider/resource_keycloak_openid_client.go
+++ b/provider/resource_keycloak_openid_client.go
@@ -106,22 +106,22 @@ func resourceKeycloakOpenidClient() *schema.Resource {
 			"standard_flow_enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Computed: true,
+				Default:  false,
 			},
 			"implicit_flow_enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Computed: true,
+				Default:  false,
 			},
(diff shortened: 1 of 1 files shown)


sschu:
What the documentation says is wrong because the provider is currently just using the Keycloak defaults - which has standard flow enabled. That's why the tests are failing.

thomasdarimont:
@thomas-VIGINUM thanks for your PR! Would you mind rebasing it to latest main and add your DCO (git commit -s ...)?

thomasdarimont:
With the change from this PR the OIDC client example configuration works correctly:
```
resource "keycloak_openid_client" "openid_client" {
  realm_id            = keycloak_realm.test.id
  client_id           = "test-client"

  name                = "test client"
  enabled             = true

  access_type         = "CONFIDENTIAL"
  valid_redirect_uris = [
    "http://localhost:8080/openid-callback"
  ]

  login_theme = "keycloak"

  extra_config = {
    "key1" = "value1"
    "key2" = "value2"
  }
}
```

thomasdarimont:
... just tried to run the tests locally, as @sschu already wrote this change breaks many tests. I think we have to think a bit more about how we can solve this.

Issue #1194, opened 2025-05-06
SMTP credentials are overwritten when a different attribute is set on realm level


dmpv-cyber:
### Describe the bug

1. Create a new realm with SMTP credentials defined.
2. Try to send an email, it works.
3. Then, create a new custom authentication flow and bind it with the resetCredentialsFlow, for example.
4. Deploy it with pulumi up.
5. If you try now to send an email it will fail because the SMTP credentials were overwritten.

### Version

26.2.2, 5.2.0 terraform-provider-keycloak

### How to Reproduce?

Create a new realm via terraform with the smtp_server attribute defined: 
```
resource "keycloak_realm" "test" {
  realm             = "test"
  enabled           = true
  display_name      = "Test"
  smtp_server {
    host = "host.com"
    from = "no-reply@pulumi.keycloak.com"
    ssl = true
    from_display_name = "test"
    auth {
      username = "username"
      password = "password"
    }
  }
}
```
Deploy it, and test the email sending, it should work.

Then, create a new custom authentication flow and bind it with the resetCredentialsFlow, for example:

```
resource "keycloak_authentication_flow" "test_reset_flow" {
  alias       = "resetCredentialsCopy"
  realm_id    = keycloak_realm.test.id
  description = "browser based authentication"
}
resource "keycloak_authentication_bindings" "test_flow_binding" {
  realm_id     = keycloak_realm.test.id
  reset_credentials_flow = keycloak_authentication_flow.test_reset_flow.alias
}
```
If you try now to send an email it will fail because the SMTP credentials were overwritten.

### Anything else?

This only started to happen after we are upgraded to keycloak version 26.2.0.
On that version they added Token based authentication for SMTP https://www.keycloak.org/docs/latest/release_notes/index.html#token-based-authenticatio.
Maybe it's related with this starting now to fail.

I noticed also another thing on the Admin events on the troublesome realm, on the update realm event the realm representation init has different attributes, on the smtpServer attribute, depending if it was updated via the admin console or via terraform, here is an example of the smtpServer attribute on those realm representations:

- Update via terraform:

```
 "smtpServer": {
    "auth": "true",
    "host": "host.com",
    "from": "no-reply@pulumi.keycloak.com",
    "fromDisplayName": "test",
    "ssl": "true",
    "user": "username",
    "password": "**********"
  },
```

- Update via admin console:

```
"smtpServer": {
    "password": "**********",
    "auth": "true",
    "host": "host.com",
[...]

Issue #1162, opened 2025-03-25
Setting configuration to sending Mails via SMTP and XOAUTH2 authentication mechanism


carlosrodovalho:
### Description

Since The Keycloak outgoing SMTP mail configuration currently supports basic authentication. Many service providers (Microsoft, Google) are moving towards SMTP OAuth authentication and end the support for basic authentication.

The following PR adds support for OAuth authentication for Keycloak on email services. It is scheduled to be relased on Keycloak 26.2.0:

https://github.com/keycloak/keycloak/pull/36665

It would be nice to be able to set this configuration use Terraform Keycloak's provider

### Motivation

In the current situation, is not possible to configure Keycloak to use the new feature using the provider

PR #1266, opened 2025-07-30
Fix the smtp server password based authentication broken on realm update

simonregn:
I don't want to get credits myself. I've copied your changes @lucdew to create this PR, so everybody can benefit from the changes you made. I didn't test it myself

PR #1293, opened 2025-09-03
Add OAUTH authentication option to SMTP servers

lucymcphail:
This patch leaves the `auth` argument to `smtp_server` as it is, but adds a new `token_auth` block to let users configure an auth token through Terraform.

Implements keycloak/terraform-provider-keycloak#1162

### DIFF
--- a/provider/resource_keycloak_realm_test.go
+++ b/provider/resource_keycloak_realm_test.go
@@ -151,6 +151,46 @@ func TestAccKeycloakRealm_SmtpServerUpdate(t *testing.T) {
 		},
 	})
 }
+func TestAccKeycloakRealm_SmtpServerOauth(t *testing.T) {
+	realm := acctest.RandomWithPrefix("tf-acc")
+	realmDisplayNameHtml := acctest.RandomWithPrefix("tf-acc")
+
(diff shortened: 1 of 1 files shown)


lucymcphail:
Looks like the `ConflictsWith` was actually working fine all along, I think something must have been doing something wrong with my testing. Should be ready to review now.

thomasdarimont:
Just saw the PR #1266, which overlaps a bit with this PR. How about integrating #1266 first and adding this PR after?

thomasdarimont:
Please ignore what I said in ... https://github.com/keycloak/terraform-provider-keycloak/pull/1293#issuecomment-3319619883 :)

Would it be possible to add 2 additional test-cases similar to the following examples?
- https://github.com/keycloak/terraform-provider-keycloak/blob/main/provider/resource_keycloak_realm_test.go#L114
- https://github.com/keycloak/terraform-provider-keycloak/blob/main/provider/resource_keycloak_realm_test.go#L135

thomasdarimont (approved):
Thanks for adding the tests. I just gave this a spin. Looks good to me :) 

I'll retrigger the build and merge if it passes.

This also fixes #1194.

PR #1295, opened 2025-09-03
Document client roles in default roles

ingokuba:
I saw alot of issues asking for the support of client roles in default roles. After some digging I found that it is indeed possible via #764. Unfortunately the documentation hints that only realm roles can be configured in `keycloak_default_roles`.


ingokuba:
I signed-off the commit as documented, but still GitHub complains :shrug:

ingokuba:
If someone could rerun the check - that'd be great :pray:

sschu:
LGTM, thanks!

Issue #1256, opened 2025-07-22
keycloak_realm_client_policy_profile_policy condition should accept array for roles


ar-or:
### Describe the bug

```
# PAR policy - applies PAR profile to clients 
resource "keycloak_realm_client_policy_profile_policy" "par_policy" {
  realm_id    = keycloak_realm.bestattungwien.id
  name        = "PAR policy for device-in-pool clients"
  description = "Enforces PAR for clients with device-in-pool scope"
  enabled     = true
  profiles    = [keycloak_realm_client_policy_profile.par_profile.name]

  condition {
    name = "client-roles"
    configuration = {
      "is-negative-logic" = "false"
      "roles"             = "requires-par-role"  #<-- THIS IS A BUG. SHOULD BE ARRAY
    }
  }

  depends_on = [
    keycloak_openid_client_scope.device_in_pool_scope
  ]
}
```

This results in 

```
╷
│ Error: error sending GET request to //admin/realms/somerealm/clients/41234533-34ee-4ed3-8b6e-7359aec02599: 500 Internal Server Error. Response body: {"error":"unknown_error","error_description":"Cannot parse the JSON"}
```

``` [org.keycloak.services.error.KeycloakErrorHandler] (executor-thread-288) Uncaught server error: java.lang.IllegalArgumentException: Cannot construct instance of `java.util.ArrayList` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('device-in-pool')
keycloak-1  |  at [Source: UNKNOWN; byte offset: #UNKNOWN] (through reference chain: org.keycloak.services.clientpolicy.condition.ClientRolesCondition$Configuration["roles"])
```

### Version

5.3.0

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

hishamanver:
@rhigdon happy to do so! 

Thanks for looking into this, keen as a bean for the fix :)

remcojansen:
I can confirm the observation shared by @rhigdon and linked a [new PR](#1338) that addresses this. The implementation is similar to PR #1315 with two differences as far as I can see:
* Added a test case to verify the content of the fields in condition configurations
* Using string checks to identify whether unmarshaling is to be applied to keep the solution more generic (a similar approach was used for the user profile resource)

PR #1298, opened 2025-09-04
[Issue-1256] Migrate Client Profiles and Client Policies into Realm Resource

rhigdon:
Since clientPolicies are stored as json on a per-realm basis it should
not be a first level resource.  Move this functionality into the realm
resource which will just marshal all of the policies and send it at once
to the server.

This will fix https://github.com/keycloak/terraform-provider-keycloak/issues/1256


rhigdon:
This fix does not appear to be stable.  It seems like the `configuration` needs to be a map but also needs to have its values encoded.  I am going to work towards this implementation.

PR #1314, opened 2025-09-15
Fix path for realm_localization documentation

ingokuba:
I noticed the path for resource `keycloak_realm_localization` was wrong.

image

Hope this fix is correct


ingokuba:
see #1123

thomasdarimont:
LGTM @sschu  I think we can merge this.

PR #1327, opened 2025-09-25
[Issue-1111] Move Realm Attributes to its own Resource

rhigdon:
this will help decouple the realms attributes and allow reuse in modules
 without relying on the entire realm configuration.


rhigdon:
> @rhigdon I think if we add this, we should offer both options as mutually exclusive. If we removed the current way, we would need a state migration and also ask everybody to adapt their configuration so this would be quite a breaking change.

That sounds good to me.  Thanks for the direction on this!

PR #1346, opened 2025-10-10
Feat/add deletion_protection to keycloak realm

tuankiettruong:
Added deletion_protection to prevent keycloak_realm resource from being deleted, even by removing resource from, which Terraform's life_cycle.prevent_destroy does not support


sschu (approved):
Thanks a lot!

Issue #1356, opened 2025-10-17
GitHub Issue - Cascade deletion of client roles in keycloak_generic_role_mapper


foch01:
## Version
- Keycloak Provider: v5.5.0 (latest tested version)
- Keycloak Version: Test in version 26.0.5. The issue is the same in 26.3.4 because the API takes an `optional body`.
- Terraform: 1.13.4

## Problem
Deleting a client role via a `keycloak_generic_role_mapper` resource causes the deletion of **all client roles** from the scope instead of just one. 

When client A has multiple roles from client B assigned to its scope, deleting one of them with Terraform causes the deletion of all other roles from the same scope.

This problem is identical to the one described in issue #838 which concerned realm roles and was fixed in version v4.1.0.

## Suspected Root Cause
According to the analysis of the provider's source code (`role_scope_mapping.go`), when deleting a client role from the scope, the Keycloak API expects to receive the roles in the request body as an array:

```
DELETE /{realm}/clients/{id}/scope-mappings/clients/{client}
```

However, in the `DeleteRoleScopeMapping` function, for client roles (`role.ClientRole = true`), the request is sent **without a body**:

```go
func (keycloakClient *KeycloakClient) DeleteRoleScopeMapping(ctx context.Context, realmId string, clientId string, clientScopeId string, role *Role) error {
	roleUrl := roleScopeMappingUrl(realmId, clientId, clientScopeId, role)
	if role.ClientRole {
		return keycloakClient.delete(ctx, roleUrl, nil) // ← Empty body (nil)
	} else {
		body := [1]RealmRoleRepresentation{
			// ... body with realm role data
		}
		return keycloakClient.delete(ctx, roleUrl, body)
	}
}
```

### 1. Terraform Configuration

```hcl
# Define a client that owns the roles
resource "keycloak_openid_client" "backend_client" {
  realm_id                 = "public"
  client_id               = "api-backend"
  access_type             = "CONFIDENTIAL"
  service_accounts_enabled = true
}

# Create multiple client roles
resource "keycloak_role" "backend_role_1" {
  realm_id    = "public"
  client_id   = keycloak_openid_client.backend_client.id
  name        = "customers:read"
}

resource "keycloak_role" "backend_role_2" {
  realm_id    = "public"
  client_id   = keycloak_openid_client.backend_client.id
  name        = "customers:write"
}

# Define a client scope
resource "keycloak_openid_client_scope" "app_client_roles" {
  realm_id               = "public"
[...]

PR #1357, opened 2025-10-17
Fix cascade deletion of client roles in keycloak_generic_role_mapper

foch01:
## Overview
This PR fixes a critical bug in the `DeleteRoleScopeMapping` function where deleting a single client role from a scope would inadvertently delete **all client roles** from that scope due to an empty request body being sent to the Keycloak API.

## Description
When using `keycloak_generic_role_mapper` to manage client roles in scopes, deleting one mapper resource would cause all other client role mappers in the same scope to be removed from Keycloak, even though Terraform only showed one resource being deleted.

### Root Cause
The issue was in the `DeleteRoleScopeMapping` function in `role_scope_mapping.go`. For client roles (`role.ClientRole = true`), the DELETE request was sent with an empty body (`nil`), while the Keycloak API expects the specific roles to be deleted in the request body as an array.

```go
// Before (buggy behavior)
if role.ClientRole {
    return keycloakClient.delete(ctx, roleUrl, nil) // ← Empty body causes cascade deletion
}
```

This contrasted with realm roles, which correctly included the role data in the request body.

## Solution
This PR standardizes the deletion behavior by ensuring both client roles and realm roles include the role data in the DELETE request body:

```go
// After (fixed behavior)
body := [1]RealmRoleRepresentation{
    {
        Id:          role.Id,
        Name:        role.Name,
        Description: role.Description,
        Composite:   role.Composite,
        ClientRole:  role.ClientRole,
        ContainerId: role.ContainerId,
    },
}
return keycloakClient.delete(ctx, roleUrl, body)
```

## Changes Made
- **Modified `DeleteRoleScopeMapping` function** in `role_scope_mapping.go`
- **Removed the conditional logic** that treated client roles differently from realm roles
- **Unified the deletion approach** to always include role data in the request body
- **Maintains backward compatibility** with existing configurations

## Testing
- ✅ Individual client role deletion now works correctly
- ✅ Other roles in the scope remain untouched
- ✅ Realm role deletion continues to work as before
- ✅ No breaking changes to existing functionality

### Before
- Deleting one client role mapper → All client roles removed from scope
- Required two `terraform apply` cycles to achieve intended result
- Unpredictable behavior and potential data loss

### After
- Deleting one client role mapper → Only that specific role is removed
- Single `terraform apply` achieves intended result
- Predictable and granular role management

## Backward Compatibility
This change is fully backward compatible:
- Existing Terraform configurations will work without modification
- No changes to resource schemas or attributes
- Only the internal API call behavior is corrected

## Related Issues
- Issue : https://github.com/keycloak/terraform-provider-keycloak/issues/1356

### Scenario 1: Single Role Deletion
```hcl
# Before: Commenting out one mapper deleted all mappers
# After: Commenting out one mapper deletes only that mapper
resource "keycloak_generic_role_mapper" "customers_read" { ... }
resource "keycloak_generic_role_mapper" "customers_write" { ... } # Deleted
```

### Scenario 2: Multiple Client Scopes
```hcl
# Before: Deletion in one scope affected other scopes unpredictably
# After: Deletion is isolated to the specific scope and role
```

## Verification Steps
1. Create multiple `keycloak_generic_role_mapper` resources for client roles
2. Apply the configuration
3. Remove one mapper from the configuration
4. Run `terraform plan` - should show only one deletion
5. Run `terraform apply` - should remove only the intended role
6. Verify in Keycloak admin console that other roles remain

## Files Changed
- `keycloak/role_scope_mapping.go` - Fixed DeleteRoleScopeMapping function

## Breaking Changes
None. This is a bug fix that corrects unintended behavior.

## Additional Notes
This fix brings consistency between client role and realm role deletion behavior, making the provider more predictable and reliable for managing Keycloak role scope mappings.


sschu:
@foch01 Are you still interested in landing this?

Issue #1267, opened 2025-07-31
Improve support for `import` of OIDC clients


Download:
### Description

Keycloak creates a bunch of default clients in `master` or new realms:

* `account`: Represents the backend for the account API
* `account-console`: Provides a frontend for user account management.
* `admin-cli`: Used for managing realms via CLI tools like kcadm.sh. 
* `broker`: Serves as the backend for identity brokering.
* `security-admin-console`: Manages the admin UI for a specific realm. 
* `realm-management`: Backend API for realm-specific management. Does not exist in `master`.

https://skycloak.io/blog/what-to-do-with-the-default-clients-in-your-newly-created-keycloak-realm-2/

Now it makes a lot of sense that we would want to manage these clients with Terraform. I want to delete some of these clients, but deleting such a resource seems to be nigh impossible via Terraform... The next best option then would be to disable them instead by setting `enabled = false`.... But before we can apply any config, we need Terraform to import the client into state. I see two ways of doing that and both seem to be completely crippled atm:

1. Use a Terraform `import` block
2. Use the `import = true` field on the `keycloak_openid_client` resource

Option 1 would be my preferred method since it is a Terraform-wide feature that in principle works for any supported resource, not just openid clients. However, this is completely crippled because of this:

> ### [Import](https://library.tf/providers/keycloak/keycloak/latest/docs/resources/openid_client#import)
> Clients can be imported using the format {{realm_id}}/{{client_keycloak_id}}, 
> where client_keycloak_id is the unique ID that Keycloak assigns to the client 
> upon creation. This value can be found in the URI when editing this client in 
> the GUI, and is typically a GUID.

Yeah, so that is completely pointless. We cannot predict what GUID a new client in a new (or default) realm will get so we can never automate anything this way. Why oh why can we not use the `admin-cli`, `broker` etc client ids here???

Maybe option 2 can save us? This second option is specific to the `keycloak_openid_client` resource and since we only have to specify `true` we neatly bypass the whole unpredictable ID is useless problem. So I tried that but my `plan` tells me it will create all these resources anyway, even though I specified `import = true`. So I did some searching and encountered this open issue:

[keycloak_oidc_client import unexpected behaviour #1007](https://github.com/keycloak/terraform-provider-keycloak/issues/1007)

So basically, the whole `plan` then `apply` process breaks when we use this option...

[...]

PR #1359, opened 2025-10-22
Revert back "import" property and replace it with native `import` block. Fixes (#870) (#1007) and unblocks (#1267)

### LINKED ISSUE #870: access_token_lifespan not rollback
Hello,
When I set "access_token_lifespan" in keycloak_openid_client to customize the token lifetime and perform a rollback, the provider does not perform the rollback with the default value.
It leaves the previous value set.

Default values:
image

Thanks

### LINKED ISSUE #1007: keycloak_oidc_client import unexpected behaviour

mikhail-putilov:
I did it to fix two issues:
fixes #870
fixes https://github.com/keycloak/terraform-provider-keycloak/issues/1007

And while I was working on it, it was natural to unblock this one: #1267

---

Reasoning:
I personally can't see a reason why `import` attribute is introduced. Not only it introduces bugs, but it is custom behavior which is not consistent with other providers. `import { id = ..., to = ... }` block is a native way to deal with default resources.
If I am wrong, I am open to hear your thoughts.

Imo, I need to deprecate `import` attribute according to https://developer.hashicorp.com/terraform/plugin/framework/deprecations instead of removing it. If you agree, I will rework this PR

Todo:

- [x] Add dedicated test for #870
- [x] Add dedicated test for #1007
- [x] Update documentation
- [x] Fix flaky tests (they fail probably because of timing issues)
- [x] DCO
- [ ] Get approval if reverting `import` attribute makes sense
- [x] Split commits into two parts (if required)


provider/resource_keycloak_role_test.go
```diff
@@ -843,10 +844,13 @@ func testKeycloakRole_importClientRole(name, clientId string) string {
...
 	realm = "%s"
 }
+import {
+	id = "${data.keycloak_realm.realm.id}/%s"
+	to = keycloak_openid_client.imported-client
+}
 resource "keycloak_openid_client" "imported-client" {
 	realm_id    = data.keycloak_realm.realm.id
 	client_id   = "%s"
-	import		= true
```
mikhail-putilov: `import		= true` is replaced with `import { ... }`

provider/resource_keycloak_role_test.go
```diff
@@ -903,20 +910,23 @@ resource "keycloak_role" "%s" {
 			importedClientIdRef := fmt.Sprintf("imported-client-%d", i)
 
 			importedRoles += fmt.Sprintf(`
```
mikhail-putilov: maybe I should split this long `fmt.Sprintf` into two parts

provider/resource_keycloak_openid_client.go
```diff
@@ -687,34 +648,51 @@ func resourceKeycloakOpenidClientUpdate(ctx context.Context, data *schema.Resour
...
 	realmId := data.Get("realm_id").(string)
 	id := data.Id()
 
-	return diag.FromErr(keycloakClient.DeleteOpenidClient(ctx, realmId, id))
+	err := keycloakClient.DeleteOpenidClient(ctx, realmId, id)
+	if err != nil {
+		return diag.Diagnostics{
+			diag.Diagnostic{
+				Severity: diag.Error,
+				Summary:  fmt.Sprintf("cannot delete openid client '%s'", data.Get("client_id")),
```
mikhail-putilov: Better error message in case of a failed destroy operation

provider/resource_keycloak_openid_client_test.go
```diff
@@ -753,20 +753,18 @@ func TestAccKeycloakOpenidClient_loginTheme(t *testing.T) {
...
-	t.Parallel()
-
 	resource.Test(t, resource.TestCase{
 		ProviderFactories: testAccProviderFactories,
 		PreCheck:          func() { testAccPreCheck(t) },
 		CheckDestroy:      testAccCheckKeycloakOpenidClientNotDestroyed(),
 		Steps: []resource.TestStep{
 			{
 				Config:      testKeycloakOpenidClient_import("non-existing-client", true),
-				ExpectError: regexp.MustCompile("Error: openid client with name non-existing-client does not exist"),
```
mikhail-putilov: previous error message was not consistent with KeyCloak glossary: it must be "Client ID", not "name"

provider/resource_keycloak_openid_client_test.go
```diff
@@ -2095,14 +2144,50 @@ func testKeycloakOpenidClient_import(clientId string, enabled bool) string {
...
 	access_type = "PUBLIC"
 	root_url    = ""
 	enabled     = %t
-	import      = true
 }
-	`, testAccRealm.Realm, clientId, enabled)
+	`, testAccRealm.Realm, clientId, clientId, enabled)
+}
+
+func testKeycloakOpenidClient_import_postcondition(clientId string, enabled bool) string {
```
mikhail-putilov: an example is taken from https://github.com/keycloak/terraform-provider-keycloak/issues/1007#issuecomment-2432989685

provider/resource_keycloak_openid_client_test.go
```diff
@@ -339,6 +339,29 @@ func TestAccKeycloakOpenidClient_AccessToken_basic(t *testing.T) {
 	})
 }
 
+func TestAccKeycloakOpenidClient_AccessToken_removed(t *testing.T) {
```
mikhail-putilov: a test for https://github.com/keycloak/terraform-provider-keycloak/issues/870

sschu:
@mikhail-putilov Thanks for this contribution and sorry for not reacting earlier!

Regarding the `import` attribute: it was added in 2022 while Terraform added support for the `import` block in 2023 only. Therefore, I would be fine in removing it but it would be s a breaking change. Therefore, we should rather deprecate it and remove it only with a new major release of the provider.

I am also a bit reserved about the changes in the schema from `Computed` to `Default`. Can you elaborate why this is necessary? In general `Computed` just takes the default from Keycloak which I would prefer over having a separate set of default values in the provider that might or might not match what Keycloak is using as default in different versions.

sschu:
@mikhail-putilov Are you still interested in landing this? Otherwise I would go ahead and close it since there are a few conflicts.

mikhail-putilov:
No, sorry. I have moved on.

PR #1365, opened 2025-11-03
Add support of dynamic scopes to openid client scope

### LINKED ISSUE #1297: add support of dynamic scopes to openid_client_scope
### Description

# Summary
In the keycloak UI I can toggle a client scope to become a dynamic scope.
However, the provider does not offer this option.
Being able to manage dynamic scopes would save us manual work in the setup process of new environments.

### Discussion

_No response_

### Motivation

We use dynamic scopes in our setup and without this option we have to manually create dynamic scopes for each new environment we spawn. This complicates the process and also prevents us from completely automating our infra setup with terragrunt / tofu, because this one manual tasks breaks the provisioning.

### Details

I'm just a user in that matter, but maybe you could add a boolean that can mark a client scope "dynamic".

Redestros:
## Description

Adds support for dynamic client scopes to the `keycloak_openid_client_scope` resource and data source. Dynamic scopes allow parameterized scope values that follow a pattern, enabling fine-grained authorization scenarios.

This feature enables clients to request scopes like `resource:read`, `resource:write`, `api:123`, etc., based on defined patterns.

## Related Issue

Closes #1297

### Resource Changes
- Added `dynamic` field to enable dynamic scope functionality
- Added `dynamic_scope_regexp` field to specify custom regex patterns for scope validation
- Updated validation logic to handle both default patterns and custom regex patterns

### Data Source Changes
- Added `dynamic` and `dynamic_scope_regexp` fields to the `keycloak_openid_client_scope` data source
- Ensures dynamic scope attributes are properly read and exposed

### Implementation Details
- Dynamic scope attributes are stored in Keycloak's `attributes` object as `is.dynamic` and `dynamic.scope.regexp`
- Added `syncToAttributes()` and `syncFromAttributes()` helper methods to manage data sync between Terraform state and Keycloak API
- Fixed `ListOpenidClientScopesWithFilter` to properly sync attributes when reading scopes

### Testing
- Added comprehensive unit tests for dynamic scope validation
- Added validation tests for both default and custom patterns
- Updated data source tests to verify dynamic scope fields

### Documentation
- Added example usage for dynamic client scopes
- Documented the requirement to enable `KC_FEATURES=dynamic-scopes:v1` when starting Keycloak

## Example Usage

```hcl
resource "keycloak_openid_client_scope" "dynamic_resource_scope" {
  realm_id               = keycloak_realm.realm.id
  name                   = "resource"
  description            = "Dynamic scope for resource-level permissions"
  dynamic                = true
  dynamic_scope_regexp   = "resource:.*"
  include_in_token_scope = true
}
```
This allows clients to request scopes like resource:read, resource:write, resource:123, etc.

Prerequisites
Dynamic scopes require Keycloak to be started with the feature flag enabled:
KC_FEATURES=dynamic-scopes:v1
Testing
 Unit tests pass
 Manual testing with local Keycloak instance
 Verified dynamic scopes are created correctly in Keycloak
 Verified data source reads dynamic scope attributes
 Verified update operations work correctly

Checklist
 Code follows project conventions
 Tests added/updated
 Documentation updated

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -101,6 +101,9 @@ jobs:
             EXTRA_HTTPS_KEY="-e KC_HTTPS_CERTIFICATE_KEY_FILE=/opt/keycloak/testdata/tls/server-key.pem"
             EXTRA_MTLS_CERTS="-e KC_TRUSTSTORE_PATHS=/opt/keycloak/testdata/tls/ca-cert.pem,/opt/keycloak/testdata/tls/client-cert.pem"
           fi
+          # Dynamic scopes feature introduced in Keycloak 17.0.0 (Feb 2022)
+          # All tested versions (22+) support it, so always enable
+          EXTRA_FEATURES="${EXTRA_FEATURES},dynamic-scopes:v1"
 
           docker run -d --name keycloak \
           -p 8080:8080 \
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
     - KC_DB_USERNAME=keycloak
     - KC_DB_PASSWORD=password
     - KC_LOG_CONSOLE_COLOR=true
-    - KC_FEATURES=preview,admin-fine-grained-authz:v1
+    - KC_FEATURES=preview,admin-fine-grained-authz:v1,dynamic-scopes:v1
     - KC_HTTPS_CERTIFICATE_FILE=/opt/keycloak/testdata/tls/server-cert.pem
     - KC_HTTPS_CERTIFICATE_KEY_FILE=/opt/keycloak/testdata/tls/server-key.pem
 # Enable for remote java debugging
--- a/provider/provider_test.go
+++ b/provider/provider_test.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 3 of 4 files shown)


sschu (changes requested):
@Redestros Thank you for your contribution! In general, it looks good. However, I think you did not capture the validation rules Keycloak has for dynamic scopes. Looking at https://github.com/keycloak/keycloak/blob/f39a37d8d168ed42052fe37a531b69ad96b6bfd9/services/src/main/java/org/keycloak/services/resources/admin/ClientScopeResource.java#L204-L233 , the rules are much tighter: A scope is only a valid dynamic scope if it is marked as dynamic and if it has a valid wildcard expression (not a regexp in contrast to what the parameter name suggests) with exactly a single wildcard. This means a lot of test cases are actually not valid. I would rather like to see an integration test that demonstrates creation of a dynamic client scope.

sschu (changes requested):
@Redestros Sorry for the delay and thanks for incorporating my remarks. It looks mostly fine now. The only thing I don't like is the testing only when the dynamic scopes feature is enabled. The first problem is that when the tests run concurrently the pre-check interferes with each other since they all create the same test scope. And in general, I would rather like to always have this tested. 

So please remove all the test code related to checking if the feature is enabled and instead just enable it in `docker-compose.yml', 'docker-compose-mtls.yaml' and in`test.yaml`  for versions that support it. You can add it to the `EXTRA_FEATURES`similar to how it is done for `admin-fine-grained-authz:v1`. Also the integration tests should only run for versions that support it.

PR #1378, opened 2025-11-10
Add shared mutex to KC Client and enforce locking for Realm Client Profiles and Policies

rhigdon:
Currently, when creating more than one Client Profile or Client Policy for a realm there is a race condition. Both resources try to update all policies or profiles for a realm.. so there becomes a race condition and only one is updated. This will ensure that things are added serially.

Adds an additional profile and policy to the example realm
adds a shared mutex to use with the Keycloak Client
use this mutex when creating/updating/deleting realm client policies and profiles


rhigdon:
@sschu Please review when you can.  I have added an additional lock for delete.  I tested locally by adding 3 distinct policies then trying to remove just one and ensuring the other two exist.  I did not add any unit tests, but I added an additional profile and policy to the integration tests which will test the creation.

PR #1385, opened 2025-11-12
fix: Use normalized Keycloak API path when querying group by path

### LINKED ISSUE #1379: problems with normalized paths in keycloak 24.4.3+
### Describe the bug

Upgrading keycloak to 26.4.3 and following 26.4.4 leads to problems with normalized paths as described within https://www.keycloak.org/docs/latest/upgrading/index.html#migrating-to-26-4-3 as:
> Previously Keycloak accepted HTTP requests with paths containing double dots (..) or double slashes (//). When processing them, it normalized the path by collapsing double slashes and normalized the path according to RFC3986. As this has led to a hard-to-configure URL filtering, for example, in reverse proxies, the normalization is now disabled, and Keycloak responds with an HTTP 400 response code.

So far i used keycloak 26.3.2 without any problems but after migrating to 26.4.4 i faced the exception:
```
│ Error: error sending GET request to /admin/realms/<realm>/group-by-path//Piveau: 400 Bad Request. Response body: {"error":"missingNormalization","error_description":"Request path not normalized"}
```

My group definition looks as:
```
resource "keycloak_group" "piveau" {
  realm_id = var.keycloak_realm
  name     = "Piveau"
}
```

A workaround i'm currently using is to define the extraEnvVars for my keycloak deployment of:
```
{
          name  = "KC_HTTP_ACCEPT_NON_NORMALIZED_PATHS"
          value = "true"
}
```

### Version

26.4.4

### Expected behavior

Trailing double slashes are not used by terraform provider keycloak and no additional flag is required.

### Actual behavior

exception:
```
│ Error: error sending GET request to /admin/realms/<realm>/group-by-path//Piveau: 400 Bad Request. Response body: {"error":"missingNormalization","error_description":"Request path not normalized"}
```

My group definition looks as:
```
resource "keycloak_group" "piveau" {
  realm_id = var.keycloak_realm
  name     = "Piveau"
}
```

### How to Reproduce?

Setup the group with a previous version older than 26.4.3 like:
```
resource "keycloak_group" "piveau" {
  realm_id = var.keycloak_realm
  name     = "Piveau"
}
```

upgrade keycloak to latest version (26.4.4 as of today) and run tofu plan again which fails with above exception.

### Anything else?

_No response_

yrd:
Closes #1379. Newer Keycloak versions [no longer accept](https://www.keycloak.org/docs/latest/upgrading/index.html#migrating-to-26-4-3:~:text=Accepting%20only%20normalized%20paths%20in%20requests) paths like `/admin/realms/example/group-by-path//Management` (for a group named _Management_ directly underneath the root).

Not quite sure if there are any other places that could benefit from normalization, but the group-by-path API seems to be the biggest low hanging fruit. Some quick grepping didn't yield any other results – I couldn't anything else that references things through a `/`-delimited path scheme instead of IDs.

### DIFF
--- a/keycloak/group.go
+++ b/keycloak/group.go
@@ -30,7 +30,7 @@ func (keycloakClient *KeycloakClient) groupParentId(ctx context.Context, group *
 
 	var parentGroup Group
 
-	err := keycloakClient.get(ctx, fmt.Sprintf("/realms/%s/group-by-path/%s", group.RealmId, parentPath), &parentGroup, nil)
+	err := keycloakClient.get(ctx, fmt.Sprintf("/realms/%s/group-by-path/%s", group.RealmId, strings.TrimPrefix(parentPath, "/")), &parentGroup, nil)
 	if err != nil {
 		return "", err
 	}


marko-modsching-bison:
I've the same Problem using keycloak 26.4.4 and keycloak provider 5.5.0. When will the fix be available in the registry? Which version will it be?

yrd:
It's not currently released yet. Until a new version is available, you can build the provider yourself and pick in https://github.com/keycloak/terraform-provider-keycloak/commit/31116ff9d3a5ec8210dfdedc5f0d81a1ebf02ed6.

Issue #1396, opened 2025-11-26
keycloak_group_permissions disabling permissions causes panic: interface conversion: interface {} is nil, not string


consooo:
### Describe the bug

when creating a `keycloak_group_permissions` and disabling permissions via the ui, the following `terraform plan` will throw this error:

```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Plugin did not respond
│ 
│   with keycloak_group_permissions.test_group_permission,
│   on main.tf line 56, in resource "keycloak_group_permissions" "test_group_permission":
│   56: resource "keycloak_group_permissions" "test_group_permission" {
│ 
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ReadResource call. The plugin logs may contain more details.
╵

Stack trace from the terraform-provider-keycloak_v4.4.0 plugin:

panic: interface conversion: interface {} is nil, not string

goroutine 14 [running]:
github.com/mrparkers/terraform-provider-keycloak/provider.resourceKeycloakGroupPermissionsRead({0x1012408c8, 0x14000380660}, 0x0?, {0x101231fa0?, 0x1400010cc00})
        github.com/mrparkers/terraform-provider-keycloak/provider/resource_keycloak_group_permissions.go:138 +0x9b0
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0x1400051b260, {0x101240900, 0x140004920f0}, 0xd?, {0x101231fa0, 0x1400010cc00})
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:724 +0xe8
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).RefreshWithoutUpgrade(0x1400051b260, {0x101240900, 0x140004920f0}, 0x140003b2b60, {0x101231fa0, 0x1400010cc00})
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:1015 +0x468
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ReadResource(0x140004982b8, {0x101240900?, 0x14000595fb0?}, 0x1400011c800)
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/grpc_provider.go:613 +0x400
[...]

PR #1397, opened 2025-11-26
fix keycloak_group_permissions - panic: interface conversion: interface {} is nil, not string

consooo:
fixes https://github.com/keycloak/terraform-provider-keycloak/issues/1396

the fix is similar to https://github.com/keycloak/terraform-provider-keycloak/pull/591/files

### DIFF
--- a/keycloak/group_permissions.go
+++ b/keycloak/group_permissions.go
@@ -10,11 +10,11 @@ type GroupPermissionsInput struct {
 }
 
 type GroupPermissions struct {
-	RealmId          string                 `json:"-"`
-	GroupId          string                 `json:"-"`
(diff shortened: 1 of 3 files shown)


sschu:
@consooo If you squash, you should have only one commit that you can sign-off. I still see 13 commits in this PR and most of them are not yours, so they probably got merged from the main. Can you try rebasing on main instead of merging?

consooo:
> @consooo Hi, no problem. Tell me what I need to fix?

@vaerh sorry, I merged instead of rebased. It's fixed.

sschu (approved):
LGTM, thanks!

PR #1457, opened 2026-02-05
added support for SAML want_authn_request_signed manual override

dscham:
After having trouble with trying to apply a converted realm export JSON I found that there is a hard coded default behaviour for `wan_authn_requests_signed`. 

Having it set to true when `signature_algorithm` is set, may be a sensible default. So to give admins the choice, I kept the default behaviour but added a check for a manual override.


dscham:
commit mistake

PR #1458, opened 2026-02-05
added support for SAML want_authn_request_signed manual override

dscham:
After having trouble with trying to apply a converted realm export JSON I found that there is a hard coded default behaviour for want_authn_requests_signed.

Having it set to true when signature_algorithm is set, may be a sensible default. So to give admins the choice, I kept the default behaviour but added a check for a manual override.

Closes: #1459


dscham:
Force Push commit due to following DCO guide to fix it's check.

sschu (changes requested):
Thanks for your PR @dscham! It looks good to me, the only thing I would ask you is to also update the documentation? See https://github.com/keycloak/terraform-provider-keycloak/blob/c657989309c179eaf9866bed24e8b3cbe5e717ab/docs/resources/saml_identity_provider.md

dscham:
Thank you for the feedback @sschu. I added the documentation for the now accessible attribute and it's default behaviour.

sschu (approved):
LGTM, thanks!

PR #1496, opened 2026-03-05
Only set want_authn_requests_signed to false when signature_algorithm is not explicitly set

RichieB2B:
The logic in for setting `want_authn_requests_signed` based on `signature_algorithm` in #1458 is flawed and was not checked in a test. We got bitten by this after upgrading to 5.7.0 we suddenly had `want_authn_requests_signed` set to `false` even though we explicitly set `signature_algorithm`.

This PR adds a test and fixes the flawed logic.


sschu (approved):
LGTM, thanks!

PR #1499, opened 2026-03-05
fix: use realm internal UUID as parentId in keystore component updates

### LINKED ISSUE #1368: Updating imported realm key causes Terraform to effectively delete it
### Describe the bug

Updating the key works correctly in realms created by terraform, where `keycloak_realm`'s `id`, `realm` and `internal_id` are all equal. However, when working with a realm that was created manually in the admin console and imported to terraform, `id` and `realm` fields correspond to the realm's name, and `internal_id` is a GUID generated by Keycloak. 

In such cases [realm_keystore_rsa_generated.go:46](github.com/keycloak/terraform-provider-keycloak/blob/959998a6b418b5ca00b183aa518f89e7678d9c18/keycloak/realm_keystore_rsa_generated.go#L46) uses the realm's name as `parentId` of Keycloak's component object in a PUT request, which causes Keycloak to change this field of the key.

After apply, the key is no longer visible in the admin console, because the key is not associated with any realm.

### Version

26.3.4

### Expected behavior

The `parentId` sent in requests to `/realms/{realm}/components/{id}` should be equal to realm's internal ID

The key is correctly updated, as stated by `terraform plan`

### Actual behavior

The `parentId` sent in requests to `/realms/{realm}/components/{id}` is realm's name

The key is effectively deleted from the realm, despite `terraform plan` displaying an in-place update

### How to Reproduce?

1. Add a new realm in Keycloak's admin console (for example "TEST_REALM")
2. Declare the realm in terraform 
```tf
resource "keycloak_realm" "test_realm" {
  realm = "TEST_REALM"
  enabled           = true
}
```
3. Import the realm
```sh
terraform import keycloak_realm.test_realm TEST_REALM
```
4. Declare the rsa-generated realm key resource
```tf
resource "keycloak_realm_keystore_rsa_generated" "signing_key_original" {
  name      = "rsa-generated"
  realm_id  = keycloak_realm.test_realm.id  

  enabled = true
  active  = true

  priority  = 100
}
```
5. Import the key
```sh
terraform import keycloak_realm_keystore_rsa_generated.signing_key_original TEST_REALM/{id of the key}
```
6. Update the resource in terraform, for example set the priority to 101.
7. `terraform plan` shows that the resource will be updated in-place
8. [Optional] Set a debugging breakpoint in [ComponentResource.java:179](https://github.com/keycloak/keycloak/blob/d5763b9c0badc570d6c5a10baa5f6d170922330a/services/src/main/java/org/keycloak/services/resources/admin/ComponentResource.java#L179) in Keycloak's source code
9. After `terraform apply`, the key is no longer visible in the admin console

### Anything else?

_No response_

agomerz:
When a realm is created manually in the Keycloak admin console and imported into Terraform, keycloak_realm.id holds the realm name (e.g. "TEST_REALM"), not the internal UUID that Keycloak assigns. All keystore resources were setting component.parentId = realm_name on create and update. Keycloak requires parentId to be the realm's internal UUID; when it receives the realm name instead it silently re-associates the component, causing the key to disappear from the admin console after apply.

Fix: fetch the realm object via GetRealm in each keystore resource's Create and Update functions to obtain the internal UUID (realm.Id), and pass it as a separate ParentId field on the keystore struct. The convertFrom*ToComponent helpers now prefer ParentId over RealmId when building the component payload, while the URL path continues to use the user-supplied RealmId (realm name or UUID) which Keycloak accepts for routing. This matches the existing pattern used by keycloak_ldap_user_federation.

Fixes #1368


sschu (changes requested):
@agomerz The PR looks good, thanks for your contribution. However, to make CI happy you need to add a sign-off to your commit. Easiest would probably to squash them all and then add a sign-off using `git commit -s --amend`.

sschu:
@agomerz Are you still interested in landing this?

Issue #1444, opened 2026-01-21
Unable to provision fine-grained permissions with group-policies on specific resources and built-in scopes


hkailantzis:
### Describe the bug

In an attempt to make use of the new fine-grained permissions feature, we encountered the following issues. Minimal TF code samples below.

### Versions

`Keycloak 26.3.3, 26.5.1`

`OpenTofu v1.10.2, v1.11.3`

`keycloak/keycloak v5.6`

Simplified TF code sample to serve as example:

```
resource "keycloak_realm" "test" {
  realm   = "my-realm"
  admin_permissions_enabled = true
}

resource "keycloak_group" "test" {
  realm   = keycloak_realm.test.id
  name    = "SUPPORT"
}

data "keycloak_openid_client" "admin_permissions" {
  realm_id  = keycloak_realm.test.id
  client_id = "admin-permissions"
}

resource "keycloak_openid_client_group_policy" "test" {
  realm_id           = keycloak_realm.test.id
  resource_server_id = data.keycloak_openid_client.admin_permissions.id
  name               = "test"
  decision_strategy  = "UNANIMOUS"
  description        = "Test Group Policy"
  logic              = "POSITIVE"

  groups {
    id              = keycloak_group.test.id
    path            = keycloak_group.test.path
    extend_children = false
  }
}

# block below throws 400 Bad Request error after TF apply.
resource "keycloak_openid_client_authorization_permission" "groups" {
  realm_id           = keycloak_realm.test.id
  resource_server_id = data.keycloak_openid_client.admin_permissions.id
  name               = "test-group"
  type               = "scope"
  policies           = [keycloak_openid_client_group_policy.test.id]
  resources          = [keycloak_group.test.id]
  scopes = [view", "view-members", "manage-members", "manage-membership", "manage-group-membership"]
}

# TF block below works. Note that all subsequent TF plan/apply runs, show that resourcesIDs are getting removed, even though the resources field is not even allowed to be declared here.

resource "keycloak_openid_client_authorization_permission" "all_users" {
  realm_id           = keycloak_realm.test.id
  resource_server_id = data.keycloak_openid_client.admin_permissions.id
  name               = "test-all-users"
  type               = "scope"
  policies           = [keycloak_openid_client_group_policy.test.id]
  resource_type      = "Users"
  scopes = [view", "view-members", "manage-members", "manage-membership", "manage-group-membership"]
}
```
### **Discussion**

[...]

PR #1517, opened 2026-03-16
Add keycloak_openid_client_authorization_permission_scope resource

wimski:
This PR currently has a copy of `keycloak_openid_client_permission` suffixed with 'scope'. This makes creating authz v2 FGAP permissions possible.

This PR is incomplete. Anyone able to help this go through?
* Missing tests.
* Better/other naming of the resource?
* Is this resource releasable by itself or is it part of a larger set?

I personally have no experience with Go or with this specific project.

Related to https://github.com/keycloak/terraform-provider-keycloak/issues/1444

### DIFF
--- a/keycloak/openid_client_authorization_permission_scope.go
+++ b/keycloak/openid_client_authorization_permission_scope.go
@@ -1,0 +1,90 @@
+package keycloak
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+type OpenidClientAuthorizationPermissionScope struct {
+	Id               string   `json:"id,omitempty"`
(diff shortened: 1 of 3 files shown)


michalvavrik:
> @wimski Thanks for your effort! I will need some time to figure out how we best support the transition from v1 to v2 so I ask for some patience.

if that is of interest: once you have some high-level idea how you want to do that (concept), I can help out with implementation

sschu:
@wimski FGAPv2 support has now been added to the provider so I would go ahead and close this PR. Do you think anything is still missing? If yes, feel free to comment...

PR #1518, opened 2026-03-16
Migrated to terraform-plugin-testing

sschu:


thomasdarimont (approved):
Looks like a big but mostly mechanical change. I'm fine with it as it prepares for a larger migration to the new terraform framework.

LGTM

Issue #1187, opened 2025-04-17
Support Fine-Grained Permissions v2 (Keycloak 26.2)


sventorben:
### Description

Keycloak 26.2 introduced Fine-Grained Permissions v2, which allows scoped and secure delegation of administrative rights across various Keycloak entities such as users, groups, roles, and clients. This improvement enables more secure and granular administrative delegation compared to the legacy realm admin role.

Currently, the Keycloak Terraform Provider does not support managing Fine-Grained Permissions v2. Supporting this feature would enable full infrastructure-as-code (IaC) workflows around admin delegation and access management.

Relevant documentation:  
- https://www.keycloak.org/docs/latest/server_admin/index.html#_fine_grained_permissions

### Motivation

- Enables secure delegation of administrative tasks using Terraform.
- Aligns Terraform-based Keycloak management with the latest Keycloak features.
- Reduces reliance on the full realm admin role for delegated use cases.
- Improves compliance and security through scoped permissions and policies.

### Details

**Requested Enhancements**

#### Realm-Level Support
- Support the `adminPermissionsEnabled` flag for realms.

#### Top-Level Permission Resources
Add new Terraform resources for managing entity-specific admin permissions:
- `keycloak_user_permissions`
- `keycloak_group_permissions`
- `keycloak_client_permissions`
- `keycloak_role_permissions`

Each should allow configuring which scopes (e.g. `view-users`, `manage-clients`) are protected and who is allowed to use them.

#### Policy Resources
Add support for managing policy types used in permissions:
- `keycloak_aggregated_policy`
- `keycloak_client_policy`
- `keycloak_client_scope_policy`
- `keycloak_group_policy`
- `keycloak_regex_policy`
- `keycloak_role_policy`
- `keycloak_time_policy`
- `keycloak_user_policy`

#### Permission–Policy Links
Add a resource like `keycloak_permission_policy_link` to associate policies with specific permission resources.

**Example Terraform Configuration**

```hcl
resource "keycloak_realm" "myrealm" {
  realm                     = "myrealm"
  enabled                   = true
  admin_permissions_enabled = true
}

resource "keycloak_role" "user_admin" {
  realm_id = keycloak_realm.myrealm.id
  name     = "user-admin"
}

resource "keycloak_user_permissions" "manage_users_permission" {
  realm_id = keycloak_realm.myrealm.id
  scope    = "manage-users"
  resource = "users"
}

[...]

PR #1519, opened 2026-03-16
Fix disabling of admin permissions in fine grained authz v2

### LINKED ISSUE #1509: keycloak_realm: admin_permissions_enabled is not disabled if previously enabled
### Describe the bug

If admin_permissions_enabled was previously set to true and you attempt to disable it again, this will not be done.

### Version

26.5.4

### Expected behavior

Admin permissions should be disabled.

### Actual behavior

Admin permissions remain enabled

### How to Reproduce?

Create realm with "admin_permissions_enabled = true".
Then use "admin_permissions_enabled = false". 

### Anything else?

_No response_

michalvavrik:
* Closes: https://github.com/keycloak/terraform-provider-keycloak/issues/1509

* Makes sure that `admin_permissions_enabled` is not omitted from PUT when it is explicitly set in `.tf` resource. It is done by distinguishing between true (set explicitly to true), false (set explicitly to false) and nil (unset, hence omitted) I think this might not be the only case, e.g. `organizations_enabled` looks like it behaves same way, but I didn't explore that as I am only fixing the linked issue.

* Conditionally switches `admin-fine-grained-authz` from `v1` to `v2` in docker compose and test CI so that we can test this PR. Currently `admin_permissions_enabled` is not tested because even by the existing test. It is ignored due to the fact that tests are using `admin-fine-grained-authz:v1`. However v1 has been deprecated since Keycloak 26.5.0 https://github.com/keycloak/keycloak/issues/44121. We should test it and make sure that the flag works.

* For version lesser than 26.5.0, 7 tests are disabled due to the fact that v2 does not support these scenarios:
  * https://github.com/keycloak/keycloak/blob/1ca15660f576eec7a81e60a00651d551d5045def/services/src/main/java/org/keycloak/services/resources/admin/ClientResource.java#L727
  * https://github.com/keycloak/keycloak/blob/1ca15660f576eec7a81e60a00651d551d5045def/services/src/main/java/org/keycloak/services/resources/admin/fgap/ClientPermissionsV2.java#L185

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,9 +88,13 @@ jobs:
           MOUNT_FEDERATION_EXAMPLE_VOLUME=""
           EXTRA_FEATURES=""
           MOUNT_FEDERATION_EXAMPLE_VOLUME="-v $PWD/custom-user-federation-example/build/libs/custom-user-federation-example-all.jar:/opt/keycloak/providers/custom-user-federation-example-all.jar:z"
-          if [[ "${{ matrix.keycloak-version }}" == "26.5.5" || "${{ matrix.keycloak-version }}" == "26.4.7"|| "${{ matrix.keycloak-version }}" == "26.3.5" || "${{ matrix.keycloak-version }}" == "26.2.5" ]]; then
+          if [[ "${{ matrix.keycloak-version }}" == "26.5.5" ]]; then
+            EXTRA_FEATURES=",admin-fine-grained-authz:v2"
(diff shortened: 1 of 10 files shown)


michalvavrik:
\cc @sschu

michalvavrik:
Looking at the https://github.com/keycloak/terraform-provider-keycloak/issues/1187 it might not be desirable to switch to v2 for now, however the bug is legit and the `admin_permissions_enabled` is documented :man_shrugging:

michalvavrik:
I can look into https://github.com/keycloak/terraform-provider-keycloak/issues/1187 as a follow-up... actually there is https://github.com/keycloak/terraform-provider-keycloak/pull/1517 in progress so maybe it is alright after all.

PR #1520, opened 2026-03-17
fix: sort multivalue attribute values on keycloak_group resource before diff

danielnickels:
The order of groups multivalue attribute values returned from the Keycloak API is not stable, especially after changes.

This PR amends what @michalvavrik has implemented for `keycloak_user` resource to the `keycloak_group` resource. Prior to the diff, the values are sorted. This avoids an update of the resource because of the value order.

The const `MULTIVALUE_ATTRIBUTE_SEPARATOR` is used in other locations as well. Not sure, if this ordering issue may also occur for `keycloak_role`, `keycloak_organization` or `keycloak_custom_user_federation`.


sschu (approved):
LGTM, thanks!

provider/resource_keycloak_group_test.go
```diff
@@ -86,6 +86,69 @@ func TestAccKeycloakGroup_createAfterManualDestroy(t *testing.T) {
...
+		"role-2",
+		"role-3",
+		"role-4",
+		"role-5",
+		"role-6",
+		"role-7",
+		"role-8",
+		"role-9",
+		"role-10",
+	}, keycloak.MULTIVALUE_ATTRIBUTE_SEPARATOR)
```
sschu:
```suggestion
	}, MULTIVALUE_ATTRIBUTE_SEPARATOR)
```

PR #1521, opened 2026-03-17
Fix using of computed values for client_secret_wo by allowing values unknown during the validation phase

### LINKED ISSUE #1438: keycloak_oidc_identity_provider client_secret_wo does not work when using value from module output
### Describe the bug

Using the resource keycloak_oidc_identity_provider without client_secret but with client_secret_wo and client_secret_wo_version is working fine when providing hardcoded values but fails with an error when using variables or other module's outputs.

### Version

26.4.2 (does not matter here), Keycloak Terraform Provider  5.6.0

### Expected behavior

No error but same behaviour as with keycloak_openid_client and as described in the [docs](https://registry.terraform.io/providers/keycloak/keycloak/latest/docs/resources/oidc_identity_provider#client_secret_wo-2).

When client_secret_wo_version is changed then client_secret_wo is used to update client secret.

### Actual behavior

```
Error: Required attribute not set
│
│   with keycloak_oidc_identity_provider.master,
│   on main.tf line xxx, in resource "keycloak_oidc_identity_provider" "master":
│  xxx: resource "keycloak_oidc_identity_provider" "master" {
│
│ The attribute [{{} client_secret}] is required when none of the following attributes are specified: [[{{} client_secret_wo}] [{{} client_secret_wo_version}]]
```

### How to Reproduce?

Create files:
```terraform
#./modules/testmodule/outputs.tf
output "testoutput" {
  value = "testvalue"
}
```
```terraform
#./main.tf
module "mytestmodule" {
  source = "./modules/testmodule"
}

resource "keycloak_oidc_identity_provider" "master" {
  realm        = "master"
  alias        = "whatever"

  client_id = "master"
  client_secret_wo         = module.testmodule.testoutput # this would be something like data.azurerm_key_vault_secret.my_secret.value
  client_secret_wo_version = module.testmodule.testoutput # this would be something like data.azurerm_key_vault_secret.my_secret.version

  # or even using a variable causes the error. var.environment is of type string
  # client_secret_wo = var.environment != "" ? "hi" : "bye" # no matter the value of var.environment, the if and the else branch always return a value
  # client_secret_wo_version = var.environment != "" ? 1 : 2

  authorization_url = "http://localhost:8080/auth/realms/master/protocol/openid-connect/auth"
  token_url         = "http://localhost:8080/auth/realms/master/protocol/openid-connect/token"
  extra_config = {
    "clientAuthMethod" = "client_secret_post"
  }
}
```
run `terraform init` and `terraform plan`

### Anything else?

I just started using Terraform and never used Go before.

But I noticed, that the resource keycloak_oidc_identity_provider is the only resource which is (on [resource_keycloak_oidc_identity_provider.go#L146](https://github.com/keycloak/terraform-provider-keycloak/blob/v5.6.0/provider/resource_keycloak_oidc_identity_provider.go#L146)) using the util [requiredWithoutAll](https://github.com/keycloak/terraform-provider-keycloak/blob/v5.6.0/provider/utils.go#L134) which produces the error message on [utils.go#L161](https://github.com/keycloak/terraform-provider-keycloak/blob/v5.6.0/provider/utils.go#L161).

As I said, I never used Go, but maybe this util is not resolving references to values from data, variables, other module's outputs, ...?

I played around and managed to build the terraform provider locally with commented out `ValidateRawResourceConfigFuncs` in   `resource_keycloak_oidc_identity_provider.go` The error was gone.

michalvavrik:
* Closes: https://github.com/keycloak/terraform-provider-keycloak/issues/1438

If computed values are set, we cannot determine their value during the validation phase. Hence this commit relaxes validation for computed values to "is set".


michalvavrik:
\cc @sschu

sschu (approved):
LGTM, thanks!

provider/resource_keycloak_oidc_identity_provider_test.go
```diff
@@ -484,6 +484,54 @@ resource "keycloak_oidc_identity_provider" "oidc" {
 	`, testAccRealm.Realm, organizationName, oidc)
 }
 
+func TestAccKeycloakOidcIdentityProvider_clientSecretWriteOnlyFromComputedValue(t *testing.T) {
+	t.Parallel()
+
+	oidcName := acctest.RandomWithPrefix("tf-acc")
+
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: testAccProviderFactories,
```
sschu:
```suggestion
		ProtoV5ProviderFactories: testAccProtoV5ProviderFactories,
```

PR #1525, opened 2026-03-19
fix: URL-encode path segments in groupParentId to support group names…

NargiT:
## Problem

Group names containing spaces (e.g. `"Foo Bar"`) or other URL-unsafe
characters does not work when the provider attempts to resolve a
subgroup's parent ID via the Keycloak `group-by-path` API.

The `groupParentId` function builds a URL like:

    /realms/{realm}/group-by-path/Foo Bar/Child

The unencoded space makes this an invalid HTTP request, or a 404:

    Error: no group with name Foo Bar found

Attempting to work around this by using `"Foo%20Bar"` in Terraform config
does not help either, as the name is then compared literally against
Keycloak's returned `name: "Foo Bar"` and the match fails.

## Fix

URL-encode each path segment individually using `url.PathEscape` before
building the `group-by-path` URL in `groupParentId`. This correctly
encodes spaces as `%20` (and other special characters) while preserving
`/` as a path separator.

## Example

A group named `"Foo Bar"` with a child group `"Foo Bar.admins"` can now
be referenced correctly in Terraform:

```hcl
data "keycloak_group" "example" {
  realm_id = keycloak_realm.realm.id
  name     = "Foo Bar.admins"
}


thomasdarimont:
@sschu I think this is a helpful bug fix that we could include in the next release.

NargiT:
Ready to help if you have question

sschu (approved):
LGTM, thanks!

PR #1530, opened 2026-03-23
feat(data/keycloak_group): add group_path attribute for precise nested group lookup

ckyvra:
Adds the ability to look up a group by its **full path** (e.g. `/IT/Backend/Developers`) in the `data.keycloak_group` data source.

This is especially useful for nested groups where multiple groups can have the same `name` in different branches — the existing name-based lookup is ambiguous and uses DFS order (non-deterministic).

The change uses Keycloak's official `/group-by-path` endpoint.


sschu:
@ckyvra Thanks for your PR. I am wondering if it would be possible to just add support to put a path directly into the name field, by always looking up by path first and if not found by name?

ckyvra:
**Proposal**

This PR adds support for group paths in `data "keycloak_group"`.

- If `name` starts with `/`, it is treated as the **full group path** (e.g. `"/Administration/Full Admins"`).
- Otherwise, legacy behavior (simple name lookup) is preserved.

**Example:**
```hcl
name = "/IT/Engineering/Developers"
```

sschu:
@ckyvra Can you please also extend the test to also check that fetching the nested group via path works? See https://github.com/keycloak/terraform-provider-keycloak/blob/acc17548ac6f17411570f16638e433c951155ca7/provider/data_source_keycloak_group_test.go#L37-L66

provider/data_source_keycloak_group.go
```diff
@@ -22,7 +22,16 @@ func dataSourceKeycloakGroup() *schema.Resource {
...
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+			},
+			// group_path enables lookup of nested groups by their full hierarchy path
+			// (e.g. "/parent/child/subgroup"). It is separate from `name` to avoid
+			// overloading semantics and follows Terraform's convention of one field
+			// per concern. Uses the Keycloak /group-by-path endpoint for deterministic,
+			// unambiguous resolution regardless of name collisions.
+			"group_path": {
```
sschu, line 34: @ckyvra If we go the route of using different parameters here, can you please add `ExactlyOneOf` to this and `name` so it is checked in code only one of both is provided?

provider/resource_keycloak_realm.go
```diff
@@ -1239,6 +1239,10 @@ func getRealmFromData(data *schema.ResourceData, keycloakVersion *version.Versio
 		if webAuthnPolicyUserVerificationRequirement, ok := webAuthnPolicy["user_verification_requirement"]; ok {
 			realm.WebAuthnPolicyUserVerificationRequirement = webAuthnPolicyUserVerificationRequirement.(string)
 		}
+	} else if data.HasChange("web_authn_policy") {
```
sschu: What are these changes regarding WebAuthPolicy related to?

provider/data_source_keycloak_group.go
ckyvra: Thanks for the review @sschu.

I've added `ExactlyOneOf` to both `name` and `group_path` in the schema. This enforces mutual exclusivity at plan time.

I also removed the now-redundant manual check `if groupName != "" && groupPath != ""` from the read function, since `ExactlyOneOf` already handles that case.

The "at least one must be set" check is kept because `ExactlyOneOf` only forbids both being set simultaneously, not both being empty.

provider/resource_keycloak_realm.go
ckyvra: These were fixes for a test stability issue (non-empty plan when removing `web_authn_policy` / `web_authn_passwordless_policy` blocks between test steps). They are unrelated to the `group_path` feature, so I've moved them out of this PR.

provider/resource_keycloak_realm.go
```diff
@@ -1239,6 +1239,10 @@ func getRealmFromData(data *schema.ResourceData, keycloakVersion *version.Versio
...
+		realm.WebAuthnPolicyAcceptableAaguids = []string{}
+		realm.WebAuthnPolicyExtraOrigins = []string{}
+		realm.WebAuthnPolicySignatureAlgorithms = []string{}
```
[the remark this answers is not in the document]
ckyvra: **WebAuthn changes:** Agreed — these were unrelated test stability fixes. I've moved them out of this PR.

provider/data_source_keycloak_group.go
```diff
@@ -49,9 +58,25 @@ func dataSourceKeycloakGroupRead(ctx context.Context, data *schema.ResourceData,
...
+	groupPath := data.Get("group_path").(string)
 	groupName := data.Get("name").(string)
 
-	group, err := keycloakClient.GetOrganizationGroupByName(ctx, realmId, organizationId, groupName)
+	if groupName == "" && groupPath == "" {
+		return diag.Errorf("one of `name` or `group_path` must be specified")
+	}
+	if groupName != "" && groupPath != "" {
+		return diag.Errorf("only one of `name` or `group_path` may be specified")
+	}
```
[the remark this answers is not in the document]
ckyvra: **`organization_id` + `group_path`:** Good catch. I've added `ConflictsWith` between `organization_id` and `group_path` in the schema, plus an explicit validation error explaining that `/group-by-path` is a realm-level endpoint and doesn't support organization scoping. This way users get a clear error at plan time rather than a silent mismatch.

sschu (approved):
LGTM, thanks a lot!

PR #1533, opened 2026-03-30
Add relative user creation DN

antonarntz:
Within the Keycloak LDAP user federation, there's the option to add new users in a relative DN to the Users DN.
This pull request adds the option to the terraform provider as well.

Update: Had to fix the TestAccKeycloakRealm_webauthn because it randomly seems to fail. Let me know if I need to make a separate pull request for it.


sschu:
@antonarntz Thanks for your contribution. It looks like tests are consistently failing for KC26.0 and KC26.1, I would assume the added field was not available back then. You can find some inspiration to add conditional support here: https://github.com/keycloak/terraform-provider-keycloak/blob/635b1ebcf7962a8bb178116da8215db01d73f7f2/provider/resource_keycloak_group.go#L108-L125
Please also add the new field to an existing test as suggested by Copilot.

sschu (approved):
LGTM, thanks a lot!

Issue #1270, opened 2025-08-06
Unable to change existing scope assigned type


bassedOutt:
### Describe the bug

I'm trying to change client scopes from optional to default
Scope `email` is set to optional on a realm level
In my client I need to set it as default

So I created a resource

```
resource "keycloak_openid_client_default_scopes" "default_client_scopes" {
  for_each = {
    for v in local.default_client_scopes : "${v.client}-default-scopes" => v
  }

  client_id = keycloak_openid_client.this[each.value.client].id
  realm_id  = keycloak_realm.this[each.value.realm].id

  default_scopes = each.value.default_scopes
}
```

and passed `emai` as a default scope
however, it fails with a message
Error: validation error: scope email is already attached to client as an optional scope

I'd like to notice that I did not explicitly set this scope as optional, this is the assigned type on a realm level

I then tried to specify all scopes - both optional and default like this

``` hcl
resource "keycloak_openid_client_optional_scopes" "optional_client_scopes" {
  for_each = {
    for v in local.optional_client_scopes : "${v.client}-optional-scopes" => v
  }

  client_id = keycloak_openid_client.this[each.value.client].id
  realm_id  = keycloak_realm.this[each.value.realm].id

  optional_scopes = each.value.optional_scopes
}

resource "keycloak_openid_client_default_scopes" "default_client_scopes" {
  for_each = {
    for v in local.default_client_scopes : "${v.client}-default-scopes" => v
  }

  client_id = keycloak_openid_client.this[each.value.client].id
  realm_id  = keycloak_realm.this[each.value.realm].id

  default_scopes = each.value.default_scopes
}

```

and in my terragrunt

``` hcl
...
          default_client_scopes = [
            "acr",
            "auth",
            "email",
            "profile",
            "roles",
            "web-origins"
          ]
          optional_client_scopes = [
            "address",
            "microprofile-jwt",
            "offline_access",
            "phone"
          ]
...
```
and it still fails - meaning there's no way for me to change the assigned type of a scope for my client with terraform

### Version

5.1.1

### Expected behavior

Terraform ovverides the assigned type

### Actual behavior

Terraform fails with this message
`Error: validation error: scope email is already attached to client as an optional scope`

### How to Reproduce?

Have a realm with defaul `email` scope with assigned type `optional`
Try to modify the client scope and set to `default` with `keycloak_openid_client_default_scopes` resource
Terraform will fail

Issue #1212, opened 2025-06-05
Adding a default optional client scope


daveekr2:
### Describe the bug

Hi,

I am trying to add an optional client scope to a client. However, this fails with an error, that the scope is already present on this client. If I do not add the default optional clients, the default optional clients are lost.

> Error: validation error: scope offline_access is already attached to client as an optional scope

Both manifests lead to the same error message. I do not know how to add a client scope without losing the default ones. This seems like an unexpected behavior to me.

Thanks for your feedback

### Version

26

### Expected behavior

Being able to add an optional client scope to existing ones.

### Actual behavior

You need to only specify the new client scope, in case you also set preassigned ones the manifest fails.

### How to Reproduce?

I tried 
```
resource "keycloak_realm_optional_client_scopes" "default_optional_mqtt_client_scope" {
    realm_id = data.keycloak_realm.mqtt_realm.id
    optional_scopes = concat(tolist(data.keycloak_realm.mqtt_realm.default_optional_client_scopes), [
        keycloak_openid_client_scope.mqtt.name
    ])
}
```
and
```
resource "keycloak_realm_optional_client_scopes" "default_optional_mqtt_client_scope" {
    realm_id = data.keycloak_realm.mqtt_realm.id
    optional_scopes = ["address", "microprofile-jwt", "offline_access", "phone",
        keycloak_openid_client_scope.mqtt.name
    ]
}
```

Issue #498, opened 2021-03-15
keycloak_openid_client_optional_scopes does not remove scopes on first apply


marknelissen:
The keycloak_openid_client_optional_scopes resource does not apply the list exactly as described on the first run of a new environment. It does add the elements listed, but does not remove the not listed default entries. This makes it so that following example requires two applies (of which one a failed one) to actually get the wanted situation: move the offline_access scope from the optional to the default scope.
```
resource "keycloak_openid_client_optional_scopes" "opt" {
  realm_id  = keycloak_realm.tenant.id
  client_id = keycloak_openid_client.client.id
  optional_scopes = [
    "address",
    "phone",
    "microprofile-jwt",
  ]
}

resource "keycloak_openid_client_default_scopes" "def" {
  realm_id  = keycloak_realm.tenant.id
  client_id = keycloak_openid_client.client.id
  default_scopes = [
    "profile",
    "email",
    "roles",
    "web-origins",
    "offline_access",
  ]
  depends_on = [
    keycloak_openid_client_optional_scopes.opt
  ]
}
```

hamiltont:
`keycloak_openid_client_default_scopes` has exactly the same issue - the first apply will add your own listed scopes, but it takes a second apply for TF to notice that the KC added some additional default "default scopes" on the server side. 

I wonder if the KC API returns a representation of the object that was just created? If yes, then perhaps a general solution that would work for multiple resource types would be to have some central logic to test if the returned resource equals the requested resource, and auto-apply that resource a second time if it appears the server added/customized the creation request based on server-side defaults. It's a fairly common problem with REST APIs that they expand a POST request and you need a followup PUT/PATCH to get what you actually requested.

mrparkers:
This was a deliberate decision I made back when creating these resources.  At the time, it felt weird to me that a "create" for a resource could result in a potentially destructive action (a client scope being removed from the list).

However, I think I've changed my mind on this, and I agree that it would be better (and more CI friendly) for this resource to fully reconcile on both create and update.  We already do this for the `keycloak_user_groups` resource, for example.

[...]

PR #1548, opened 2026-04-10
feat(openid_client): allow default/optional update of existing scopes

EliSauder:
Replicate the Keycloak admin behavior of deleting then re-adding the scope. While this does delete and re-add a resource rather than pure mutation (mentioned as a consideration in #498), this matches the expected behavior while using the resource.
Addresses #1212, #1270


sschu:
@EliSauder Are you still interested in landing this? There are some test failures that need to be addressed.

EliSauder:
@sschu yeah, I think this just got lost in my backlog. ~~Any chance the CI could be run again so I can see what failed?~~ Remembered I needed to update from the primary branch anyways.

PR #1553, opened 2026-04-14
commit for referral addition

demetthyl:
Second attempt to add the referral parameter on ldap user federation


sschu:
@demethyl Something is still wrong, there are a lot of commits and changes in your PR. I will go ahead and close it, feel free to open a new one only changing what you are actually trying to change.

Issue #1510, opened 2026-03-13
Allow strings for *_wo_version


dominiktopp:
### Description

The attributes *_wo_version are of type Int (e.g. https://github.com/keycloak/terraform-provider-keycloak/blob/42ae5d837d4a2e2997e3670bbdefbc6ff3781811/provider/resource_keycloak_openid_client.go#L84-L90)

Please change this to string for more flexibility.

### Motivation

We are using Azure Key Vault as a source for the client secrets used in the Keycloak Terraform provider.
Azure Key Vault secrets provide a [version attribute](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/key_vault_secret#version-1) but it is of type string. Therefore we cannot just pass this to the Keycloak Terraform provider.
I dont't know how other cloud providers handle versioning of secrets, but type string seems to be very flexible because you could also just use e.g. the last modified timestamp for *_wo_version.

Fortunately this version string from Azure uses the hexadecimal alphabet, so we can convert it with `parseint(data.azurerm_key_vault_secret.secret.version, 16)` as a workaround.

PR #1559, opened 2026-04-16
convert wo_version attributes to string

dominiktopp:
The attributes *_wo_version are of type Int.
Type string is much more flexible: For example the Azure Key Vault Secret resource provides a version attribute of type string. But this cannot be used, because Keycloak Terraform Provider forces type int.

solves https://github.com/keycloak/terraform-provider-keycloak/issues/1510


sschu:
@dominiktopp I fear this could be a breaking change. Can you check what happens when you create an int version and then switch to a provider using string instead?

dominiktopp:
Yeah, I also thought about this. I hope I'll find time to test it during the next one or two weeks

remcojansen:
Changing version type to string makes it more versatile indeed. I had the same intuition about backward compatibility and ran a few tests.

1. Create `keycloak_openid_client` resource with `client_secret_wo_version = 1` using provider v5.8.0
2. Change `client_secret_wo_version` value from `1` -> `2`
3. Apply same configuration (`client_secret_wo_version = 2`) using custom built provider from your branch to confirm no drift would be reported
4. Change `client_secret_wo_version` value from `2` -> `3`
5. Change `client_secret_wo_version` value from `3` -> `"four"`

It turns out the provider handles all cases well. After step 3 it silently converts integers to strings and treats the integers in the configuration as well as in the state as strings.

That said, when I asked copilot, I learned that it is recommended to handle the conversion explicitly by introducing a new version of the resource and adding a function that upgrades the state to the new type. You can see here what that could look like: https://github.com/dominiktopp/terraform-provider-keycloak/compare/wo_version_as_string...remcojansen:terraform-provider-keycloak:wo_version_as_string_upgrade

I ran the tests with this code as well and did not observe a difference in behavior.

dominiktopp:
Thanks @remcojansen, I merged your code and also adapted it for the identity provider ressource.

Currently the test / acceptance is failing. But it seems to be flaky, on fail has been in 26.5.7 and one in 26.1.4...

sschu (approved):
Thanks @dominiktopp and @remcojansen , LGTM!

Issue #1494, opened 2026-03-04
Add `import` arguments for all resources created by Keycloak by default


dominiktopp:
### Description

Keycloak creates many default resources when creating a realm: scopes (profile, email, ...), clients (account, broker, ...), roles and so on.

Some of those resources like e.g. [keycloak_openid_client ](https://registry.terraform.io/providers/keycloak/keycloak/5.7.0/docs/resources/openid_client#import-2) provide an attribute `import` which allows to import instead of create this resource.

Please add this `import` attribute to all resource types created by Keycloak during realm creation.

### Motivation

It is not easy to workaround those limitation without manual steps:

1. comment out the resource (e.g. `"keycloak_openid_client_scope" "gkp_external_profile"` )
2. terraform apply
3. add/uncomment the resource, add data and import block
```
data "keycloak_openid_client_scope" "profile" {
  realm_id = "my-realm"
  name     = "profile"
}
import {
    to = keycloak_openid_client_scope.profile
    id = "my-realm/${data.keycloak_openid_client_scope.profile.id}"
}
resource "keycloak_openid_client_scope" "profile" {
    realm_id               = "my-realm"
    name                   = "profile"
    consent_screen_text    = "our custom consent screen text for client scope profile"
}
```
4. terraform apply

### Details

At least `keycloak_openid_client_scope` and `keycloak_generic_protocol_mapper` are missing an `import` attribute for our use case.

#732 and #1083 describe the same problem but limited to protocol mappers.

UristMcEngineer:
One resource which I would currently need an import attribute for is `keycloak_openid_full_name_protocol_mapper`. This is necessary to overwrite which user information gets mapped to the `name` claim where currently the automatically generated full name mapper applies. Simply adding a `keycloak_openid_user_attribute_protocol_mapper` which also targets the `name` claim does not work reliably and the existing `keycloak_openid_full_name_protocol_mapper` has to be manually configured to not be applied or manually deleted.

FWest98:
We are running into the same issue. Specifically, we have the following objectives:
- Augment/modify an existing OIDC client scope - the `import` key could work here
- Remove existing mappers inside client scopes - the `import` key does not seem to be a complete solution yet. Similar to @UristMcEngineer, we want to remove the default "full name" mapper and replace it with a custom mapper that provides the same `name` claim.

PR #1579, opened 2026-05-01
Add Import to KC realm and generic protomapper resources

fourjr:
Related to #1494

Adds import keyword to allow importing pre created resources such as master realm without running the tf import command seperately.


sschu:
@fourjr Thanks for you contribution! The DCO check if failing, can you please sign-off your commits?

sschu:
@fourjr Since Terraform supports import blocks natively now, the current direction is to not add special import attributes to the provider anymore. Therefore, I will go ahead and close this PR. Feel free to comment if you disagree.

PR #1583, opened 2026-05-07
Document allow_utf8 parameter for keycloak_realm smtp_server

ingokuba:
This parameter has been added with [this commit](https://github.com/keycloak/terraform-provider-keycloak/commit/2c7e6790525e2cc09c65070a96634b1e7983b4f1) but is missing in the documentation: https://registry.terraform.io/providers/keycloak/keycloak/latest/docs/resources/realm#smtp


sschu (approved):
@ingokuba Thanks for spotting!

PR #1589, opened 2026-05-19
feat(google-idp): expose login_hint as top-level attribute

springerigor:
## What

Adds `login_hint` as a top-level attribute on `keycloak_oidc_google_identity_provider`, mirroring what `keycloak_oidc_identity_provider` and `keycloak_saml_identity_provider` already expose.

## Why

The underlying `keycloak.IdentityProviderConfig` struct already has the field:

```go
LoginHint string `json:"loginHint,omitempty"`
```

and Keycloak's `AbstractOAuth2IdentityProvider#createAuthorizationUrl` honors it for the Google IDP, since `GoogleIdentityProvider` extends `OIDCIdentityProvider` which extends `AbstractOAuth2IdentityProvider`. Closed upstream PR [keycloak/keycloak#8374](https://github.com/keycloak/keycloak/pull/8374) was specifically about adding the Google admin UI toggle - the runtime support is already there.

Today this attribute is unreachable through this provider:

- The Google resource schema doesn't expose it as a top-level attribute.
- Setting `extra_config = { loginHint = "true" }` is rejected with `extra_config key "loginHint" is not allowed, as it conflicts with a top-level schema attribute`, because `validateExtraConfig` reflects over `IdentityProviderConfig` and finds the matching JSON tag.
- Setting it out-of-band via `kcadm` is wiped on the next `terraform apply` because the resource PUTs a config that omits the field.

This leaves users with no way to enable `login_hint` forwarding for a Google IDP via Terraform, even though the underlying behavior is fully supported by Keycloak. This PR closes that gap.

## Changes

- `provider/resource_keycloak_oidc_google_identity_provider.go` - add `login_hint` to the schema; wire it through `getOidcGoogleIdentityProviderFromData` and `setOidcGoogleIdentityProviderData`.
- `provider/resource_keycloak_oidc_google_identity_provider_test.go` - add `TestAccKeycloakOidcGoogleIdentityProvider_loginHint` acceptance test.
- `docs/resources/oidc_google_identity_provider.md` - document the new attribute.

Note: the underlying Keycloak attribute `loginHint` is a boolean stored as a string (parsed via `Boolean.parseBoolean`), so values should be `"true"` or `"false"`. This matches how the generic OIDC and SAML resources already model the same attribute (`schema.TypeString`).

## Usage

```hcl
resource "keycloak_oidc_google_identity_provider" "google" {
  realm         = keycloak_realm.realm.id
  client_id     = var.google_client_id
  client_secret = var.google_client_secret
  login_hint    = "true"
}
```

### DIFF
--- a/docs/resources/oidc_google_identity_provider.md
+++ b/docs/resources/oidc_google_identity_provider.md
@@ -50,6 +50,7 @@ resource "keycloak_oidc_google_identity_provider" "google" {
 - `request_refresh_token` - (Optional) Sets the "access_type" query parameter to "offline" when redirecting to google authorization endpoint,to get a refresh token back. This is useful for using Token Exchange to retrieve a Google token to access Google APIs when the user is offline.
 - `default_scopes` - (Optional) The scopes to be sent when asking for authorization. It can be a space-separated list of scopes. Defaults to `openid profile email`.
 - `accepts_prompt_none_forward_from_client` - (Optional) When `true`, unauthenticated requests with `prompt=none` will be forwarded to Google instead of returning an error. Defaults to `false`.
+- `login_hint` - (Optional) Pass `login_hint` to the Google identity provider. Set to `"true"` to forward the `login_hint` query parameter from the inbound OIDC request to Google. The underlying Keycloak attribute `loginHint` is a boolean string, so the value should be `"true"` or `"false"`.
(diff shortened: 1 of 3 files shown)


sschu (approved):
LGTM, thanks!

PR #1590, opened 2026-05-19
fix(authentication-execution-config): use executionId-aware endpoint for GET

felipe-gon:
## What

Fixes `GetAuthenticationExecutionConfig` in the SDK so it queries the
endpoint that current Keycloak versions actually expose:

- before: `GET /realms/{realm}/authentication/config/{id}`
- after:  `GET /realms/{realm}/authentication/executions/{executionId}/config/{id}`

The first form returns `404` on recent Keycloak releases, which made
`keycloak_authentication_execution_config` resources unreadable after
creation/refresh.

## Why

`keycloak.AuthenticationExecutionConfig` already carries `ExecutionId`,
and other operations in this file (`NewAuthenticationExecutionConfig`,
`UpdateAuthenticationExecutionConfig`, `DeleteAuthenticationExecutionConfig`)
already use the executionId-scoped path. Only the GET was using the legacy
shape, which is why the resource only broke on read/refresh.

## Changes

- `keycloak/authentication_execution_config.go`: switch the GET URL to the
  executionId-scoped path.
- `provider/resource_keycloak_authentication_execution_config_test.go`:
  populate `ExecutionId` in the destroy check helper, otherwise the helper
  would call the new endpoint without the required path segment.

## Test plan

Existing acceptance tests cover this code path:

- `TestAccKeycloakAuthenticationExecutionConfig_basic`
- `TestAccKeycloakAuthenticationExecutionConfig_updateForcesNew`
- `TestAccKeycloakAuthenticationExecutionConfig_import`

Both `testAccCheckKeycloakAuthenticationExecutionConfigExists` and
`testAccCheckKeycloakAuthenticationExecutionConfigDestroy` exercise
`GetAuthenticationExecutionConfig`, so the fix is covered end-to-end.

### DIFF
--- a/keycloak/authentication_execution_config.go
+++ b/keycloak/authentication_execution_config.go
@@ -25,7 +25,7 @@ func (keycloakClient *KeycloakClient) NewAuthenticationExecutionConfig(ctx conte
 
 // GetAuthenticationExecutionConfig https://www.keycloak.org/docs-api/latest/rest-api/index.html#_get_adminrealmsrealmauthenticationconfigid
 func (keycloakClient *KeycloakClient) GetAuthenticationExecutionConfig(ctx context.Context, config *AuthenticationExecutionConfig) error {
-	return keycloakClient.get(ctx, fmt.Sprintf("/realms/%s/authentication/config/%s", config.RealmId, config.Id), config, nil)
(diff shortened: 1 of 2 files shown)


sschu:
@felipe-gon Thanks for your contribution! If the current behaviour is wrong, shouldn't we modify the test so it actually fails and is successful only with your fix?

sschu:

PR #1591, opened 2026-05-20
feat(authentication-execution): add flow_id and wire priority into create payload

felipe-gon:
## What

Allow `keycloak_authentication_execution` to model a sub-flow execution
via a new optional `flow_id` attribute, and send `priority` directly in
the create payload (gated to Keycloak >= 25).

## Why

The Keycloak API exposes two kinds of executions on the same endpoint
(`POST /authentication/flows/{alias}/executions/execution`): leaf
executions (with `provider`) and sub-flow executions (with `flowId`).
The resource only modeled the leaf case, so nesting a flow as a step of
another flow was not possible from Terraform.

`priority` was already in the schema but ignored on create — the
execution was always appended and then re-ordered via raise/lower in a
follow-up call. Keycloak 25 accepts `priority` on create, which makes
the ordering deterministic and saves API round-trips.

## Changes

- `provider/resource_keycloak_authentication_execution.go`
  - Add `flow_id` (string, Optional, ForceNew).
  - `authenticator` becomes Optional (was Required) — a sub-flow
    execution does not need a provider id.
  - Read/write `flow_id` to/from `AuthenticationExecution`.
- `keycloak/authentication_execution.go`
  - Extend `authenticationExecutionCreate` with `Priority` and `FlowId`
    (both `omitempty`).
  - In `NewAuthenticationExecution`, always send `FlowId`; send
    `Priority` only when the Keycloak server is `>= 25`.
  - In `UpdateAuthenticationExecution`, propagate `Priority` into the
    requirement update payload, also gated to `>= 25`.
  - Mark `Priority` on `AuthenticationExecution` /
    `AuthenticationExecutionInfo` as `omitempty` so older Keycloak
    versions don't receive a zero value they didn't ask for.
- `docs/resources/authentication_execution.md`: document `flow_id` and
  the new optionality of `authenticator`.

## Backward compatibility

- Existing configs (leaf executions) keep working unchanged.
- On Keycloak < 25, the create payload is unchanged: `priority` is
  omitted and still applied via the existing update step.

## Usage

```hcl
resource "keycloak_authentication_flow" "subflow" {
  realm_id = data.keycloak_realm.realm.id
  alias    = "my-subflow"
}

resource "keycloak_authentication_execution" "nested" {
  realm_id          = data.keycloak_realm.realm.id
  parent_flow_alias = keycloak_authentication_flow.parent.alias
  flow_id           = keycloak_authentication_flow.subflow.id
  requirement       = "ALTERNATIVE"
  priority          = 10
}
```

### DIFF
--- a/docs/resources/authentication_execution.md
+++ b/docs/resources/authentication_execution.md
@@ -31,6 +31,7 @@ resource "keycloak_authentication_execution" "execution_one" {
   authenticator     = "auth-cookie"
   requirement       = "ALTERNATIVE"
   priority			= 10
+  flow_id           = "${keycloak_authentication_flow.flow.id}"
 }
 
 # second execution
@@ -40,6 +41,7 @@ resource "keycloak_authentication_execution" "execution_two" {
   authenticator     = "identity-provider-redirector"
   requirement       = "ALTERNATIVE"
   priority			= 20
+  flow_id           = "${keycloak_authentication_flow.flow.id}"
 }
 ```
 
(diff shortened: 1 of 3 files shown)


sschu:
@felipe-gon Can you please add a test to verify that the new flow-id feature works?

sschu:

PR #1593, opened 2026-05-20
fix: handle non-string types in keycloak_realm_user_profile validator config

felipe-gon:
## What

Fixes refresh/import on `keycloak_realm_user_profile` when a validator's `config` map contains values that Keycloak returns as something other than a string (e.g. a JSON number for `length.min`/`length.max`, or a JSON array for `options.options`).

## Why

The Terraform schema declares the validator `config` as `map[string]string`. The previous implementation of `getRealmUserProfileAttributeData` built `map[string]interface{}` from the API response, JSON-encoded only the `[]interface{}` case, and passed everything else through unchanged. When Keycloak returned a number or an array for a key the SDK could not coerce the result into the schema and the apply/refresh/import would fail.

## Changes

`provider/resource_keycloak_realm_user_profile.go`: in `getRealmUserProfileAttributeData`, switch the local map to `map[string]string` and convert each value with a type switch over the kinds the API actually returns:

- `string`: pass through.
- `[]interface{}`: JSON-encode to a string.
- `float64`: format with `strconv.FormatFloat(..., 'f', -1, 64)`.

### DIFF
--- a/provider/resource_keycloak_realm_user_profile.go
+++ b/provider/resource_keycloak_realm_user_profile.go
@@ -3,6 +3,7 @@ package provider
 import (
 	"context"
 	"encoding/json"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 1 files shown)


sschu:

PR #1603, opened 2026-05-27
fix(openid-client): prevent stale client_secret in state from overwriting client_secret_wo

### LINKED ISSUE #1602: keycloak_openid_client: stale client_secret in state silently overwrites write-only secret on unrelated changes
### Describe the bug

Using client_secret_wo to manage a client secret works correctly when first setting or rotating the secret (by incrementing client_secret_wo_version). However, any subsequent apply that changes an unrelated attribute - such as `extra_config` or name - while leaving `client_secret_wo_version` unchanged will silently replace the active Keycloak secret with a stale value from Terraform state.

This happens because `client_secret` is `Computed: true`: Terraform always holds a value for it, either one the user explicitly provided or the secret Keycloak auto-generated for the CONFIDENTIAL client. When write-only mode is active, the provider never clears this stored value. So on any apply that does not trigger a secret rotation, the stale `client_secret` gets sent to Keycloak in the PUT body, overwriting the write-only secret that was previously set.

This affects two scenarios:
1. A user who previously had `client_secret` in their config and later migrated to `client_secret_wo`.
2. A user who never had an explicit `client_secret` in config (e.g. using `random_password` as the write-only source from the start) — Keycloak auto-generates a secret for every CONFIDENTIAL client, and the provider stores it as a computed value in state.

### Version

Reproducible on current main

### Expected behavior

When `client_secret_wo_version` is unchanged, Keycloak's current secret is left untouched regardless of what other attributes are updated. terraform plan after such an apply shows no drift.

### Actual behavior

Keycloak's active secret is silently replaced with the stale client_secret value from Terraform state. The apply succeeds without error. Any application using the write-only secret stops working.

### How to Reproduce?

```hcl
  # Step 1 — CONFIDENTIAL client, no explicit secret
  resource "keycloak_openid_client" "client" {
    client_id   = "my-client"
    realm_id    = data.keycloak_realm.realm.id
    access_type = "CONFIDENTIAL"
  }

  # Step 2 — switch to write-only (e.g. backed by random_password)
  resource "keycloak_openid_client" "client" {
    client_id                = "my-client"
    realm_id                 = data.keycloak_realm.realm.id
    access_type              = "CONFIDENTIAL"
    client_secret_wo         = random_password.secret.result
    client_secret_wo_version = 1
  }

  # Step 3 — change any unrelated attribute, keep version = 1
  resource "keycloak_openid_client" "client" {
    client_id                = "my-client"
    realm_id                 = data.keycloak_realm.realm.id
    access_type              = "CONFIDENTIAL"
    client_secret_wo         = random_password.secret.result
    client_secret_wo_version = 1
    extra_config             = { "key1" = "value1" }
  }
```

After step 3 applies, the Keycloak client secret is the auto-generated value from step 1, not the random_password value set in step 2.

plafue:
client_secret is Computed: true, so Terraform always stores a value for it — either one the user explicitly set, or the secret Keycloak generated automatically for a CONFIDENTIAL client. This means every client that ever existed without an explicit client_secret still carries the Keycloak-generated value in state.

The typical real-world pattern is: an existing CONFIDENTIAL client (possibly managed via random_password or without any secret config at all) is migrated to client_secret_wo. The old computed client_secret remains in state because setOpenidClientData never clears it when write-only mode is active. On any subsequent apply where client_secret_wo_version is unchanged —  an unrelated attribute change, a drift correction, anything — getOpenidClientFromData reads the stale value via data.Get("client_secret") and includes it in the PUT body, silently overwriting the active write-only secret with the old one.

Fix: add an else-if branch to the write-only block in getOpenidClientFromData. When the write-only version is non-zero but unchanged (i.e. the version-gate intentionally skips the secret update), explicitly zero out ClientSecret. Because OpenidClient is marshalled with encoding/json and ClientSecret carries the omitempty tag, the empty value is omitted from the PUT body entirely, leaving Keycloak's current secret untouched.

Three acceptance tests cover the fix:
- explicit client_secret migrated to write-only, then unrelated change
- CONFIDENTIAL client with no explicit secret (computed value in state) migrated to write-only, then unrelated extra_config change
- fresh write-only client, unrelated extra_config change, PlanOnly no-drift

Fixes #1602

### DIFF
--- a/provider/resource_keycloak_openid_client.go
+++ b/provider/resource_keycloak_openid_client.go
@@ -484,6 +484,10 @@ func getOpenidClientFromData(data *schema.ResourceData) (*keycloak.OpenidClient,
 		}
 
 		openidClient.ClientSecret = clientSecretWriteOnly.AsString()
+	} else if data.Get("client_secret_wo_version").(int) != 0 {
+		// write-only mode, version unchanged: discard any stale client_secret from state
+		// so it is not sent to Keycloak and does not overwrite the existing secret
+		openidClient.ClientSecret = ""
 	}
 
 	if !openidClient.ImplicitFlowEnabled && !openidClient.StandardFlowEnabled {
--- a/provider/resource_keycloak_openid_client_test.go
+++ b/provider/resource_keycloak_openid_client_test.go
@@ -2107,7 +2107,7 @@ resource "keycloak_openid_client" "client" {
 	access_type              = "CONFIDENTIAL"
 	name                     = "%s"
 	client_secret_wo         = "%s"
-	client_secret_wo_version = "%d"
+	client_secret_wo_version = %d
 }
 	`, testAccRealm.Realm, clientId, name, clientSecretWriteOnly, clientSecretWriteOnlyVersion)
 }
--- a/provider/resource_keycloak_openid_client_test.go
+++ b/provider/resource_keycloak_openid_client_test.go
@@ -496,6 +496,104 @@ func TestAccKeycloakOpenidClient_secretWriteOnly(t *testing.T) {
 	})
 }
 
+func TestAccKeycloakOpenidClient_secretWriteOnlyNotClearedAfterExplicitSwitch(t *testing.T) {
+	t.Parallel()
+	clientId := acctest.RandomWithPrefix("tf-acc")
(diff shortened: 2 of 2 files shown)


sschu (approved):
LGTM, thanks!

PR #1616, opened 2026-06-06
oidc idp: prevent client_secret from being cleared on update when using client_secret_wo

### LINKED ISSUE #1576: Client secret is reset when identity provider is updated while using `client_secret_wo`
### Describe the bug

Client secret is set to an empty string when identity provider is updated while using `client_secret_wo`.

### Version

26.5.5

### Expected behavior

Client secret is not modified unless `client_secret_wo_version` changes.

### Actual behavior

Client secret is modified even though `client_secret_wo_version` was not modified. Also, client secret is set to an empty string instead of the configured value in `client_secret_wo`.

### How to Reproduce?

Using the example project below:
1. `docker compose up -d`
2. `terraform init`
3. `terraform apply -auto-approve`
4. Verify that client secret has been set (shows `••••••••••`) using the [web interface](http://localhost:8080/admin/master/console/#/test/identity-providers/oidc/test/settings)
5. Change any value in the identity provder config in `main.tf`, e.g. `backchannel_supported` to `true`
6. `terraform apply -auto-approve`
7. Check client secret again: the web interface shows an empty input field for the client secret

### Example project
`compose.yaml`
```yaml
services:
  keycloak:
    image: quay.io/keycloak/keycloak:26.5.5
    command: start-dev
    environment:
      KC_BOOTSTRAP_ADMIN_USERNAME: admin
      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
    ports:
      - 8080:8080
```

`main.tf`
```terraform
terraform {
  required_providers {
    keycloak = {
      source  = "keycloak/keycloak"
      version = "5.7.0"
    }
  }
}

provider "keycloak" {
  url       = "http://localhost:8080"
  client_id = "admin-cli"
  username  = "admin"
  password  = "admin"
}

  realm = "test"
}

resource "keycloak_oidc_identity_provider" "idp" {
  realm = keycloak_realm.realm.id
  alias = "test"

  authorization_url = "https://example.org"
  token_url         = "https://example.org"

  client_id                = "something"
  client_secret_wo         = "test"
  client_secret_wo_version = 1

  backchannel_supported = false # change to trigger update
}

```

plafue:
## Problem

  When using `client_secret_wo` / `client_secret_wo_version` to manage the OIDC identity provider client secret as a write-only value, any update to an **unrelated** attribute (e.g.
  `backchannel_supported`) would silently clear the configured secret in Keycloak, locking out federation until the secret was manually re-entered.

  Fixes #1576

  ## Root cause

  Keycloak clears `clientSecret` whenever the key is absent from the identity provider PUT body. When `client_secret_wo_version` is unchanged across an apply, the write-only value is not available
  in state, so the outgoing config had no `clientSecret` — causing Keycloak to wipe the stored secret on every unrelated update.

  ## Fix

  The update path already fetches the current identity provider from Keycloak before writing changes (for organisation link/unlink handling). When the outgoing config carries no `clientSecret`, the
  existing value returned by Keycloak — `**********` — is copied into the PUT payload. Keycloak recognises the masked placeholder and preserves the stored secret.

  ## Test

  `TestAccKeycloakOidcIdentityProvider_clientSecretWriteOnlyNotClearedOnUpdate` covers the regression:

  1. Create an OIDC identity provider with `client_secret_wo`
  2. Update an unrelated field (`backchannel_supported`) without changing the version
  3. Assert Keycloak still has the secret set
  4. Assert there is no perpetual diff on the subsequent plan


sschu (approved):
LGTM, thanks @plafue !

PR #1618, opened 2026-06-08
feat: support import for realm default and optional client scopes

nlenepveu:
## Description

`keycloak_realm_default_client_scopes` and `keycloak_realm_optional_client_scopes` reconcile a realm's default/optional client scopes, but neither defines an `Importer`. As a result, an `import {}` block (or `terraform import`) against an existing realm errors with *"resource ... doesn't support import"*, and the only way to bring a realm's pre-existing scopes under management is an opaque create.

This PR adds a `StateContext` importer to each resource. The import id is the **realm id** (which is also the resource id). The importer sets `realm_id` from the id; the existing `Read` then resolves the realm's current scopes into state. This lets operators adopt a realm's existing (e.g. Keycloak factory) scopes and review the exact delta to their declared set in the plan, instead of starting from a create.

## Changes

- `keycloak_realm_default_client_scopes`: add importer + import function.
- `keycloak_realm_optional_client_scopes`: same, for consistency (it had the identical gap).
- Add an `ImportStateVerify` step to each resource's basic acceptance test (import id derived from `realm_id`).
- Replace the "does not support import" note in both doc pages with an `## Import` section + `terraform import` example.
- CHANGELOG entry.

## Usage

```bash
terraform import keycloak_realm_default_client_scopes.default_scopes <realm-id>
terraform import keycloak_realm_optional_client_scopes.optional_scopes <realm-id>
```

## Checklist

- [x] `make vet` passes
- [x] `make fmtcheck` passes
- [x] `ImportStateVerify` acceptance test coverage added for both resources
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## What's Changed
 
+FEATURES:
+
+* feat: support import for `keycloak_realm_default_client_scopes` and `keycloak_realm_optional_client_scopes` using the realm id
+
 ## 5.8.0 (June 05, 2026)
 
 FEATURES:
--- a/docs/resources/realm_default_client_scopes.md
+++ b/docs/resources/realm_default_client_scopes.md
@@ -43,5 +43,13 @@ resource "keycloak_realm_default_client_scopes" "default_scopes" {
 
 ## Import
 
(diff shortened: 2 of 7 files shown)


sschu (approved):
LGTM, thanks a lot!

PR #1486, opened 2026-02-26
1485 drop support for outdated Keycloak versions

sschu:
Closes #1485

PR #1631, opened 2026-06-17
Document removal of Keycloak < 26.x Support in Changelog

RafaelWO:
This change adds a "BREAKING CHANGES" section to the changelog mentioning the removal of support for Keycloak versions older than 26.

See also https://github.com/keycloak/terraform-provider-keycloak/pull/1486


sschu (approved):
LGTM, thanks!

PR #1636, opened 2026-06-23
fix: add the `idpEntityId` as a parameter for identity providers

dgnfu:
Noted that the IdpEntityId was not included as an option when setting up the identity providers.


sschu:
@dgnfu Thanks for your contribution! If you are still interested in landing this, can you address the review comments from Copilot? Also you need to add a sign-off to your commit.

PR #1637, opened 2026-06-23
fix: gracefully handle 404 on delete for groups and users 

Filirom1:
Handle API 404 errors gracefully when deleting resources that represent user-group memberships, group-role mappings, user-group assignments, and user-role mappings. 

If the parent resource (the Realm, User, or Group) has already been deleted, these mappings do not exist anymore. 

Deleting them should succeed rather than failing or causing nil pointer dereferences/panics.


sschu (approved):
LGTM, thanks!

PR #1639, opened 2026-06-26
feat: add generic client authorization policy resource and restore keycloak_openid_client_js_policy

### LINKED ISSUE #1638: `keycloak_openid_client_js_policy` removed in v5.8.0 — regression for deployed JS authorization policies, no replacement
### Describe the bug

`keycloak_openid_client_js_policy` was removed in v5.8.0 (#1486, rationale in #1485). The stated motivation — removing features "not supported by Keycloak >= 26" — does not apply to this resource. Deployed JavaScript authorization policies (JAR + `META-INF/keycloak-scripts.json`) are still supported by Keycloak; they are a Technology Preview gated behind the `scripts` feature flag, not a removed feature. What Keycloak removed (~KC 18) is the inline `upload_scripts` path, not deployed script providers, which are the officially recommended approach and work in KC 26. This resource explicitly supports the deployed path — see the provider's own CHANGELOG ("support deployed JavaScript policies for `keycloak_openid_client_js_policy` resource", #275).

Two follow-on problems: 
(1) there is no replacement — only type-specific policy resources remain (`role`/`group`/`user`/`client`/`authorization_client_scope`/`aggregate`/`time`/`regex`) plus a read-only `openid_client_authorization_policy` data source, none of which can create a policy whose type is a deployed script provider; the `generic_*` resources cover mappers only. 
(2) The removal shipped in a minor release (5.7.0 → 5.8.0); the v5.8.0 BREAKING CHANGES section lists only "Remove support for Keycloak versions older than 26", not the removal of this resource.

### Version

26.6.2

### Expected behavior

`keycloak_openid_client_js_policy` continues to be available to manage deployed JavaScript authorization policies, since the underlying Keycloak feature is still supported. If deprecation is still intended, it follows the normal cycle (deprecation warning in a minor, removal in the next major) and is documented under BREAKING CHANGES. Alternatively, a supported replacement exists (e.g. a generic authorization-policy resource able to reference a deployed-script policy type).

### Actual behavior

The resource is gone as of v5.8.0 (present through v5.7.0). There is no supported way to manage deployed JS authorization policies via Terraform, and the removal is not listed as a breaking change in the v5.8.0 changelog (only the Keycloak-version-support drop is). Existing configurations using the resource fail on upgrade to >= 5.8.0.

### How to Reproduce?

1. With provider v5.7.0 and Keycloak 26 started with `--features=scripts`, deploy a JS policy as a JAR (`META-INF/keycloak-scripts.json` referencing e.g. `my-policy.js`).
2. Manage it with:

```hcl
   resource "keycloak_openid_client_js_policy" "example" {
     realm_id           = keycloak_realm.main.id
     resource_server_id = keycloak_openid_client.app.id
     name               = "user-is-customer-admin"
     decision_strategy  = "UNANIMOUS"
     logic              = "POSITIVE"
     code               = "my-policy.js"   # deployed script reference
     lifecycle { ignore_changes = [code] }
   }
```

   `terraform apply` works as expected.
3. Bump the provider to v5.8.0 and run `terraform init -upgrade && terraform plan`.
4. The configuration no longer validates — the resource type `keycloak_openid_client_js_policy` does not exist — and there is no alternative resource to bind the deployed policy.

### Anything else?

First off, a big thank you to the maintainers. This provider does a lot of heavy lifting in our infrastructure, and the ongoing work, including the cleanup and the move toward the plugin framework, is genuinely appreciated. I'm raising this only because the resource is important to our setup, and I hope the context is useful rather than just another complaint.

pboehm:
Fixes https://github.com/keycloak/terraform-provider-keycloak/issues/1638

`keycloak_openid_client_js_policy` was removed in v5.8.0 as part of dropping
support for old Keycloak versions, but that removal was not intentional —
deployed JavaScript authorization policies are still supported by Keycloak 26.
This PR addresses the regression in two ways.

## 1. New resource: `keycloak_generic_client_authorization_policy`

A generic authorization policy resource, in the spirit of the existing
`keycloak_generic_*` resources. It manages a client authorization policy of any
provider type by referencing that provider through the `type` attribute, so it
works with:

- custom policy providers implemented as a Java SPI (`type` is whatever the
  `PolicyProviderFactory.getId()` returns)
- JavaScript policies deployed as a script in a JAR (`type` is the
  `script-<fileName>` id Keycloak assigns to the deployed script)

This is the recommended, future-proof way to manage deployed and custom policy
providers. It has no `code` attribute, so it does not suffer from the read-back
drift the JS-specific resource has.

## 2. Restored resource: `keycloak_openid_client_js_policy`

Brought back so existing configurations keep working without a migration. Two
changes reflect current Keycloak behaviour:

- It now handles only deployed script policies. `code` must be the deployed
  provider id (`script-<fileName>`), and is validated as such. Inline
  JavaScript upload was removed in Keycloak 18, which is no longer supported by
  this provider.
- On read Keycloak returns the script source in `code`, so the docs and example
  show `lifecycle { ignore_changes = [code] }`. The resource is not marked
  deprecated, but its documentation points to
  `keycloak_generic_client_authorization_policy` as the better alternative.

## Testing

A minimal scripts provider JAR (`custom-authz-policy-example`) is added and
deployed into the local and CI Keycloak so both resources have a real
deployed-script policy to reference. Acceptance tests for both resources pass
against Keycloak 26.6.3, covering create, read, update, import, and the
deployed-script `code` handling.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,6 +93,9 @@ jobs:
       - name: Build user-federation-example with Gradle
         run:  make user-federation-example
 
+      - name: Build authz-policy-example
+        run:  make authz-policy-example
+
       - name: Get Keycloak Version
         id: keycloak-version
         shell: bash
@@ -118,6 +121,7 @@ jobs:
           EXTRA_HTTPS_KEY=""
           EXTRA_MTLS_CERTS=""
           MOUNT_FEDERATION_EXAMPLE_VOLUME="-v $PWD/custom-user-federation-example/build/libs/custom-user-federation-example-all.jar:/opt/keycloak/providers/custom-user-federation-example-all.jar:z"
+          MOUNT_AUTHZ_POLICY_EXAMPLE_VOLUME="-v $PWD/custom-authz-policy-example/build/keycloak-authz-policy-example.jar:/opt/keycloak/providers/keycloak-authz-policy-example.jar:z"
           if [[ "$KC_26_2_OR_LATER" == "true" ]]; then
             # When fgap-version is set explicitly use it; otherwise default to v2
             # (the default since Keycloak 26.2). The feature must be enabled
@@ -147,6 +151,7 @@ jobs:
           -e KC_FEATURES=preview${EXTRA_FEATURES} \
           -v $PWD/provider/testdata:/opt/keycloak/testdata:z \
           $MOUNT_FEDERATION_EXAMPLE_VOLUME \
+          $MOUNT_AUTHZ_POLICY_EXAMPLE_VOLUME \
           quay.io/keycloak/keycloak:${{ matrix.keycloak-version }} --verbose start-dev
 
       - name: Initialize Keycloak
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@ custom-user-federation-example/bin
 custom-user-federation-example/build
 !custom-user-federation-example/build/libs
 
+# custom authz policy example
+custom-authz-policy-example/build
+
 # docs
 site/
 
--- a/custom-authz-policy-example/META-INF/keycloak-scripts.json
+++ b/custom-authz-policy-example/META-INF/keycloak-scripts.json
@@ -1,0 +1,9 @@
+{
+  "policies": [
+    {
+      "name": "Always Granting Policy",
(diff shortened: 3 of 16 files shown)


sschu (approved):
LGTM, thanks a lot!

PR #1646, opened 2026-07-01
docs: add example for dynamic scope client

RafaelWO:
This change adds an example for how to create an OpenID client scope as a dynamic scope via `extra_config` using the `keycloak_openid_client_scope` resource.

The example is based on [this comment](https://github.com/keycloak/terraform-provider-keycloak/pull/1475#issuecomment-3929621713) in the PR that introduced `extra_config`. But I adapted the value of `dynamic.scope.regexp` based on the value I saw `attributes` of a client scope for which I toggled "Dynamic scope" in the GUI.

image

### DIFF
--- a/docs/resources/openid_client_scope.md
+++ b/docs/resources/openid_client_scope.md
@@ -27,6 +27,28 @@ resource "keycloak_openid_client_scope" "openid_client_scope" {
 }
 ```
 
+## Example Usage (Dynamic Scope)
+
+```hcl
+resource "keycloak_realm" "realm" {
+  realm   = "my-realm"
+  enabled = true
+}
+
+resource "keycloak_openid_client_scope" "dynamic_scope" {
+  realm_id                            = keycloak_realm.realm.id
+  name                                = "my-dynamic-scope"
+  description                         = "When requested, this scope will map a user's group memberships to a claim"
+  include_in_token_scope              = true
+  gui_order                           = 1
+
+  extra_config = {
+    "is.dynamic.scope"     = "true"
+    "dynamic.scope.regexp" = "my-dynamic-scope:*"
+  }
+}
+```
+
 ## Argument Reference
 
 - `realm_id` - (Required) The realm this client scope belongs to.


sschu:
@RafaelWO Thanks for your contribution! It is definitely an improvement. However, can you please also document that this example only works for Keycloak <26.7 and add an example showing how it looks like starting Keycloak 26.7 when it is called "Parameterized Scope"? As far as I can tell, its mostly identical, its just named `parameterized.scope....`. Also the regexp should be a regexp so use `.*` in the example. Thanks!

RafaelWO:
> However, can you please also document that this example only works for Keycloak <26.7 and add an example showing how it looks like starting Keycloak 26.7 

Thanks for letting me know that it is called "Parameterized scope" in Keycloak 26.7. I will add another example for that.

I will also add `"parameterized.scope.type": "custom"` for the example in 26.7.

> Also the regexp should be a regexp so use .* in the example. Thanks!

You are right that the regexp should be a real one, but I noticed that this is only the case for >=26.7.

For example, we have an instance running with version 26.2.5 and you cannot configure the scope format there:

image

In the JSON of this dynamic scope, the regexp is always fixed to `stage:*` and you cannot change it in the UI:

image

That's why I configured it the same way in the `keycloak_openid_client_scope` resource with the pattern `<scope-name>:*`.

Or am I missing something?

---

P.S. I just launched an Keycloak instance with version 26.6.0 and the dynamic scope behaves the same as described above with 26.2.5 (regexp is fixed an not correct).

sschu:
@RafaelWO As far as I could see, before Keycloak 26.7, the parameter was called regexp but it was really just a string with a single wildcard character. The value was always `<scopename>:*`(at least if defined by the UI). Starting Keycloak 26.7, to use a regular expression the type `Custom` has to be selected and the regular expression really is a regular expression with default value `.+`. The name of the scope and `:` is prepended to the regular expression when checking the scope name so the scope needs to have value `<scopename>:<regexp-matching-string>` to be valid. Does this help?

PR #1647, opened 2026-07-03
feat: adds name-based import support for protocol mappers

remcojansen:
This pull request introduces support for importing protocol mappers by name, in addition to the existing ID-based import. Documentation for all relevant protocol mapper resources has been updated to explain the new name-based import format, including usage instructions and examples. This enhancement improves usability, especially when the protocol mapper ID is not easily accessible.

**Functionality** 

Protocol mappers can now be imported by name in addition to by ID:

 terraform import keycloak_openid_user_attribute_protocol_mapper.example \
   realm/client/client-id/name/mapper-name

The import path parser is extended to handle a 5-part `realm/client/{clientId}/name/{urlEncodedName}` format alongside the existing 4-part ID-based format. Name lookup is centralized in a new `getProtocolMapperByName` helper in `generic_protocol_mapper.go`, which returns a clear error on duplicate names.

**Documentation**

* Updated the documentation files to describe and provide examples for the new name-based import format.

**Additional changes**
- Replace `ConflictsWith` with `ExactlyOneOf` for `client_id` / `client_scope_id` uniformly across all 14 protocol mapper resources, providing earlier plan-time validation.
- Refactor duplicate-name validation in `openid_user_attribute` and `openid_user_property` mappers to use the shared name-lookup helper; validation is now also enforced on Update, not only on Create.
- Add acceptance tests covering name-based import for both mapper types.

**Limitations**
Partially addresses #1494, but with this caveat: realm id and client id or client scope id need to be known during apply-time of the protocol mapper resources. Depending on the configuration this might require running staged `terraform apply` commands to ensure the realm and client/scope resources are available and their IDs are known when the protocol mappers are imported.


sschu (approved):
LGTM, thanks a lot @remcojansen !

PR #761, opened 2022-11-04
FEAT: Allow multivalue for Config attibutes

ChrisHubinger:
#548 

Implemented & documented the same multivalue behavior for `keycloak_custom_user_federation` resource as it's done in `keycloak_user`.

PR #1649, opened 2026-07-03
fix(custom_user_federation): keep all values of a multivalued config

### LINKED ISSUE #1398: custom_user_federation - config with string multi values
### Describe the bug

Multiple values ​​are overwritten during the application. Only first value is inserted into Keycloak.

```terraform
variable "config" {
  type        = map(string)
  description = "(Optional) The provider configuration handed over to your custom user federation provider. In order to add multivalued settings, use ## to separate the values."
  default     = {
    "originChannels" = "test=test##toto=toto##titi=titi"
  }
}

resource "keycloak_custom_user_federation" "custom_user_federation" {
  realm_id            = var.realm_id
  name                = var.name
  provider_id         = var.provider_id
  enabled             = var.enabled
  priority            = var.priority
  cache_policy        = var.cache_policy
  full_sync_period    = var.full_sync_period
  changed_sync_period = var.changed_sync_period
  config              = var.config
}
```
- Post apply, screen verification, only the value test=test

Image

- Plan relaunch
```
 ~ resource "keycloak_custom_user_federation" "custom_user_federation" {
      ~ config              = {
          ~ "originChannels"       = "test=test" -> "test=test##toto=toto##titi=titi"
            # (4 unchanged elements hidden)
        }
```

### Version

26.4.5

### Expected behavior

Image

### Actual behavior

If I enter the data on screen or apply it via Terraform, the Terraform provider only sees the first line during the plan and only applies one line during the application.

### How to Reproduce?

_No response_

### Anything else?

_No response_

emmanuel:
config values are split on `##` into a `[]string`, but only the first element was written to Keycloak (`j[0]`) and only the first was read back (in `getConfig`). Settings with more than one `##`-separated value lost everything after the first, and because the read side dropped them too the plan never settled.

Write and read the full slice instead.

#761 added the `##` split/join to `getCustomUserFederationFromData`/`setCustomUserFederationData`, but the two conversions in `keycloak/custom_user_federation.go` were left as-is, so the slice was re-collapsed to its first element crossing into and out of the component struct. This patches those two remaining spots.

Fixes #1398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ FEATURES:
 
 * feat: support import for `keycloak_realm_default_client_scopes` and `keycloak_realm_optional_client_scopes` using the realm id
 
+BUG FIXES:
+
+* fix(custom_user_federation): keep all values of a multivalued `##` config setting instead of only the first ([#1398](https://github.com/keycloak/terraform-provider-keycloak/issues/1398))
+
 ## 5.8.0 (June 05, 2026)
 
 BREAKING CHANGES:
--- a/docs/resources/custom_user_federation.md
+++ b/docs/resources/custom_user_federation.md
@@ -27,7 +27,7 @@ resource "keycloak_custom_user_federation" "custom_user_federation" {
   config = {
     dummyString = "foobar"
     dummyBool   = true
-    multivalue = "value1##value2"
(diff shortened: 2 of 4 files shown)


sschu (approved):
LGTM, thanks!

Issue #1657, opened 2026-07-15
Deprecate require_resident_key in Realm WebAuthn settings


Laureatgrepi:
### Description

Keycloak version 26.7.0 deprecated the `require_resident_key` config in WebAuthn settings and replaced it with `resident_key` (Confidential credential)

PR #1658, opened 2026-07-15
Deprecate require_resident_key config value in Realm WebAuthn settings (KC version 26.7.0)

Laureatgrepi:
Adds support for the new WebAuthn `discoverable_credential` setting (webAuthnPolicyResidentKey / webAuthnPolicyPasswordlessResidentKey, introduced in Keycloak 26.7 by keycloak/keycloak#49687) to both the web_authn_policy and web_authn_passwordless_policy blocks of the realm resource and data source.

The existing `require_resident_key` attribute is marked deprecated. It would still keep working and is only used by Keycloak when discoverable_credential is "not specified".

```
The setting is named in `Discoverable credential` in the Keycloak admin UI (hence the TF attribute naming)
```

The new attribute accepts `required`, `preferred`, `discouraged`, or `not specified`

Keycloak version >26.7.0 logs this annoyingly when setting the deprecated value:
```
[org.keycloak.models.WebAuthnPolicy] The WebAuthn policy option 'requireResidentKey' is deprecated and will be removed in the future. Use the 'residentKey' option instead.
[org.keycloak.models.WebAuthnPolicy] The WebAuthn policy option 'requireResidentKey' is deprecated and will be removed in the future. Use the 'residentKey' option instead.
[org.keycloak.models.WebAuthnPolicy] The WebAuthn policy option 'requireResidentKey' is deprecated and will be removed in the future. Use the 'residentKey' option instead.
```

Addresses #1657


sschu (approved):
LGTM, thanks @Laureatgrepi !

PR #1677, opened 2026-08-23
fix: Add allow_utf8 field to Keycloak realm data source

omarsamir27:
allow_utf8 is present in realm resource schema, but absent in the data source schema, which leads to using smtp_server from a realm data source to fail as deserialization does not find allow_utf8 attribute. This PR adds the missing attriubte


sschu:
@omarsamir27 Thanks for your contribution. I think the Copilot feedback is correct. Can you please address this to get this PR merged?

PR #1607, opened 2026-05-28
feat: Add support for Keycloak organization groups

dmavrommatis:
## Summary

Adds support for managing Keycloak organization groups through `keycloak_group`.

This includes:
- `organization_id` support on the `keycloak_group` resource
- create, read, update, delete, and import support for organization groups
- nested organization group support via `parent_id`
- `organization_id` support on the `keycloak_group` data source
- acceptance tests for resource and data source behavior
- documentation updates for resource usage, data source usage, and import format

Organization groups require Keycloak 26.6.0 or later.

`GetGroup` is kept as the existing realm-group API for backwards compatibility. It now delegates to `GetOrganizationGroup` with an empty organization ID, so existing callers keep working while organization-scoped callers can use the new Keycloak organization group endpoint.

Fixes #1561

PR #1680, opened 2026-08-27
feat: add organization_id support to keycloak_group_roles

Benthem:
## Summary

Adds an optional `organization_id` attribute to the `keycloak_group_roles` resource. When set, all role mapping operations use the Keycloak Organization API path instead of the regular groups API.

## Problem

Keycloak 26.6.0+ introduced organization groups, which must be managed via the Organization API. Attempting to manage role mappings for organization groups through the regular groups API results in:

```
Cannot manage organization related group via non Organization API.
```

The `data.keycloak_group` data source already supports `organization_id` (added in #1607, v5.8.0), but `keycloak_group_roles` does not — making it impossible to assign roles to organization groups via Terraform.

## Changes

- **`keycloak/group_role_mappings.go`**: Added a `groupRoleMappingsUrl` helper that routes to the organization API when `organizationId` is set. Added Organization variants of all role mapping functions. Original functions delegate with an empty `organizationId` to preserve backward compatibility.
- **`provider/resource_keycloak_group_roles.go`**: Added `organization_id` schema attribute (optional, ForceNew). All CRUD functions now use `GetOrganizationGroup` and organization-aware role mapping functions. Import supports both `{{realm}}/{{groupId}}` and `{{realm}}/{{organizationId}}/{{groupId}}` formats.
- **`docs/resources/group_roles.md`**: Added organization group example, updated argument reference and import format.

## Example usage

```hcl
data "keycloak_group" "org_group" {
  realm_id        = keycloak_realm.realm.id
  organization_id = keycloak_organization.my_org.id
  name            = "my-org-group"
}

resource "keycloak_group_roles" "org_group_roles" {
  realm_id        = keycloak_realm.realm.id
  group_id        = data.keycloak_group.org_group.id
  organization_id = keycloak_organization.my_org.id
  exhaustive      = false

  role_ids = [
    keycloak_role.client_role.id,
  ]
}
```

## Notes

- Requires Keycloak 26.6.0 or later for organization group support
- Fully backward compatible — omitting `organization_id` uses the existing groups API
- Builds on the organization group support added in #1607

### DIFF
--- a/docs/resources/group_roles.md
+++ b/docs/resources/group_roles.md
@@ -118,16 +118,35 @@ resource "keycloak_group_roles" "group_role_association2" {
 }
 
 ```
+
+## Example Usage (organization group roles)
+
+Organization groups can be managed by setting `organization_id`. This requires Keycloak 26.6.0 or later.
+
+```hcl
+resource "keycloak_group_roles" "org_group_roles" {
+  realm_id        = keycloak_realm.realm.id
+  group_id        = data.keycloak_group.org_group.id
+  organization_id = keycloak_organization.my_org.id
+  exhaustive      = false
+
+  role_ids = [
+    keycloak_role.client_role.id,
+  ]
+}
+```
+
 ## Argument Reference
 
 - `realm_id` - (Required) The realm this group exists in.
 - `group_id` - (Required) The ID of the group this resource should manage roles for.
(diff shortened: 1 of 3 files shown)


sschu:
@Benthem Thanks for your contribution! Can you please add some tests as Copilot suggests?

Issue #1562, opened 2026-04-17
Add `keycloak_realm_client_registration_policy` resource


QuellcodeWanderer:
### Description

There is currently no native Terraform resource for managing [Client Registration Policies](https://www.keycloak.org/docs/latest/securing_apps/#_client_registration_policies) in Keycloak. These are the policies found in the Keycloak Admin Console under **Realm Settings → Client Registration → Client Registration Policies**.

### Motivation

Without a native resource, users are forced to manage these policies outside of Terraform — typically via `data "external"` with a shell script or similar workarounds that call the Keycloak Admin REST API directly. This breaks the declarative, plan-aware lifecycle management that Terraform provides.

Common use cases:
- Creating custom policies backed by a custom SPI implementation
- Managing the `anonymous` vs `authenticated` policy sets per realm
- Taking ownership of Keycloak's default policies (e.g. "Trusted Hosts", "Max Clients Limit", "Consent Required") via `terraform import`, after which they can be updated or deleted through the normal Terraform lifecycle

### Details

Client Registration Policies are implemented as [components](https://www.keycloak.org/docs-api/latest/rest-api/index.html#_component_resource) in Keycloak with `providerType: org.keycloak.services.clientregistration.policy.ClientRegistrationPolicy`. A new `keycloak_realm_client_registration_policy` resource would wrap the existing `/admin/realms/{realm}/components` CRUD endpoints.

```hcl
resource "keycloak_realm_client_registration_policy" "custom" {
  realm_id    = keycloak_realm.realm.id
  name        = "My Custom Policy"
  provider_id = "my-client-registration-policy"   # custom SPI provider ID
  sub_type    = "anonymous"                       # "anonymous" | "authenticated"
}
```

The existing `component` struct in `keycloak/component.go` can be reused directly, following the same pattern already used by keystore resources (e.g. `keycloak_realm_keystore_rsa`). The only addition needed to the `component` struct is the `subType` field, which is used to differentiate anonymous vs. authenticated policies.

I have a working implementation ready and am happy to open a PR.

nlenepveu:
There are now two open PRs implementing this feature from different angles, and since they're easy to read as competing, here's a side-by-side in case it helps maintainers pick a direction.

**#1563** (@QuellcodeWanderer, closes this issue) — a single **generic** resource `keycloak_realm_client_registration_policy` with `config = map(string)`, `provider_id`, `sub_type`.

[...]

Issue #802, opened 2023-02-02
wrong parentId for realm's components when creating new realm with terraform


pescobar:
I have realized that the provider is not defining the right `parentId` for the realm components when creating new realms. 

I tested it with keycloak 19.0.3 and 20.0.3.  You can reproduce the problem like this:

First boot a keycloak test instance listening in http://localhost:8080 with username `admin` and password `admin`:
```
$> docker run -p 8080:8080 -e KEYCLOAK_ADMIN='admin' -e KEYCLOAK_ADMIN_PASSWORD='admin' quay.io/keycloak/keycloak:19.0.3 start-dev
```

Then you can create a new realm using this terraform code:

```hcl
terraform {
  required_providers {
    keycloak = {
      source = "mrparkers/keycloak"
      version = "4.1.0"
    }
  }
}

provider "keycloak" {
    client_id     = "admin-cli"
    username      = "admin"
    password      = "admin"
    url           = "http://localhost:8080"
}

resource "keycloak_realm" "myrealm" {
  realm           = "myrealm"
}
```

If you query the components for the new realm `myrealm` created with terraform you get this:

```bash
$> mastertoken=$(curl -s -k -g -d "client_id=admin-cli" -d "username=admin" -d "password=admin" -d "grant_type=password" -d "client_secret=" "http://localhost:8080/realms/master/protocol/openid-connect/token" | sed 's/.*access_token":"//g' | sed 's/".*//g')

$> curl -s -k -X GET http://localhost:8080/admin/realms/myrealm/components/ -H "Content-Type: application/json" -H "Authorization: bearer $mastertoken" |jq .|grep parent
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
    "parentId": "myrealm",
```

But if I create another realm named `anotherrealm` **using the keycloak webui** the parentIDs for the components are in different format:

```bash
$> mastertoken=$(curl -s -k -g -d "client_id=admin-cli" -d "username=admin" -d "password=admin" -d "grant_type=password" -d "client_secret=" "http://localhost:8080/realms/master/protocol/openid-connect/token" | sed 's/.*access_token":"//g' | sed 's/".*//g')

[...]

PR #1683, opened 2026-09-01
fix(realm_client_registration_policy): let keycloak resolve the component parentId

kchledowski:
`realm_client_registration_policy` (introduced for #1562 ) specific fix for an issue discribed in #802 

Components attached to a realm must carry the realm's internal id as parentId, not the realm name. The two differ when the realm was created outside terraform (e.g. in the admin console or by Keycloak itself, like master realm) and imported.

Instead of resolving the internal id client-side via GetRealm (the approach used in PR #707), omit parentId from the component payload entirely: keycloak defaults a missing parentId to the realm's internal id on create and preserves the stored parent on update, so the server resolves the correct value without an extra API call. This also avoids sending an empty parentId when GetRealm returns a stripped realm representation (token without full view permissions on the realm).


sschu (approved):
LGTM, thanks a lot!

PR #1690, opened 2026-09-07
fix: sort multivalue attribute values on keycloak_role and organization resource before diff

### LINKED ISSUE #859: Group attributes are not organized in a linear way
Currently, group mappers are organized according to their IDs in the database, and that is the order they will have in the state file. 
This makes it so that even if your mappers are the same as they are in the state but unordered they will show up in the plan as an update. This happens at least when the realm was already created and the group has been imported. It would be interese
To give an example of the problem:
1- Create group resource and variable:

```
resource "keycloak_group" "example_group" {
  realm_id   = var.realm_id
  name       = var.group.name
  attributes = var.group.attributes
}

variable "group" {
type = any
}

group = {
name="example_group"
attributes = "attribute1##attribute2##attribute3"
}
```

Run apply. If you are lucky the attributes will be created in the database in the same order you created them. But this will probably not happen. If you run `terraform plan` again. You may see something like this:

imagem

My suggestion is to order the attributes that terraforms receives when refreshing the state (e.g. alphabetically) and order the inputs from the user in the same way. 

I am available to create a PR since this is a fix we require.

starsep:
Inspired by #1520
Fixes #859 

---

AI Disclaimer: Change prepared with Claude Opus 5

I confirmed manually that with this change `apply` no longer displays diffs caused by order change

### DIFF
--- a/provider/diff_suppress_test.go
+++ b/provider/diff_suppress_test.go
@@ -1,0 +1,120 @@
+package provider
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestSuppressDiffForMultivalueAttributeOrder(t *testing.T) {
+	t.Parallel()
+
+	suppress := suppressDiffForMultivalueAttributeOrder()
+
+	tests := []struct {
+		name     string
+		old      string
+		new      string
+		expected bool
+	}{
+		{
+			name:     "identical single value",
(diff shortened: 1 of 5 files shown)


sschu (approved):
LGTM, thanks!

